PR #7, opened 2015-08-25
Hikey: full GP support

pascal-brand-st-dev:


pascal-brand-st-dev:
@jforissier @vchong what do you think of this pull-request?

jforissier:
@pascal-brand-st-dev this looks OK, and the comments I have made might be better done in a separate patch since they are applicable elsewhere.
How does one test this patch? Extract the GP zip then run `make xtest-patch` before plain `make`? (I am currently doing `repo sync` with `hikey.xml`).

pascal-brand-st-dev:
Yes, this is the way I do it.

Then once you added the RB, I can merge. I'll add a new PR about all your comments, trying to commonalize when possible

pascal-brand-st-dev:
You should have the following defined:

    CFG_GP_TESTSUITE_ENABLE=y
    CFG_GP_PACKAGE_PATH as you did

`make xtest-patch` will patch the xtest, so that more tests are generated (host .c file and TA) and some are modified. It has to be ran a single time.

Then `make xtest` will compile them (so with GP tests when `make xtest-patch` has been run).

jforissier:
Ha! Right, I forgot to define `CFG_GP_TESTSUITE_ENABLE=y`... thanks. IMO there should be only one variable (`*_PATH`) to begin with, but you could also set `*_ENABLE` automatically when `*_PATH` is defined.
Some instructions would be welcome, too. Maybe in https://github.com/OP-TEE/manifest/blob/master/README.md, or in the commit description.

Anyway:

pascal-brand-st-dev:
Right. Anyway, I think that more cleanup could be done with respect to the patch.

vchong:
@pascal-brand-st-dev Sorry for late reply and thanks for updating this! Better more tests and less.

PR #8, opened 2015-09-01
Common

pascal-brand-st-dev:
Waiting for comments. I still need to validate on every platform (currently only check on qemu).

In case the "concept" is adopted, more common part can be found (busybox, linux,...)

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -2,7 +2,7 @@
 # Common definition to all platforms
 #
 
-BASH := $(shell which bash)
+BASH ?= bash
 ROOT ?= $(shell pwd)/..
 
 #
@@ -23,6 +23,45 @@ OPTEE_TEST_PATH			?= $(ROOT)/optee_test
 OPTEE_TEST_OUT_PATH 		?= $(ROOT)/optee_test/out
 
 CCACHE ?= $(shell which ccache) # Don't remove this comment (space is needed)
+
+################################################################################
+# defines, macros, configuration etc
+################################################################################
+define KERNEL_VERSION
+$(shell cd $(LINUX_PATH) && make --no-print-directory kernelversion)
+endef
+DEBUG ?= 0
+
+################################################################################
+# OP-TEE
+################################################################################
+optee-os-common:
+	make -C $(OPTEE_OS_PATH) \
+		CROSS_COMPILE=$(CROSS_COMPILE_S_USER) \
+		CROSS_COMPILE_core=$(CROSS_COMPILE_S_KERNEL) \
+		CFG_TEE_CORE_LOG_LEVEL=3 \
+		DEBUG=$(DEBUG) \
+
+optee-os-clean-common:
+	make -C $(OPTEE_OS_PATH) \
+		clean
+
+optee-client-common:
+	make -C $(OPTEE_CLIENT_PATH) \
+		CROSS_COMPILE=$(CROSS_COMPILE_NS_USER)
+
+optee-client-clean-common:
+	make -C $(OPTEE_CLIENT_PATH) clean
+
+optee-linuxdriver-common: linux
+	make -C $(LINUX_PATH) \
+		CROSS_COMPILE=$(CROSS_COMPILE_NS_KERNEL) \
+		LOCALVERSION= \
+		M=$(OPTEE_LINUXDRIVER_PATH) modules
+
+optee-linuxdriver-clean-common:
+	make -C $(LINUX_PATH) \
+		M=$(OPTEE_LINUXDRIVER_PATH) clean
 
 ################################################################################
 # xtest / optee_test
--- a/common.mk
+++ b/common.mk
@@ -1,0 +1,62 @@
+#
+# Common definition to all platforms
+#
+
+BASH := $(shell which bash)
+ROOT ?= $(shell pwd)/..
+
+#
+# Must declare in platform specific makefiles:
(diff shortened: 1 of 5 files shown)


pascal-brand-st-dev:
Compilation fine on all platforms (qemu / fvp / mediatek / hikey)

pascal-brand-st-dev:
@jbech-linaro  @jforissier let me know if this looks good and if I should continue on this path or not. Thanks, Pascal.

jbech-linaro:
@pascal-brand-st-dev I think this looks like a good proposal and I have no problem going this way. I believe @jforissier are also doing some changes in this area for the toolchain etc. Just be careful so the both of you aren't trying to make some incompatible changes.

jforissier:
IMO this looks good, and yes @jbech-linaro I do have some pending changes to `qemu.mk` that will likely conflict but nothing difficult to fix, so I can rebase my changes when this PR is merged.

common.mk
```diff
@@ -0,0 +1,102 @@
+#
+# Common definition to all platforms
+#
+
+BASH := $(shell which bash)
```
jforissier: `?= bash` (we don't really need an absolute path, do we?)

common.mk
```diff
@@ -0,0 +1,102 @@
...
+
+LINUX_PATH			?= $(ROOT)/linux
+OPTEE_OS_PATH			?= $(ROOT)/optee_os
+OPTEE_CLIENT_PATH		?= $(ROOT)/optee_client
+OPTEE_CLIENT_EXPORT		?= $(OPTEE_CLIENT_PATH)/out/export
+OPTEE_LINUXDRIVER_PATH		?= $(ROOT)/optee_linuxdriver
+OPTEE_TEST_PATH			?= $(ROOT)/optee_test
+OPTEE_TEST_OUT_PATH 		?= $(ROOT)/optee_test/out
+
+CCACHE ?= $(shell which ccache) # Don't remove this comment (space is needed)
```
jforissier: Same as above -- `CCACHE ?= ccache # ...`

common.mk
pascal-brand-st-dev: Right.

common.mk
```diff
@@ -0,0 +1,102 @@
...
+optee-client-common:
+	make -C $(OPTEE_CLIENT_PATH) \
+		CROSS_COMPILE=$(CROSS_COMPILE_NS_USER)
+
+optee-client-clean-common:
+	make -C $(OPTEE_CLIENT_PATH) clean
+
+optee-linuxdriver-common: linux
+	make -C $(LINUX_PATH) \
+		V=0 \
```
jforissier: Remove (it's the default and prevents override)

common.mk
jbech-linaro: Shouldn't be needed, I think I added that for "better safe than sorry". But .. when I'm thinking about it, doesn't `which` just use what is in your `PATH`? I.e, if yes, then would be really useless to have. I mean "`which`"  isn't something that is tracked with some internal database (like the `locate` command)?

common.mk
pascal-brand-st-dev: No for this one. In case ccache is **not** installed, then `which ccache` will not output anything, and this will not lead to an error.

jbech-linaro: Exactly Pascal.

fvp.mk
```diff
@@ -1,40 +1,29 @@
...
+EDK2_PATH		?= $(ROOT)/edk2
+EDK2_BIN		?= $(EDK2_PATH)/Build/ArmVExpress-FVP-AArch64/RELEASE_GCC49/FV/FVP_AARCH64_EFI.fd
 
-CCACHE ?= $(shell which ccache) # Don't remove this comment (space is needed)
+GEN_ROOTFS_PATH		?= $(ROOT)/gen_rootfs
+GEN_ROOTFS_FILELIST	?= $(GEN_ROOTFS_PATH)/filelist-tee.txt
 
+FOUNDATION_PATH		?= $(ROOT)/Foundation_Platformpkg
+
+DEBUG = 1
```
jforissier: Instead of this, maybe set `DEBUG ?= 1` before `-include common.mk` to allow override with an environment variable? (and for consistency)

common.mk
jforissier: @jbech-linaro exactly, only `PATH` according to the man page of which(1).

common.mk
jforissier: Ha! Right, my mistake.

common.mk
jbech-linaro: +1

fvp.mk
jbech-linaro: +1

pascal-brand-st-dev:
Updated wrt comments

jbech-linaro:

qemu.mk
```diff
@@ -1,38 +1,27 @@
...
 
-################################################################################
-# defines, macros, configuration etc
-################################################################################
-define KERNEL_VERSION
-$(shell cd $(LINUX_PATH) && $(MAKE) --no-print-directory kernelversion)
-endef
-
-CCACHE ?= $(shell which ccache) # Don't remove this comment (space is needed)
+DEBUG = 1
```
jforissier: Move this one too (sorry for not seeing it before)

jforissier:

PR #9, opened 2015-09-04
Rework common.mk

pascal-brand-st-dev:
The code added by commits d6536da64deb ("Commonlize xtest compilation
among all platforms") and 070d95595e1d1 ("common build: optee
components") introduces race conditions that break parallel build (make
-jX). This is due to manually launching a make sub-process to run the
-common targets, and can be fixed by just declaring a dependency
instead. Specific flags are appended to common flags, rather than
passed as arguments to a new "make" sub-process.

So, basically:

 foo-common:
 	$(MAKE) -C foo_dir [A=a B=B]

 foo:
 	$(MAKE) foo-common [C=c]

is turned into:

 FOO_COMMON_FLAGS ?= A=a B=b
 foo-common:
 	$(MAKE) -C foo_dir $(FOO_COMMON_FLAGS)

 FOO_COMMON_FLAGS += C=c
 foo: foo-common

At the same time this allows foo: to override flags, for instance
set A=aa in the above example (prior to this, A=a would have had a
higher priority).

### DIFF
--- a/qemu.mk
+++ b/qemu.mk
@@ -101,23 +101,20 @@ linux-clean:
 ################################################################################
 # OP-TEE
 ################################################################################
-optee-os:
-	$(MAKE) \
-		PLATFORM=vexpress \
-		PLATFORM_FLAVOR=qemu_virt \
-		CFG_TEE_CORE_LOG_LEVEL=3 \
-			optee-os-common
+OPTEE_OS_COMMON_FLAGS += PLATFORM=vexpress-qemu_virt
+optee-os: optee-os-common
 
-optee-os-clean:
-	$(MAKE) \
-		PLATFORM=vexpress \
-		PLATFORM_FLAVOR=qemu_virt \
-			optee-os-clean-common
+OPTEE_OS_CLEAN_COMMON_FLAGS += PLATFORM=vexpress-qemu_virt
+optee-os-clean: optee-os-clean-common
 
 optee-client: optee-client-common
+
 optee-client-clean: optee-client-clean-common
-optee-linuxdriver:
-	$(MAKE)	ARCH=arm optee-linuxdriver-common
+
+OPTEE_LINUXDRIVER_COMMON_FLAGS += ARCH=arm
+optee-linuxdriver: optee-linuxdriver-common
+
+OPTEE_LINUXDRIVER_CLEAN_COMMON_FLAGS += ARCH=arm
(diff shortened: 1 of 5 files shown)


qemu.mk
```diff
@@ -195,7 +191,8 @@ endef
 
 .PHONY: run
 # This target enforces updating root fs etc
-run: | bios-qemu run-only
+run: bios-qemu
```
jforissier: You need qemu and soc-term, too. Or just use: `run: all`.

jforissier:
@pascal-brand-st-dev thanks for creating the PR.

With the comment I made re. `qemu.mk`, and for the files that were not in my original patch (i.e., `fvp.mk`, `hikey.mk` and `mediatek.mk`):

You should perhaps add your `Signed-off-by`, since you wrote parts of this patch?

pascal-brand-st-dev:
Updated

pascal-brand-st-dev:
* hikey / mediatek / qemu are compiling fine with `make -j8 all`
* Still some issues on fvp with parallel build.

Should I merge anyway?

jforissier:
We'd better fix the remaining issue(s) I think. I have a QEMU environment, is there a way to switch to FVP without cloning everything again? Then I can take a look if you wish.

jbech-linaro:
@jforissier I was switching back and forth when creating the manifest and usually it goes without too much trouble. If you have files that has been modified (kernel, qemu I think), then I recommend that you run `git checkout -f && git clean -xdf` before switching. When you are about to switch, just simply run
`repo init ....` with the fvp xml file instead, and then just run `repo sync` again.

By the way, consider setting up a reference for future use as described in our instructions. It save a huge amount of time.

jforissier:
@jbech-linaro thanks for the clarification. I'll setup a reference and let it clone during the week end.

As I was reading README.md I thought maybe this reference could be used only for one platform (i.e., have a fixed set of git repositories and just switch branches or commits). Now I understand that subprojects and references/branches/commits are all defined in the manifest files, so it makes sense one can switch as desired.

pascal-brand-st-dev:
@jforissier I'll work on this then.

pascal-brand-st-dev:
`make -j8 run` is working fine on fvp. So now parallel compilation can be used on all platforms.

jforissier:
Apart from the style issues which you may or may not change, as you prefer:

Issue #18, opened 2015-10-26
'make run' for QEMU config hangs after starting the first terminal window


pm215:
If I do a 'make run' (or 'make run-only') for the QEMU config, following the instructions in https://github.com/OP-TEE/optee_os, then this hangs after starting the first terminal window.

This seems to be because the makefile process does (more or less)
   gnome_terminal
   gnome_terminal
   qemu

and since the gnome_terminal execution isn't backgrounded then we never continue past starting the first terminal window. Adding an '&' to the definiton of the launch-terminal function fixes this for me:

```
diff --git a/qemu.mk b/qemu.mk
index d278895..3754554 100644
--- a/qemu.mk
+++ b/qemu.mk
@@ -193,7 +193,7 @@ endef
 
 define launch-terminal
        @nc -z  127.0.0.1 $(1) || \
-       gnome-terminal -e "$(BASH) -c '$(SOC_TERM_PATH)/soc_term $(1); exec /bin/bash -i'" --title=$(2)
+       gnome-terminal -e "$(BASH) -c '$(SOC_TERM_PATH)/soc_term $(1); exec /bin/bash -i'" --title=$(2) &
 endef
 
 .PHONY: run
```

jforissier:
Hi Peter, thanks for reporting this.
Which distribution/gnome-terminal are you running? I have no problem with Ubuntu 15.04. Now, adding a '&' certainly doesn't hurt, so I guess we should do that.

pm215:
This is Ubuntu trusty.
```
$ gnome-terminal --version
GNOME Terminal 3.6.2
```

pm215:
Ah, it turns out that if you already have *one* gnome-terminal running, then running 'gnome-terminal' from the command line will start a new window (presumably by talking to the other process) and then return to the command line. But if this 'gnome-terminal' is the first one, then it does not background itself or return to the command line until all the windows have closed.

I guess that you use gnome-terminal as your standard terminal, so it works for you. I use xterm, so the gnome-terminal started by this makefile is typically the only one running.

ototo:
It might make sense to replace gnome-terminal in all scripts with xterm as xterm seems to be installed by default on majority of distros (haven't seen one that would not have it yet), while gnome-terminal is only installed by default when distro is gnome based or gnome DE has been installed.

jbech-linaro:
I've never seen this problem myself, but indeed, I use gnome-terminal as my standard terminal.

[...]

PR #19, opened 2015-10-27
qemu.mk: use xterm instead of gnome-terminal

jforissier:
- Make sure launch-terminal cannot block by adding a '&'
- Use xterm instead of gnome-terminal because it is more likely to be
installed by default.

Fixes https://github.com/OP-TEE/build/issues/18.

### DIFF
--- a/qemu.mk
+++ b/qemu.mk
@@ -193,7 +193,7 @@ endef
 
 define launch-terminal
 	@nc -z  127.0.0.1 $(1) || \
-	gnome-terminal -e "$(BASH) -c '$(SOC_TERM_PATH)/soc_term $(1); exec /bin/bash -i'" --title=$(2)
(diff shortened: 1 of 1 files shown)


jbech-linaro:
Any "Tested-by: ..." Jerome?

jforissier:
@jbech-linaro well that's tested by me, but I would think this T-b would be implicit here?

pascal-brand-st-dev:
Reviewed-by: Pascal Brand <pascal.brand@linaro.org>
    Tested-by: Pascal Brand <pascal.brand@linaro.org> (QEMU)

jforissier:
Tags applied.

pm215:
Tested-by: Peter Maydell <peter.maydell@linaro.org>

This fixes the issue I originally reported.

jforissier:
Updated -- thanks Peter. I have also added your Reported-by:, for the record ;)

PR #24, opened 2015-11-12
hikey: Add sleep to prevent umount failure

jbech-linaro:
Sometimes it happens that the umount fails saying that a device/resource
is busy when trying umount the device after the filesystem has been
created. We don't know how and why this happens, but in the meantime
while waiting for a proper solution we have added a sleep that seems to
solve the problem.


jenswi-linaro:
```
```

jbech-linaro:
Tag applied, ready to be merged.

PR #25, opened 2015-11-18
Fix the errors present in QEMU instructions-a64.h

lordqwerty:
When compiling OP-TEE for QEMU with a recent version of GCC it
causes build errors `-Werror=unused-variable`.

By adding `--extra-cflags="-Wno-error"` to the makefile it safely
supresses the errors and allows OP-TEE to be compiled and built. This
error is also present in the `default_stable.xml` configuration with the
repo tool of the project.

See [OPTEE_OS issue 550](https://github.com/OP-TEE/optee_os/issues/550).

@vchong is also aware of issue.

### DIFF
--- a/qemu.mk
+++ b/qemu.mk
@@ -57,7 +57,7 @@ bios-qemu-clean:
 	$(call bios-qemu-common) clean
 
 qemu:
-	cd $(QEMU_PATH); ./configure --target-list=arm-softmmu --cc="$(CCACHE)gcc"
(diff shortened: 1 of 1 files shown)


lordqwerty:
@vchong already done a fix. See: https://github.com/vchong/build/commit/7f58c96e71163c74b506cfae44813495e5e4aa96

lordqwerty:
Advised to re-open, sorry for the confusion.

PR #27, opened 2015-11-19
qemu: enable kernel CONFIG_PREEMPT

jforissier:
This will hopefully fix occasional errors observed during OP-TEE Travis
regression testing (make check):
  rcu_sched detected stalls on CPUs/tasks...

### DIFF
--- a/qemu.mk
+++ b/qemu.mk
@@ -79,10 +79,12 @@ busybox-cleaner: busybox-cleaner-common
 ################################################################################
 # Linux kernel
 ################################################################################
-$(LINUX_PATH)/.config:
+$(LINUX_PATH)/.config: kconfig_preempt.conf
 	# Temporary fix until we have the driver integrated in the kernel
 	sed -i '/config ARM$$/a select DMA_SHARED_BUFFER' $(LINUX_PATH)/arch/arm/Kconfig;
-	$(MAKE) -C $(LINUX_PATH) ARCH=arm vexpress_defconfig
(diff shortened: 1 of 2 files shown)


jenswi-linaro:
```
```

qemu.mk
```diff
@@ -79,10 +79,12 @@ busybox-cleaner: busybox-cleaner-common
 ################################################################################
 # Linux kernel
 ################################################################################
-$(LINUX_PATH)/.config:
+$(LINUX_PATH)/.config: kconfig_preempt.conf
```
jbech-linaro: Does the target `kconfig_preempt.conf` really make any sense here? I mean this makefile isn't in `$(LINUX_PATH)` when being called. I.e, shouldn't it be `$(LINUX_PATH)/kconfig_preempt.conf`?

However, I have test built using it as it is and indeed it is working.

jbech-linaro:
Besides my comment:
`Tested-by: Joakim Bech <joakim.bech@linaro.org> (QEMU)`

qemu.mk
jforissier: `kconfig_preempt.conf` is a file added by this commit in the same location as `qemu.mk`, so the path is correct.

jforissier:
Tags added.

qemu.mk
jbech-linaro: You're right, my bad! For some reason I was just thinking it was added to the kernel dir.

Issue #28, opened 2015-11-23
RFC: Bug in *_CLEAN_COMMON_FLAGS?


vchong:
Hi,

In common.mk:
```LINUX_CLEAN_COMMON_FLAGS ?= $(LINUX_COMMON_FLAGS)```

In hikey.mk:
```LINUX_COMMON_FLAGS += ARCH=arm64 Image modules dtbs```
```LINUX_CLEAN_COMMON_FLAGS += ARCH=arm64```

```
> make linux-clean
rm -f /home/devel/optee/repo/build/../linux/.config
make -C /home/devel/optee/repo/build/../linux
LOCALVERSION= CROSS_COMPILE="/usr/bin/ccache
/home/devel/optee/repo/build/../toolchains/aarch64/bin/aarch64-linux-gnu-"
ARCH=arm64 Image modules dtbs ARCH=arm64 clean
```

So we're re-making everything before cleaning?

For other platforms, it's not as bad since we run ```make``` without specifying any targets so we only repeat the ```ARCH=arm64``` flag, i.e.

```
> make linux-clean
rm -f /home/devel/optee/repo/build/../linux/.config
make -C /home/devel/optee/repo/build/../linux
LOCALVERSION= CROSS_COMPILE="/usr/bin/ccache
/home/devel/optee/repo/build/../toolchains/aarch64/bin/aarch64-linux-gnu-"
ARCH=arm64 ARCH=arm64 clean
```

Same situation for ```OPTEE_OS_CLEAN_COMMON_FLAGS``` and ```OPTEE_LINUXDRIVER_CLEAN_COMMON_FLAGS```.

How should we fix this? Quick and easy way is to just redefine ```LINUX_CLEAN_COMMON_FLAGS``` in hikey.mk, i.e.
```LINUX_CLEAN_COMMON_FLAGS = LOCALVERSION= CROSS_COMPILE=$(CROSS_COMPILE_NS_KERNEL) ARCH=arm64```?
Or ```LINUX_CLEAN_COMMON_FLAGS := LOCALVERSION= CROSS_COMPILE=$(CROSS_COMPILE_NS_KERNEL) ARCH=arm64```?
Or remove the dependency of *_CLEAN_COMMON_FLAGS on *_COMMON_FLAGS?
Or something else?

Thanks!

jbech-linaro:
[...]

PR #29, opened 2015-11-24
[RFC] Remove *_CLEAN_COMMON_FLAGS dependency on *_COMMON_FLAGS

vchong:
Addresses https://github.com/OP-TEE/build/issues/28

The *_CLEAN_COMMON_FLAGS don't really need the variables/arguments/etc defined in *_COMMON_FLAGS for ```make clean```, so I just commented them out. The required variables are already defined in the platform .mk files themselves.

The only exception is maybe xtest. Do ```TA_DEV_KIT_DIR```, ```CFG_DEV_PATH``` and ```O``` have to be defined for cleaning?

Is this ok, or another method preferred?

Signed-off-by: Victor Chong <victor.chogn@linaro.org>

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -73,15 +73,17 @@ linux-common: linux-defconfig
 linux-defconfig-clean-common:
 	@rm -f $(LINUX_PATH)/.config
 
-LINUX_CLEAN_COMMON_FLAGS ?= $(LINUX_COMMON_FLAGS)
-	
+# LINUX_CLEAN_COMMON_FLAGS can be defined in specific makefiles (hikey.mk,...)
+# if necessary
+
 linux-clean-common: linux-defconfig-clean
 	$(MAKE) -C $(LINUX_PATH) $(LINUX_CLEAN_COMMON_FLAGS) clean
 
-LINUX_CLEANER_COMMON_FLAGS ?= $(LINUX_COMMON_FLAGS)
+# LINUX_CLEANER_COMMON_FLAGS can be defined in specific makefiles (hikey.mk,...) 
+# if necessary
 
 linux-cleaner-common: linux-defconfig-clean
-	$(MAKE) -C $(LINUX_PATH) $(LINUX_CLEAN_COMMON_FLAGS) distclean
+	$(MAKE) -C $(LINUX_PATH) $(LINUX_CLEANER_COMMON_FLAGS) distclean
 
 ################################################################################
 # EDK2 / Tianocore
@@ -113,7 +115,8 @@ OPTEE_OS_COMMON_FLAGS ?= CROSS_COMPILE=$(CROSS_COMPILE_S_USER) \
 optee-os-common:
 	$(MAKE) -C $(OPTEE_OS_PATH) $(OPTEE_OS_COMMON_FLAGS)
 
-OPTEE_OS_CLEAN_COMMON_FLAGS ?= $(OPTEE_OS_COMMON_FLAGS)
+# OPTEE_OS_CLEAN_COMMON_FLAGS can be defined in specific makefiles
+# (hikey.mk,...) if necessary
 
 optee-os-clean-common: xtest-clean
 	$(MAKE) -C $(OPTEE_OS_PATH) $(OPTEE_OS_CLEAN_COMMON_FLAGS) clean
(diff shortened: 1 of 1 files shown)


vchong:
Based on https://github.com/OP-TEE/optee_test/blob/master/Makefile#L41-L49 added  ```TA_DEV_KIT_DIR``` and ```O``` to ```XTEST_CLEAN_COMMON_FLAGS```.

```CFG_DEV_PATH``` doesn't seem to be used anywhere.

File: common.mk [hunk 1]
```diff
@@ -73,15 +73,15 @@ linux-common: linux-defconfig
 linux-defconfig-clean-common:
 	@rm -f $(LINUX_PATH)/.config
 
-LINUX_CLEAN_COMMON_FLAGS ?= $(LINUX_COMMON_FLAGS)
+#LINUX_CLEAN_COMMON_FLAGS ?= UNCOMMENT_AND_DEFINE_AS_NECESSARY
```
pascal-brand-st-dev: I would prefer instead:

    # LINUX_CLEAN_COMMON_FLAGS can be defined in specific makefiles (hikey.mk,...) if necessary

pascal-brand-st-dev:
In the commit message, I think that "Addresses #28" should be modified in "Fixes #28". Also, the commit message should contain the reason of the change instead of only relying on the description in #28.

common.mk
jbech-linaro: I'm not fond of having code not actually used, even though it might sound like it is convenient. Also for the Linux related flags we must propagate the ARCH here. Today we're stating that directly at the line on the individual makefiles, like here for example: https://github.com/OP-TEE/build/blob/master/qemu.mk#L91

Maybe we should create an `ARCH` variable in the individual makefiles and just include that here?

jbech-linaro: This would be sufficient, the other stuff could be kept as it is in qemu.mk, hikey.mk etc... since we already have the correct ARCH flag there.

```
diff --git a/common.mk b/common.mk
index ae98c70..ece5ff5 100644
--- a/common.mk
+++ b/common.mk
@@ -73,7 +73,7 @@ linux-common: linux-defconfig
 linux-defconfig-clean-common:
        @rm -f $(LINUX_PATH)/.config
 
-LINUX_CLEAN_COMMON_FLAGS ?= $(LINUX_COMMON_FLAGS)
+LINUX_CLEAN_COMMON_FLAGS ?= CROSS_COMPILE=$(CROSS_COMPILE_NS_KERNEL)
        
 linux-clean-common: linux-defconfig-clean
        $(MAKE) -C $(LINUX_PATH) $(LINUX_CLEAN_COMMON_FLAGS) clean
```

qemu gives ...
```
make -C /home/jbech/devel/optee/build/../linux CROSS_COMPILE="/usr/bin/ccache /home/jbech/devel/optee/build/../toolchains/aarch32/bin/arm-linux-gnueabihf-" ARCH=arm clean
```

hikey gives ...
```
make -C /home/jbech/devel/optee/build/../linux CROSS_COMPILE="/usr/bin/ccache /home/jbech/devel/optee/build/../toolchains/aarch64/bin/aarch64-linux-gnu-" ARCH=arm64 clean
```

etc ...

vchong: @pascal-brand-st-dev Do you prefer that just for LINUX_CLEAN_COMMON_FLAGS or all the *CLEAN_COMMON_FLAGS?

@jbech-linaro Yes, ARCH is propagated here from the <platform>.mk even if LINUX_CLEAN_COMMON_FLAGS is commented out here, so why do we need to define CROSS_COMPILE for cleaning? Nothing is built.

vchong:
Ok. Will modify and add more details to the commit message.

common.mk
jbech-linaro: @vchong , maybe the `CROSS_COMPILE` isn't needed, it may very well be the case. Usually there is no need to include that variable when cleaning. One can just test it:
```
build && make clean && git clean -xfdn > only_clean.txt 
build && make CROSS_COMPILE=... clean && git clean -xfdn > cc_clean.txt 
diff -Naur only_clean.txt cc_clean.txt
```
If there's no diff, then

vchong: Ok. Just tested and diff returned empty, so I'll just comment this out with the description Pascal suggested above.

Also will close this and submit a new PR for it.

vchong:
Replaced by #31.

PR #30, opened 2015-11-25
qemu: disable kernel CONFIG_JFFS2_FS

jforissier:
We have observed random failures of "make check" when run in the
Travis environment (OP-TEE regression testing). Every time it fails,
the log file has the following errors [1]:

  Starting QEMU...send: spawn id exp7 not open
      while executing
  "send -- "\r""
      (file "qemu-check.exp" line 54)
  == /home/travis/optee_repo/build/serial0.log:
  spawn ../qemu/arm-softmmu/qemu-system-arm -nographic [...]
  Using QEMU provided DTB at 0x40000000
  [...]
  squashfs: version 4.0 (2009/01/31) Phillip Lougher
  jffs2: version 2.2. (NAND) send: spawn id exp7 not open
      while executing
  "send -- "\r""
      (file "qemu-check.exp" line 54)
  == end of /home/travis/optee_repo/build/serial0.log

It seems that QEMU exits unexpectedly as the kernel is initializing
the JFFS2 filesystem driver. Since this is something we don't need, I
thought I'd just disable this driver and see if it makes things more
stable.
Well, indeed I could not reproduce the issue with this patch applied.

[1] https://travis-ci.org/OP-TEE/optee_os/builds/90051975#L4577

### DIFF
--- a/kconfig_no_jffs2.conf
+++ b/kconfig_no_jffs2.conf
@@ -1,0 +1,1 @@
+# CONFIG_JFFS2_FS is not set
--- a/qemu.mk
+++ b/qemu.mk
@@ -79,12 +79,13 @@ busybox-cleaner: busybox-cleaner-common
 ################################################################################
 # Linux kernel
 ################################################################################
-$(LINUX_PATH)/.config: kconfig_preempt.conf
+KCONFIGS := kconfig_preempt.conf kconfig_no_jffs2.conf
+$(LINUX_PATH)/.config: $(KCONFIGS)
 	# Temporary fix until we have the driver integrated in the kernel
 	sed -i '/config ARM$$/a select DMA_SHARED_BUFFER' $(LINUX_PATH)/arch/arm/Kconfig;
 	cd $(LINUX_PATH) && ARCH=arm scripts/kconfig/merge_config.sh \
 		arch/arm/configs/vexpress_defconfig \
-		$(CURDIR)/kconfig_preempt.conf
+		$(patsubst %,$(CURDIR)/%,$(KCONFIGS))
 
 linux-defconfig: $(LINUX_PATH)/.config
 


kconfig_no_jffs2.conf
```diff
@@ -0,0 +1 @@
+# CONFIG_JFFS2_FS is not set
```
jenswi-linaro: Does this really unset CONFIG_JFFS2_FS?

jforissier: Yes it does. For the record... here is how it works.

[merge_config.sh](https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/scripts/kconfig/merge_config.sh) reads all the supplied files, and considers the lines that match "CONFIG_something =" and "# CONFIG_something" (with exactly *one* space between the hash sign and the config name!). It merges all these values, reporting conflicts as it goes, into a unique file.

Then, this file is processed by "make alldefconfig" which will copy the supplied definitions into the final `.config` file as long as dependencies are not violated. In any case, `merge_config.sh` will report any discrepancy between what was requested and what is actually used. Variables that are unspecified take their default values (alldefconfig) but it is also possible to ask for them to be disabled (allnoconfig, `merge_config.sh -n`).

Let's compare the final .config with or without this `kconfig_no_jffs2.conf`:
```
# Use only the "preempt" fragment
$ rm -f ../linux/.config ; make linux-defconfig KCONFIGS=kconfig_preempt.conf >/dev/null
$ grep CONFIG_JFFS2_FS ../linux/.config
CONFIG_JFFS2_FS=y
CONFIG_JFFS2_FS_DEBUG=0
CONFIG_JFFS2_FS_WRITEBUFFER=y
# CONFIG_JFFS2_FS_WBUF_VERIFY is not set
# CONFIG_JFFS2_FS_XATTR is not set
$

# Use both the "preempt" and "no_jffs2" fragments
$ rm -f ../linux/.config ; make linux-defconfig >/dev/null
$ grep CONFIG_JFFS2_FS ../linux/.config
# CONFIG_JFFS2_FS is not set
$ 
```

jenswi-linaro: OK, thanks for the explanation. :-)

jenswi-linaro:
```
```

pascal-brand-st-dev:
Reviewed-by: Pascal Brand <pascal.brand@linaro.org>

jbech-linaro:
I'm OK with the changes, but since we now have two .conf files for kernel flags, can we:
a) Put both of them in a single file instead.
b) Create a subfolder where we put patches, configs etc, so so we don't clutter down the top level directory too much?

Also, the DMA flag we set today using sed, couldn't that also be included in such a common and extended kernel config? 

jforissier:
@jbech-linaro 
a) Yes, a single file for both would do. More or less it means we would have one config fragment file per platform (assuming we handle the DMA stuff likewise). My approach was rather to keep each feature in its own file so that they can be shared by several platforms. For instance, all platforms need `DMA_SHARED_BUFFER`; in addition, QEMU needs `PREEMPT` and "no `JFFS2_FS`". Thinking about it, it may be simpler to just have one file per platform.
b) Yes

As for the `DMA_SHARED_BUFFER` flag: this one cannot be directly handled by `merge_config.sh`, because it is a "hidden" option, i.e., there is no description text after "bool" in `linux/drivers/base/Kconfig` which means it is not user-selectable. Such configs are ignored by `make KCONFIG_ALLCONFIG=<input_file> alldefconfig` (I spent quite some time trying to figure this out!).
But... there is a workaround: just enable a feature that depends on `DMA_SHARED_BUFFER`, such as `DRM` for instance.

I will not change the current patch since we have 3 R-b tags, but I will propose a new one.

jbech-linaro:
Thanks @jforissier , let's keep this in the back of our head, and yes, you don't have to make any additional changes to this patch right now.

jforissier:
Review tags applied.

PR #32, opened 2015-11-26
Use common rule for linux/.config and per-platform config fragments

jforissier:
- Kernel config fragments that were used by the QEMU makefile are
merged into kconfigs/qemu.conf. Now, each platform has its own file
in kconfigs/
- All platforms: CONFIG_DMA_SHARED_BUFFER cannot be enabled directly
because it is  a not a visible symbol (has no prompt in Kconfig).
CONFIG_DRM is set instead which depends on DMA.
- kconfigs/hikey.conf: keep only what is necessary


jforissier:
This is a follow-up to https://github.com/OP-TEE/build/pull/30.

jbech-linaro:
```
Reviewed-by: Joakim Bech <joakim.bech@linaro.org>
Tested-by: Joakim Bech <joakim.bech@linaro.org> (QEMU, MTK8173)
```

PR #31, opened 2015-11-26
common: Remove *_CLEAN_COMMON_FLAGS dependency on *_COMMON_FLAGS

vchong:
Fixes https://github.com/OP-TEE/build/issues/28.

In common.mk *_CLEAN_COMMON_FLAGS ?= *_COMMON_FLAGS but in the
platform makefiles (e.g. hikey.mk) they're defined as
such:
*_COMMON_FLAGS += x y z
*_CLEAN_COMMON_FLAGS += x

When expanded during a build, *_CLEAN_COMMON_FLAGS becomes
"x y z x". Variables can be repeated, which may be nothing more
than a cosmetic issue, but if any of those variables are make
targets, then 'make clean' will build them before cleaning.

To fix this, we remove the dependency of *_CLEAN_COMMON_FLAGS on
*_COMMON_FLAGS.

Signed-off-by: Victor Chong <victor.chong@linaro.org>
Tested-by: Victor Chong <victor.chong@linaro.org> (hikey)

PR #33, opened 2015-12-02
hikey: Update build instructions

vchong:
@jbech-linaro @jforissier I've tested about 10+ consecutive boots without problems so this seems stable at least from a booting point of view I suppose. If you can test (both booting and xtest) on your board that would be helpful as well.

```
Result of testsuite XTEST_TEE_TEST:  
25953 subtests of which 0 failed                                                                                       
59 test cases of which 0 failed                                                                                        
0 test case was skipped                                                                                                
TEE test application done!
```

The first commit also replaces #31.

### DIFF
--- a/hikey.mk
+++ b/hikey.mk
@@ -1,12 +1,30 @@
+NSU ?= 64
+SU ?= 32
+SK ?= 64
+
 -include common.mk
 
 ################################################################################
 # Mandatory definition to use common.mk
 ################################################################################
+ifeq ($(NSU),64)
 CROSS_COMPILE_NS_USER		?= "$(CCACHE)$(AARCH64_CROSS_COMPILE)"
+MULTIARCH			:= aarch64-linux-gnu
+else
+CROSS_COMPILE_NS_USER		?= "$(CCACHE)$(AARCH32_CROSS_COMPILE)"
+MULTIARCH			:= arm-linux-gnueabihf
+endif
 CROSS_COMPILE_NS_KERNEL		?= "$(CCACHE)$(AARCH64_CROSS_COMPILE)"
+ifeq ($(SU),64)
+CROSS_COMPILE_S_USER		?= "$(CCACHE)$(AARCH64_CROSS_COMPILE)"
+else
 CROSS_COMPILE_S_USER		?= "$(CCACHE)$(AARCH32_CROSS_COMPILE)"
+endif
+ifeq ($(SK),64)
 CROSS_COMPILE_S_KERNEL		?= "$(CCACHE)$(AARCH64_CROSS_COMPILE)"
+else
+CROSS_COMPILE_S_KERNEL		?= "$(CCACHE)$(AARCH32_CROSS_COMPILE)"
+endif
 OPTEE_OS_BIN 			?= $(OPTEE_OS_PATH)/out/arm-plat-hikey/core/tee.bin
 OPTEE_OS_TA_DEV_KIT_DIR		?= $(OPTEE_OS_PATH)/out/arm-plat-hikey/export-user_ta
 
@@ -18,9 +36,6 @@ ifeq ($(DEBUG),1)
 ARM_TF_BUILD			?= debug
 else
 ARM_TF_BUILD			?= release
-# In case user types something like 'make DEBUG=nonsensical ..',
-# we default to release build
-DEBUG				= 0
 endif
 
 EDK2_PATH 			?= $(ROOT)/edk2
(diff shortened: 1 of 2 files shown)


vchong:
Works with https://github.com/OP-TEE/manifest/pull/10.

vchong:
Most of this is based on https://github.com/linaro-swg/hikey_optee/blob/master/Makefile.

jbech-linaro:
@vchong If I go to the office tomorrow (which is the plan and where I have my HiKey), I'll test it.

jbech-linaro:
I've build using OP-TEE/manifest#10 and using this patch. Everythings seems to builds fine and I can flash it without any problem. But when I try to boot the device it stops at:

```
...
DBG TEE-CORE:generic_boot_cpu_on_handler:364: cpu 7: a0 0x0
DBG TEE-CORE:init_secondary_helper:352: Secondary CPU Switching to normal world boot
```
Then nothing more, one of the green LED's are blinking so it's probably still alive, but nothing more is happening on the UART.

vchong:
@jbech-linaro Do you have access to UART3, i.e. the one on the LS expansion port?

jbech-linaro:
@vchong, nope, I'm still using UART0 (i.e, no sonic screwdriver). Basically you're indicating that this -> https://github.com/OP-TEE/optee_os/pull/568 is the reason?

jbech-linaro:
I rebuilt using `make CFG_HIKEY_UART3=n all` and now our generated files says:
```
CFG_HIKEY_UART3 is not set ('n')
```
But still nothing on UART0 after switching to normal world.

vchong:
96boards repos now default to UART3 for ARM-TF (except BL1), EDK2 and Linux, hence https://github.com/linaro-swg/gen_rootfs/pull/6. https://github.com/OP-TEE/optee_os/pull/568 allows the use of UART3, but UART0 is still the default for optee_os, to keep SW logs apart from NW logs.

Please change the following and rebuild to use UART0.

ARM-TF:
Add "CONSOLE_BASE=PL011_UART0_BASE CRASH_CONSOLE_BASE=PL011_UART0_BASE" to ARM_TF_FLAGS.

EDK2:
Add "EDK2_MACROS="-DSERIAL_BASE=0xF8015000" to the end of edk2-call

Linux:
Change ttyAMA3 to ttyAMA0 in ```patches_hikey/grub/grub.cfg```.

I will push a change to have a flag defined to specify either UART0 and UART3.

jbech-linaro:
@vchong it's working fine after I did what you suggested. I'm OK with merging this as it is and later create a separate patch that makes the switch from UART3 (default) to UART0 a bit easier.

vchong:
Ok thanks. I'll send a separate patch after this is merged, but will need another reviewer?

hikey.mk
```diff
@@ -123,26 +138,33 @@ edk2-clean: edk2-clean-common
 ################################################################################
 LINUX_DEFCONFIG_COMMON_ARCH ?= arm64
 LINUX_DEFCONFIG_COMMON_FILES ?= $(LINUX_PATH)/arch/arm64/configs/defconfig \
-				$(CURDIR)/kconfigs/hikey.conf
+				$(CURDIR)/kconfigs/hikey.conf \
+				$(PATCHES_PATH)/kernel_config/usb_net_dm9601.conf \
+				$(PATCHES_PATH)/kernel_config/ftrace.conf
```
jforissier: Are these additional `.conf` really useful and worth adding here?

hikey.mk
```diff
@@ -123,26 +138,33 @@ edk2-clean: edk2-clean-common
...
-LINUX_COMMON_FLAGS += ARCH=arm64 Image modules dtbs
+LINUX_COMMON_FLAGS += ARCH=arm64 Image modules
+UPSTREAM_KERNEL = 0
+ifeq ($(UPSTREAM_KERNEL),0)
+LINUX_COMMON_FLAGS += hi6220-hikey.dtb
+DTB = $(LINUX_PATH)/arch/arm64/boot/dts/hi6220-hikey.dtb
+else
+LINUX_COMMON_FLAGS += hisilicon/hi6220-hikey.dtb
+DTB = $(LINUX_PATH)/arch/arm64/boot/dts/hisilicon/hi6220-hikey.dtb
+endif
```
jforissier: That's OK but this would be even better:
```
UPSTREAM_KERNEL := $(if $(wildcard $(LINUX_PATH)/arch/arm64/boot/dts/hisilicon/hi6220-hikey.dts),1,0)
```

hikey.mk
```diff
@@ -179,123 +206,140 @@ optee-linuxdriver-clean: optee-linuxdriver-clean-common
...
+	./configure --target=aarch64 --enable-boot-time $(grub-flags); \
+	$(MAKE) -C $(GRUB_PATH); \
+	./grub-mkimage \
+		--verbose \
+		--output=$(OUT_PATH)/grubaa64.efi \
+		--config=$(PATCHES_PATH)/grub/grub.configfile \
+		--format=arm64-efi \
+		--directory=grub-core \
+		--prefix=/boot/grub \
+		boot chain configfile efinet ext2 fat gettext help linux loadenv lsefi normal part_gpt part_msdos read search search_fs_file search_fs_uuid search_label terminal terminfo tftp time
```
jforissier: You probably don't need all those modules, although it won't hurt to have them all. FYI, I'm using only `configfile fat linux normal help part_gpt`.

hikey.mk
```diff
@@ -32,38 +47,34 @@ EDK2_BIN 			?= $(EDK2_PATH)/Build/HiKey/RELEASE_GCC49/FV/BL33_AP_UEFI.fd
...
-MCUIMAGE_BIN			?=$(ROOT)/out/mcuimage.bin
-USBNETSH_PATH			?=$(ROOT)/out/usbnet.sh
+MCUIMAGE_BIN			?=$(EDK2_PATH)/HisiPkg/HiKeyPkg/NonFree/mcuimage.bin
 STRACE_PATH			?=$(ROOT)/strace
 BOOT_IMG			?=$(ROOT)/out/boot-fat.uefi.img
 LLOADER_PATH			?=$(ROOT)/l-loader
 NVME_IMG			?=$(ROOT)/out/nvme.img
+OUT_PATH			?=$(ROOT)/out
+GRUB_PATH			?=$(ROOT)/grub
+PATCHES_PATH			?=$(ROOT)/patches_hikey
```
jforissier: Why yet another Git tree? Add this to `$(ROOT)/build/hikey` maybe? (not sure it makes sense in the repo model)

hikey.mk
```diff
@@ -351,7 +379,7 @@ nvme-cleaner:
 .PHONY: flash
 flash:
 	python $(ROOT)/burn-boot/hisi-idt.py --img1=$(LLOADER_PATH)/l-loader.bin
```
jforissier: You need `sudo` here or else one needs to do `sudo make flash` (in which case some more code than necessary is run as root...).

jforissier:
I consider that my review comments are minor, the build works well in any case so feel free to keep the PR as-is and add my:

I could not test fully due to the UART0 vs. UART3 issue, but the board boots and looks healthy (heartbeat LED does blink!).

hikey.mk
vchong: In hikey_optee, ftrace.conf is commented out but usb_net_dm9601.conf is actually included. I figured it's probably easier for the users to enable everything and have them use it if needed or ignore if not. Easier than changing makefile before building perhaps..

hikey.mk
vchong: Yes thanks! This makes it automatic. Will change.

hikey.mk
vchong: Got the list from B&B team's build script, but yes, figured it won't hurt. There were actually only 6 modules too originally, but the updated grub.configfile is using the search* modules, so it seems better to just include all of them, in case more modules need to be used later, then no changes to the build are needed.

hikey.mk
vchong: For this, I considered 3 choices:

1. Originally the required patch/helper files were included in this makefile itself, i.e. created using echo commands. As more files are required, the makefile becomes really long and looks cluttered, so it does not seem to scale.

2. Add the patch/helper files to this repo under a ```hikey``` dir as you'd mentioned, but there's already a ```kconfigs``` dir here that contains ```hikey.conf``` among other things, so this seems like it might clutter up the repo as well and we'll have to go to multiple locations/paths for the files.

3. Yea another Git tree. This was chosen as it was already done for the mtk platform, and I'd already gone down this path before ```kconfigs``` was created :s.

When the ```kconfigs``` PR (32) was proposed, I thought I could maybe utilize it using some structure like below:
```
build/patches
    |-- hikey
        |-- dir1
        |-- dir1
    |-- mtk
        |-- dir1
        |-- dir2
```
OR
```
build
    |-- hikey
        |-- patches
            |-- dir1
            |-- dir2
    |-- mtk
        |-- patches
            |-- dir1
            |-- dir2
```
This would be similar to 2 above and more generic, but the PR only added a ```kconfigs``` dir, so I couldn't really fit all the patch/helper files there.

Perhaps 3 can be something to consider in the future?

hikey.mk
vchong: Ok. Will fix.

vchong:
Comments addressed, rebased and tags applied.
Ok for merge?

PR #36, opened 2015-12-09
hikey: make console configurable

vchong:
Signed-off-by: Victor Chong victor.chong@linaro.org
Tested-by: Victor Chong <victor.chong@linaro.org> (hikey)

### DIFF
--- a/hikey.mk
+++ b/hikey.mk
@@ -1,6 +1,7 @@
 ################################################################################
 # User-defined variables
 # Edit so these match your target
+# NOTE: If making changes after a build, please clean before rebuilding!
 ################################################################################
 # Non-secure user mode (root fs binaries): 32 or 64-bit [default 64]
 NSU ?= 64
@@ -8,6 +9,10 @@ NSU ?= 64
 SK ?= 64
 # Secure user mode (Trusted Apps): 32 or 64-bit [default 32, requires SK=64 for 64]
 SU ?= 32
+
+# Normal/secure world console UARTs: 3 or 0 [default 3]
+CFG_NW_CONSOLE_UART ?= 3
+CFG_SW_CONSOLE_UART ?= 3
 
 ################################################################################
 # Includes
@@ -107,6 +112,12 @@ ARM_TF_FLAGS ?= \
 	PLAT=hikey \
 	SPD=opteed
 
+ARM_TF_CONSOLE_UART ?= $(CFG_SW_CONSOLE_UART)
+ifeq ($(ARM_TF_CONSOLE_UART),0)
+	ARM_TF_FLAGS += CONSOLE_BASE=PL011_UART0_BASE \
+			CRASH_CONSOLE_BASE=PL011_UART0_BASE
+endif
+
 arm-tf: optee-os edk2
 	$(ARM_TF_EXPORTS) $(MAKE) -C $(ARM_TF_PATH) $(ARM_TF_FLAGS) all fip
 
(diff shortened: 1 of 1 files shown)


vchong:
Result of testsuite XTEST_TEE_TEST:
25953 subtests of which 0 failed                                                                                       
59 test cases of which 0 failed                                                                                        
0 test case was skipped                                                                                                
TEE test application done!

hikey.mk
```diff
@@ -1,12 +1,35 @@
...
+################################################################################
+# Non-secure user mode (root fs binaries): 32 or 64-bit [default 64]
 NSU ?= 64
-SU ?= 32
+# Secure kernel (OP-TEE OS): 32 or 64-bit [default 64]
 SK ?= 64
+# Secure user mode (Trusted Apps): 32 or 64-bit [default 32, requires SK=64 for 64]
+SU ?= 32
+
+# Console UARTs: 3 or 0
```
jbech-linaro: I don't really like the idea of setting all this individually. I mean either you set it to be UART0 or UART3? In the end I would like to, from make, call:
```
make
# Building with both NW and SW at UART0 (default)
```
And ...
```
make CFG_HIKEY_UART3=y
# Building with NW UART3 and SW at UART0
```
This approach will hardcode the secure side to UART0, but do we think we will use something else? I.e, is there any need for making that configurable using a make argument?

So, what about using a global FLAG something like this ... psuedo code
```
CFG_HIKEY_UART ?= 0
ARM_TF_CONSOLE_UART ?= $CFG_HIKEY_UART
OPTEE_CONSOLE_UART ?= 0
EDK2_CONSOLE_UART ?= $CFG_HIKEY_UART
LINUX_CONSOLE_UART ?= $CFG_HIKEY_UART
```

Also I probably would like to move the different ARM_..., OPTEE_..., EDK2_... and LINUX_... down to the section about the component itself.

jbech-linaro: Talked to Victor about it, most common setup is most likely [sonic screwdriver](https://www.96boards.org/products/mezzanine/uarts/) (UART3), since the other option UART0, requires the 4-pin connector that has to be soldered and also you need the special 1.8V cable. Therefore we decided to use UART3 as the default. ARM-TF + OP-TEE will be tied to the SW flag, whilst EDK2 and LINUX will be tied to the NW flag.

vchong: Ok. So here's the plan:
1. Submit PR to optee_os to make default UART 3 instead of 0.
2. 
 ```
 make
 # Building with both NW and SW at UART3 (default)
 
 make CFG_HIKEY_UART=0
 # Building with NW (EDK2 and LINUX) UART3 and SW (ARM-TF and optee_os) at UART0
 ```
3.
 ```
 CFG_HIKEY_UART ?= 3
 ARM_TF_CONSOLE_UART ?= $CFG_HIKEY_UART
 OPTEE_CONSOLE_UART ?= $CFG_HIKEY_UART
 EDK2_CONSOLE_UART ?= 3
 LINUX_CONSOLE_UART ?= 3
 ```
 The different ARM_..., OPTEE_..., EDK2_... and LINUX_... will be moved down to the section about the component itself.

This approach will 'hard' code the normal side to UART3, but the more advanced users can still configure each component to the UART of their choosing if they wish.

Any objections?

vchong:
https://github.com/OP-TEE/optee_os/pull/582 created.

jbech-linaro:
Besides Jerome's comment in OP-TEE/optee_os#582 (which also affects this PR).

vchong:
Comments addressed, rebased and tags applied.
Please merge. Thanks!

> Maybe "Normal/secure world console UARTs" (the names are good, however the first time I read them > I thought "what are the northwest and southwest consoles?" ;-)).
> Would it help to mention the physical location of 0 and 3 on the board?
>
> Reviewed-by: Jerome Forissier <jerome.forissier@linaro.org>

@jforissier The silk screen names on the board differs between the circuitco and lemaker version, and if using the uart adapter, we'll need those locations and/or the LS connector pin locations, and not to mention jumper settings as well. Seem like a lot of extra info to add to a makefile so perhaps best to refer to the wiki?

jforissier:
@vchong agreed

PR #40, opened 2016-01-18
HiKey: Add recovery target and instructions

jbech-linaro:
To make it a bit easier to understand how to do a recovery and how to
flash HiKey, we have added a new make target for recovery which tells
how to configure the jumpers and the initiates the recovery itself. In
addition to this we have changed the "flash" make target so that it
does the same as stated in the instructions on the 96boards "Getting
started" page.

Tested-by: Joakim Bech <joakim.bech@linaro.org> (HiKey)

### DIFF
--- a/hikey.mk
+++ b/hikey.mk
@@ -420,9 +420,32 @@ nvme-cleaner:
 ################################################################################
 # Flash
 ################################################################################
+define flash_help
+	@read -r -p "1. Connect USB OTG cable, the micro USB cable (press any key)" dummy
+	@read -r -p "2. Connect HiKey to power up (press any key)" dummy
+endef
+
+.PHONY: recovery
+recovery:
+	@echo "Enter recovery mode to flash a new bootloader"
+	@echo "Jumper 1-2: Closed (Auto power up = Boot up when power is applied)"
+	@echo "       3-4: Closed (Boot Select = Recovery: program eMMC from USB OTG)"
+	$(call flash_help)
+	sudo python $(ROOT)/burn-boot/hisi-idt.py --img1=$(LLOADER_PATH)/l-loader.bin
+	@$(MAKE) --no-print flash FROM_RECOVERY=1
+
 .PHONY: flash
 flash:
-	sudo python $(ROOT)/burn-boot/hisi-idt.py --img1=$(LLOADER_PATH)/l-loader.bin
+ifneq ($(FROM_RECOVERY),1)
+	@echo "Flash binaries using fastboot"
+	@echo "Jumper 1-2: Closed (Auto power up = Boot up when power is applied)"
+	@echo "       3-4: Open   (Boot Select = Boot from eMMC)"
+	@echo "       5-6: Closed (GPIO3-1 = Low: UEFI runs Fastboot app)"
+	$(call flash_help)
+	@echo "3. Wait until you see the (UART) message"
+	@echo "    \"Android Fastboot mode - version x.x Press any key to quit.\""
+	@read -r -p "   Then press any key to continue flashing" dummy
+endif
 	fastboot flash ptable $(LLOADER_PATH)/ptable-linux-4g.img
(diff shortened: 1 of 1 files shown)


hikey.mk
```diff
@@ -420,9 +420,33 @@ nvme-cleaner:
...
+	@read -r -p "1. Connect USB OTG cable, the micro USB cable (press any key)" dummy
+	@read -r -p "2. Connect HiKey to power up (press any key)" dummy
+endef
+
+.PHONY: recovery
+recovery:
+	@echo "Enter recovery mode to flash a new bootloader"
+	@echo "Jumper 1-2: Closed (Auto power up)"
+	@echo "       3-4: Closed (Boot Select)"
+	@echo "       5-6: Open   (GPIO3-1)"
```
vchong: I know the getting started and user guide both says GPIO3-1, but for our purposes here, isn't it more descriptive to say or at least add the wording 'Fastboot Mode'? Also for Boot Select, maybe say or add 'Recovery Mode'?

hikey.mk
```diff
@@ -420,9 +420,33 @@ nvme-cleaner:
...
 flash:
-	sudo python $(ROOT)/burn-boot/hisi-idt.py --img1=$(LLOADER_PATH)/l-loader.bin
+ifneq ($(FROM_RECOVERY),1)
+	@echo "Flash binaries using fastboot"
+	@echo "Jumper 1-2: Closed (Auto power up)"
+	@echo "       3-4: Open   (Boot Select)"
+	@echo "       5-6: Closed (GPIO3-1)"
+	$(call flash_help)
+	@echo "3. Wait until you see the (UART) message"
+	@echo "    \"Android Fastboot mode - version x.x Press any key to quit.\""
```
vchong: Same as above.

vchong:
With or without my comments.

```Reviewed-by: Victor Chong <victor.chong@linaro.org>```

File: hikey.mk (hunk 1 above)
Comment (jbech-linaro) [inline hikey.mk]: I totally agree that GPIO3-1 doesn't say much, but I just copy/pasted from the docs so that we don't deviate from their nomenclature. You will also find the GPIO3-1 in HiKey user's guide (i.e, not only the "getting started guide"), so I'd like to keep this name here. I could expland it by adding "fastboot" to it also, but .. you're also getting into fastboot when you're in recovery mode, hence it could be a bit confusing I think and confusion is the last thing we need with this already tricky (according to me) way of dealing with various flash modes, cables and various time outs.

Comment (jforissier) [inline hikey.mk]: @vchong +1
Like you, at first I read the text in parentheses as the meaning of the "Closed" or "Open" setting, while in fact it is the purpose of the jumper. What about adding both information?
```
Jumper 1-2: Closed (Auto power = Boot up when power is applied)
       3-4: Closed (Boot select = Recovery: program eMMC from USB OTG)
```
Regarding jumper 5-6, I think its state is irrelevant here (it is checked by UEFI, not by the boot ROM nor by l-loader.bin).

File: hikey.mk (hunk 2 above)
Comment (jforissier) [inline hikey.mk]: This maybe?
```
Jumper 1-2: Closed (Auto power = Boot up when power is applied)
       3-4: Open   (Boot select = Boot from eMMC)
       5-6: Closed (GPIO3-1 = Low: UEFI runs Fastboot app)
```

hikey.mk
jbech-linaro: @jforissier , you're right, it doesn't seem like it matter whether 5-6 are open or closed, you end up in recovery mode if 3-4 is closed regardless of the state of 5-6. I wonder why the explicitly tell that this should be open in the getting started. I think I'll remove that since it seems to be irrelevant and I can change according to your suggestion.

hikey.mk
jbech-linaro: OK, I'll send an update.

jbech-linaro:
@jforissier is this OK from your side also?

jforissier:

jbech-linaro:
Rebased, tags has been applied.

PR #47, opened 2016-03-01
fvp: possibility to choose SU and SK compilation mode

pascal-brand-st-dev:
Signed-off-by: Pascal Brand <pascal.brand@st.com>

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -26,6 +26,98 @@ OPTEE_TEST_OUT_PATH 		?= $(ROOT)/optee_test/out
 
 CCACHE ?= $(shell which ccache) # Don't remove this comment (space is needed)
 
+
+################################################################################
+# Check coherency of compilation mode
+################################################################################
+
+ifneq ($(COMPILE_NS_USER),)
+ifeq ($(COMPILE_NS_KERNEL),)
+$(error COMPILE_NS_KERNEL must be defined as COMPILE_NS_USER=$(COMPILE_NS_USER) is defined)
+endif
+ifeq (,$(filter $(COMPILE_NS_USER),32 64))
+$(error COMPILE_NS_USER=$(COMPILE_NS_USER) - Should be 32 or 64)
+endif
+endif
+
+ifneq ($(COMPILE_NS_KERNEL),)
+ifeq ($(COMPILE_NS_USER),)
+$(error COMPILE_NS_USER must be defined as COMPILE_NS_KERNEL=$(COMPILE_NS_KERNEL) is defined)
+endif
+ifeq (,$(filter $(COMPILE_NS_KERNEL),32 64))
+$(error COMPILE_NS_KERNEL=$(COMPILE_NS_KERNEL) - Should be 32 or 64)
+endif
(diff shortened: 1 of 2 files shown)


common.mk
```diff
@@ -26,6 +26,98 @@ OPTEE_TEST_OUT_PATH 		?= $(ROOT)/optee_test/out
 
 CCACHE ?= $(shell which ccache) # Don't remove this comment (space is needed)
 
+
+################################################################################
+# Check coherency of compilation mode
+################################################################################
+
+ifneq ($(COMPILE_NS_USER),)
```
jbech-linaro: All good below, but since this is the common.mk file I wonder if we can merge this as it is without affecting the other platforms supported i build.git? (Not a statement, it's a question).

fvp.mk
```diff
@@ -1,25 +1,25 @@
...
+# S_USER (Secure User - TA) are compiled
 ################################################################################
-CROSS_COMPILE_NS_USER	?= "$(CCACHE)$(AARCH64_CROSS_COMPILE)"
-CROSS_COMPILE_NS_KERNEL	?= "$(CCACHE)$(AARCH64_CROSS_COMPILE)"
-CROSS_COMPILE_S_USER	?= "$(CCACHE)$(AARCH64_CROSS_COMPILE)"
-CROSS_COMPILE_S_KERNEL	?= "$(CCACHE)$(AARCH64_CROSS_COMPILE)"
-OPTEE_OS_BIN		?= $(OPTEE_OS_PATH)/out/arm-plat-vexpress/core/tee.bin
-OPTEE_OS_TA_DEV_KIT_DIR	?= $(OPTEE_OS_PATH)/out/arm-plat-vexpress/export-ta_arm64
+
+COMPILE_NS_USER   := 64
```
jbech-linaro: We will probably always run `NS_/{USER/KERNEL}` in 64bit mode for FVP, but why not make those two "` FOO ?= XX`" also?

jbech-linaro:
Commit message needs to be update to more clearly state what it is about, it's hard to understand what SU and SK means if you're unaware of how this works.

fvp.mk
pascal-brand-st-dev: NS_USER=32 is not working right now. So I would like to keep it with :=

common.mk
pascal-brand-st-dev: It should be backward compatible. I'll check on QEMU to see if it works or not.

fvp.mk
jbech-linaro: Ok

jbech-linaro:
After checking that it's working with QEMU.

pascal-brand-st-dev:
Depends on https://github.com/OP-TEE/optee_test/pull/72 to compile with `COMPILE_S_KERNEL=32` and `COMPILE_S_USER=32`.

    Tested-by: Pascal Brand <pascal.brand@linaro.org> (QEMU)
    Tested-by: Pascal Brand <pascal.brand@linaro.org> (FVP)

PR #48, opened 2016-03-04
Use mix compilation mode on all platforms

pascal-brand-st-dev:
Signed-off-by: Pascal Brand <pascal.brand@st.com>

### DIFF
--- a/hikey.mk
+++ b/hikey.mk
@@ -1,14 +1,12 @@
 ################################################################################
-# User-defined variables
-# Edit so these match your target
-# NOTE: If making changes after a build, please clean before rebuilding!
+# Following variables defines how the NS_USER (Non Secure User - Client
+# Application), NS_KERNEL (Non Secure Kernel), S_KERNEL (Secure Kernel) and
+# S_USER (Secure User - TA) are compiled
 ################################################################################
-# Non-secure user mode (root fs binaries): 32 or 64-bit [default 64]
-NSU ?= 64
-# Secure kernel (OP-TEE OS): 32 or 64-bit [default 64]
-SK ?= 64
-# Secure user mode (Trusted Apps): 32 or 64-bit [default 32, requires SK=64 for 64]
-SU ?= 32
+COMPILE_NS_USER   ?= 64
+override COMPILE_NS_KERNEL := 64
+COMPILE_S_USER    ?= 32
+COMPILE_S_KERNEL  ?= 64
 
 # Normal/secure world console UARTs: 3 or 0 [default 3]
 CFG_NW_CONSOLE_UART ?= 3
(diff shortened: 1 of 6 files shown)


jbech-linaro:

hikey.mk
```diff
@@ -1,14 +1,12 @@
...
+# S_USER (Secure User - TA) are compiled
 ################################################################################
-# Non-secure user mode (root fs binaries): 32 or 64-bit [default 64]
-NSU ?= 64
-# Secure kernel (OP-TEE OS): 32 or 64-bit [default 64]
-SK ?= 64
-# Secure user mode (Trusted Apps): 32 or 64-bit [default 32, requires SK=64 for 64]
-SU ?= 32
+COMPILE_NS_USER   ?= 64
+COMPILE_NS_KERNEL := 64
```
jforissier: Any particular reason not to be using `?=` everywhere? (same in other mk files)

vchong: Per Pascal @ https://github.com/OP-TEE/build/pull/47
"NS_USER=32 is not working right now. So I would like to keep it with :="

vchong: Sorry wrong comment. Please ignore.

pascal-brand-st-dev: I wanted to leave the choice (that is `?=`) when the feature has been tested, and force it (that is `:=`) when not tested or not supported (qemu in 64bits for example, or hikey with Linux compiled in 32bits).

jforissier: OK, I agree we don't want `?=` here. But `:=` won't force the value if it's provided as an argument to `make` (as opposed to via an environment variable). For this to happen you need `override COMPILE_NS_KERNEL := 64`.

pascal-brand-st-dev:
Updated

jforissier:
I got build errors with HiKey and `make COMPILE_NS_USER=32`. It looks like the `strace` configure arguments are wrong: https://github.com/OP-TEE/build/blame/master/hikey.mk#L292-L293. Also, after `busybox` is built an invalid path is used:
```
[...]
make[1]: Leaving directory '/home/jerome/work/optee_repo_hikey/busybox'
find: `/home/jerome/work/optee_repo_hikey/build/../toolchains/aarch32/aarch64-linux-gnu/libc/lib': No such file or directory
common.mk:135: recipe for target 'busybox-common' failed
make: *** [busybox-common] Error 1
```

pascal-brand-st-dev:
Updated

jforissier:
The following patch fixes the non-secure user 32-bit mode for HiKey, i.e. `make COMPILE_NS_USER=32`. Compile-tested + checked that output files have the expected architecture with the `file` command.
```
diff --git a/hikey.mk b/hikey.mk
index 9167989..dc18b3a 100644
--- a/hikey.mk
+++ b/hikey.mk
@@ -3,7 +3,7 @@
 # Application), NS_KERNEL (Non Secure Kernel), S_KERNEL (Secure Kernel) and
 # S_USER (Secure User - TA) are compiled
-override COMPILE_NS_USER   := 64
+COMPILE_NS_USER   ?= 64
 override COMPILE_NS_KERNEL := 64
 COMPILE_S_USER    ?= 32
 COMPILE_S_KERNEL  ?= 64
@@ -105,10 +105,13 @@ arm-tf-clean:
 # Busybox
-BUSYBOX_COMMON_TARGET = hikey nocpio
 BUSYBOX_CLEAN_COMMON_TARGET = hikey clean
 
-busybox: busybox-common
+busybox:
+       cd $(GEN_ROOTFS_PATH) && \
+               export CROSS_COMPILE="$(AARCH$(COMPILE_NS_USER)_CROSS_COMPILE)" \
+               PATH=${PATH}:$(LINUX_PATH)/usr ; \
+               $(GEN_ROOTFS_PATH)/generate-cpio-rootfs.sh hikey nocpio
 
 .PHONY: busybox-clean
 busybox-clean: busybox-clean-common
@@ -252,8 +255,8 @@ strace:
        cd $(STRACE_PATH); \
        ./bootstrap; \
        set -e; \
-       ./configure --host=$(MULTIARCH) CC="$(CCACHE)$(AARCH64_CROSS_COMPILE)gcc" LD=$(AARCH64_CROSS_COMPILE)ld; \
-       CC="$(CCACHE)$(AARCH64_CROSS_COMPILE)gcc" LD=$(AARCH64_CROSS_COMPILE)ld $(MAKE) -C $(STRACE_PATH)
+       ./configure --host=$(MULTIARCH) CC="$(CCACHE)$(AARCH$(COMPILE_NS_USER)_CROSS_COMPILE)gcc" LD=$(AARCH$(COMPILE_NS_USER)_CROSS_COMPILE)ld; \
+       CC="$(CCACHE)$(AARCH$(COMPILE_NS_USER)_CROSS_COMPILE)gcc" LD=$(AARCH$(COMPILE_NS_USER)_CROSS_COMPILE)ld $(MAKE) -C $(STRACE_PATH)
 
 .PHONY: strace-clean
 strace-clean:
```

pascal-brand-st-dev:
Updated. Depends on https://github.com/linaro-swg/gen_rootfs/pull/10

jforissier:

PR #49, opened 2016-03-15
common: rm optee_os/out dir on clean

vchong:
Several build errors reported recently due to code changes and optee_os/out dir containing outdated files. Hoping this will make the build more robust.


jenswi-linaro:
`rm -rf` is very dangerous, what if `$(OPTEE_OS_PATH)/out` happens to point to something important?

`make clean` should leave the out directory empty, at least of conflicting files, if it doesn't then we should fix that instead of treating the symptom.

PR #52, opened 2016-03-17
common: Set CFG_ARM32=y for xtest when required

vchong:
CFG_ARM32=y must be set when app is compiled in 32-bit mode,
i.e. COMPILE_NS_USER=32.


common.mk
```diff
@@ -228,11 +228,19 @@ XTEST_COMMON_FLAGS ?= CROSS_COMPILE_HOST=$(CROSS_COMPILE_NS_USER)\
 	COMPILE_NS_USER=$(COMPILE_NS_USER) \
 	O=$(OPTEE_TEST_OUT_PATH)
 
+ifeq ($(COMPILE_NS_USER),32)
```
pascal-brand-st-dev: Is it required with https://github.com/OP-TEE/optee_test/pull/72/files?

vchong: Oh nvm then. I think that takes care of it. Sorry for the noise.

PR #54, opened 2016-03-18
mt8173: Fixed booting issue

m943040028:
Signed-off-by: SY Chiu <sy.chiu@linaro.org>
Tested-by: SY Chiu <sy.chiu@linaro.org> (MT8173 EVB)


jbech-linaro:
This fix solves the boot issue, however, there it crashes in the kernel driver, I've not investigated it and I don't intend to do so, since we will switch to the new driver any day.

PR #58, opened 2016-04-02
misc: remove driver modules from filesystem list

vchong:
optee.ko and optee_armtz.ko not required anymore with generic
driver


jenswi-linaro:
If the driver is compiled as a loadable module then it needs to be included in the file list. The only drawback with having it there now is that you get a harmless but perhaps confusing error message.

vchong:
Ah, ok. I will close this then. Thanks!

jforissier:
It would be good to include the modules only when the driver is compiled as a module. We don't want confusing warning messages, do we?

jenswi-linaro:
Agree, that would be the best.

vchong:
Perhaps we can add an if statement to the makefiles to see if the .ko files are there and only include if yes?

vchong:
Then again, there's no guarantee that everyone will use the same path to build the modules?

jenswi-linaro:
If we can assume that they don't change the directory layout in the kernel tree it should be OK.

vchong:
Ok. I'll submit a new PR with 'if' checks.

PR #60, opened 2016-04-18
Debian fs

jbech-linaro:
This  dependent on:
* OP-TEE/manifest/pull/27
* OP-TEE/optee_os/pull/744

It doesn't really affect other builds (besides the nproc thing), so there is no harm merging one before another. It's just that the entire setup won't work until all dependencies has been merged.

Please also read the commit message, that contains a lot of valuable information this time.


common.mk
```diff
@@ -135,7 +136,7 @@ busybox-cleaner-common:
 LINUX_COMMON_FLAGS ?= LOCALVERSION= CROSS_COMPILE=$(CROSS_COMPILE_NS_KERNEL)
 
 linux-common: linux-defconfig
-	$(MAKE) -C $(LINUX_PATH) $(LINUX_COMMON_FLAGS)
+	$(MAKE) -C $(LINUX_PATH) -j$(NPROC) $(LINUX_COMMON_FLAGS)
```
jforissier: This will not play nicely with `make -jN`. It will generate a warning (`make[1]: warning: -jN forced in submake: disabling jobserver mode`) and you will likely end up running more jobs than requested. For example: `make -j1` will still spawn `$(NPROC)` jobs here.

We would need to do `MAKEFLAGS += -j$(shell nproc)` globally, but only if `$(MAKEFLAGS)` does not already contain `-j`, which I can't find a way to detect unfortunately :( at least with GNU make 4.0.

jbech-linaro: We can skip this commit for now, while trying to figure out how to solve this in a proper way. I had a gut feeling that there might be questions about it, so therefore I made sure to not combine that with anything else.

pascal-brand-st-dev:
Without the 1st commit (-jN):
`Reviewed-by: Pascal Brand <pascal.brand@linaro.org>`

d3zd3z:
For the curious (not sure it matters for this), https://github.com/android/platform_build/tree/master/tools/makeparallel is how Google solved this problem for Android builds. This program basically consumes all of the jobs from the job server, and then passes '-jn' to the subinvocation. They use this when invoking 'ninja' which doesn't use the job server, but wants to be able to run parallel builds.

jforissier:
Le 20 avr. 2016 9:25 PM, "David Brown" <notifications@github.com> a écrit :
>
> For the curious (not sure it matters for this),
https://github.com/android/platform_build/tree/master/tools/makeparallel is
how Google solved this problem for Android builds. This program basically
consumes all of the jobs from the job server, and then passes '-jn' to the
subinvocation. They use this when invoking 'ninja' which doesn't use the
job server, but wants to be able to run parallel builds.
>

That's interesting, but not useful for our need since the subinvocation is
$(MAKE) which already knows about the number of jobs available.

Thanks,
-- 
Jerome

hikey_debian.mk
```diff
@@ -0,0 +1,408 @@
...
+DEBPKG_CONTROL_PATH		?= $(DEBPKG_PATH)/DEBIAN
+
+################################################################################
+# Targets
+################################################################################
+all: prepare arm-tf linux boot-img lloader system-img nvme deb
+
+clean: arm-tf-clean edk2-clean linux-clean optee-os-clean optee-client-clean xtest-clean boot-img-clean lloader-clean aes-perf-clean sha-perf-clean
+
+cleaner: clean prepare-cleaner linux-cleaner nvme-cleaner system-img cleaner
```
vchong: system-img? or system-cleaner?
Maybe this is what's causing the rebuild from scratch?

jbech-linaro: Thanks for the catch Victor. This doesn't cause the re-builds, but still this needs to be fixed. I'll fix that when rebasing and applying the tags.

jbech-linaro:
I'll rebase, remove the nproc patch and apply tags. By the way, it's not that the parallel builds doesn't work today, it's working perfectly fine already. The only difference with my patch is that instead of building **all** targets in parallel at the same time, you build target by target, where each individual target builds in parallel. In the end you get a much cleaner build log. But for speed, what we have today is totally fine (even better I guess).

hikey_debian.mk
```diff
@@ -0,0 +1,408 @@
...
+	rm -rf $(SHAPERF_PATH)/out
+
+################################################################################
+# Boot Image
+################################################################################
+.PHONY: boot-img
+boot-img:
+ifeq ("$(wildcard $(BOOT_IMG))","")
+	echo "Downloading Debian HiKey boot image ..."
+	wget https://builds.96boards.org/snapshots/hikey/linaro/debian/410/boot-fat.uefi.img.gz -O $(BOOT_IMG).gz
```
vchong: Instead of 410, why not latest?

vchong: I guess we have to match this with the system img below which doesn't have a 'dynamic' link due to the date in the filename?

vchong: Maybe define BOOT_IMG_URL and SYSTEM_IMG_URL at the top?

hikey_debian.mk
```diff
@@ -0,0 +1,408 @@
...
+ifeq ($(DEVELOPER_EDITION),1)
+	@echo "Downloading Debian root fs (530MB) ..."
+	wget https://builds.96boards.org/snapshots/hikey/linaro/debian/410/hikey-jessie_developer_20160225-410.emmc.img.gz -O $(SYSTEM_IMG).gz
+	gunzip $(SYSTEM_IMG).gz
+else
+	@echo "Since there is a click through license, there is some manual work involved here."
+	@echo "The downloaded image, should be gunzip'd, renamed and placed here (don't change the name of the file)"
+	@echo "  $(SYSTEM_IMG)"
+	@read -r -p "Will spawn a new tab in your browser, press ENTER when ready to continue" dummy
+	@xdg-open https://builds.96boards.org/snapshots/hikey/linaro/debian/410/hikey-jessie_alip_20160225-410.emmc.img.gz
```
vchong: Don't need to rename this to $(SYSTEM_IMG).gz and then gunzip it?

hikey_debian.mk
```diff
@@ -0,0 +1,408 @@
...
+	@read -r -p "   Then press any key to continue flashing" dummy
+endif
+	fastboot flash ptable $(LLOADER_PATH)/ptable-linux-4g.img
+	fastboot flash fastboot $(ARM_TF_PATH)/build/hikey/$(ARM_TF_BUILD)/fip.bin
+	fastboot flash nvme $(NVME_IMG)
+	fastboot flash boot $(BOOT_IMG)
+	fastboot flash system $(SYSTEM_IMG)
+
+.PHONY: flash-fip
+flash-fip:
```
vchong: Add arm-tf here as target dependency? This is similarly done for below 2 targets as well.

hikey_debian.mk
```diff
@@ -0,0 +1,408 @@
...
+.PHONY: help
+help:
+	@echo " 1. WiFi on HiKey debian"
+	@echo " ======================="
+	@echo " Open /etc/network/interfaces and add:"
+	@echo "  auto wlan0"
+	@echo "  	iface wlan0 inet dhcp"
+	@echo " 	wpa-ssid \"my-ssid\""
+	@echo " 	wpa-psk \"my-wifi-password\""
+	@echo " Reboot and you're should have WiFi access"
```
vchong: Nit: s/you're/you

hikey_debian.mk
```diff
@@ -0,0 +1,408 @@
...
+#   IP=111.222.333.444 make send
+# If you don't want to edit the makefile itself.
+################################################################################
+.PHONY: send
+send:
+	@tar czf - $(shell cd $(OUT_PATH) && echo $(OUT_PATH)/*.deb && echo $(ROOT)/linux-image-*.deb) | ssh linaro@$(IP) "cd /tmp; tar xvzf -"
+	@echo "Files has been sent to $$IP/tmp/ and $$IP/tmp/out"
+	@echo "On the device, run:"
+	@echo " dpkg --force-all -i /tmp/out/optee_$(OPTEE_PKG_VERSION).deb"
+	@echo " dpkg --force-all -i /tmp/linux-image-*.deb"
```
vchong: Any thoughts for choosing this approach over mounting the sys img and just copying the required files over locally? Then it's just flashing instead of having to deal with IPs, i.e. flashing, booting the board, configuring wifi and rebooting, before we can finally send?

hikey_debian.mk
jbech-linaro: I think I'll do some changes here, the alip image on the RPB 16.03 ([link](https://builds.96boards.org/releases/reference-platform/debian/hikey/16.03/)) doesn't have any click through license and that makes life a lot easier. So, having that said, I'll change to take both the boot image and the system image from that location instead.

Define URLs on top: sure, sounds like a good idea.

hikey_debian.mk
jbech-linaro: My comment is badly worded and I will rewrite it. My intention was to say, that the gunzip'd file should have the name provided by the `@echo "  $(SYSTEM_IMG)"` line. I'll fix this.

hikey_debian.mk
jbech-linaro: I've been on that route also. It for sure would work and there are a lot of ways that we can use to put the files on the device, to name few, mount the image and put files there, ssh deb packages, ssh files individually, rsync a set of files, ftp files etc. Mounting and putting the files is for sure something that we could do, at least for the initial flash and I think we should do that as soon as possible. Having that said I did that, but I had some issues when flashing the system image, my modified image sometimes failed to flash and I have no time to investigate why right now. Doing the "deb" way is rather convenient (besides the build time in Linux). That will sort out all the grub stuff for you. 

What I'd like to do in short term is:
1. Fix the image like you're suggesting.
2. Configure it so rsync will update all files that needs to be refreshed, it could be both optee_client stuff as well as the kernel and the modules.

But, it'd like to take that in separate steps, there is already too much ongoing in this tiny patch set.

hikey_debian.mk
jbech-linaro: I understand that you'd like to have that here also, but I would like to keep this as a flash only target and "you need to know what to do before flashing". Even though it's nice to ensure that the binaries are in a good state, it's the same time rather annoying to see it jumping in and out of all sorts of gits when you just want to flash.

As a related side note. The "make run" target for QEMU etc. In the beginning that was rather quick and just checked a few well chosen files and just did the minimal amount of work to update the rootfs. Now, it seems like it's doing the same as "make all", which is kind of annoying I think, since it does a lot of stuff.

Summary: I won't add a dependency here :)

hikey_debian.mk
jbech-linaro: Thanks! I'll fix.

jbech-linaro:
Tags applied, addressed review comments, ready to be merged.

PR #65, opened 2016-05-20
gcc5: Update to use gcc5

jbech-linaro:
- Adds legacy toolchain for use in older edk2 projects, this is the only
  project using the pre 5.x toolchain for now
- Removes the bare metal toolchain
- qemu: Change the path for libraries for GCC5
- fvp: compiler changes for ARM-TF and EDK2
       Update libteec path in initramfs
- mtk: Toolchain and libteec path updates
- juno: Toolchain and libteec path updates

Tested-by: Joakim Bech <joakim.bech@linaro.org> (QEMU, FVP, HiKey)

### DIFF
--- a/mediatek.mk
+++ b/mediatek.mk
@@ -33,7 +33,7 @@ all-clean: arm-tf-clean linux-clean busybox-clean optee-os-clean \
 ################################################################################
 ARM_TF_EXPORTS ?= \
 	CFLAGS="-O0 -gdwarf-2" \
-	CROSS_COMPILE="$(CCACHE)$(AARCH64_NONE_CROSS_COMPILE)"
+	CROSS_COMPILE="$(CCACHE)$(AARCH64_CROSS_COMPILE)"
 
 ARM_TF_FLAGS ?= \
 	DEBUG=1 \
--- a/fvp.mk
+++ b/fvp.mk
@@ -39,7 +39,7 @@ all-clean: arm-tf-clean busybox-clean edk2-clean optee-os-clean \
 ################################################################################
 ARM_TF_EXPORTS ?= \
 	CFLAGS="-O0 -gdwarf-2" \
-	CROSS_COMPILE="$(CCACHE)$(AARCH64_NONE_CROSS_COMPILE)"
+	CROSS_COMPILE="$(CCACHE)$(AARCH64_CROSS_COMPILE)"
 
 ARM_TF_FLAGS ?= \
 	BL32=$(OPTEE_OS_BIN) \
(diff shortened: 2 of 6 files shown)


jbech-linaro:
FYI: I have [another branch](https://github.com/jbech-linaro/build/commits/gcc5) where the patches aren't squashed. But after thinking a bit, I thought that it's still better to have this as a single patch.

mediatek.mk
```diff
@@ -33,7 +33,7 @@ all-clean: arm-tf-clean linux-clean busybox-clean optee-os-clean \
 ################################################################################
 ARM_TF_EXPORTS ?= \
 	CFLAGS="-O0 -gdwarf-2" \
-	CROSS_COMPILE="$(CCACHE)$(AARCH64_NONE_CROSS_COMPILE)"
+	CROSS_COMPILE="$(CCACHE)$(AARCH64_CROSS_COMPILE)"
```
jbech-linaro: I should probably test on the MTK device also, since I'd expect that we need the /lib/foo.xyz changes there too.

pascal-brand-st-dev:
`Tested-by: Pascal Brand <pascal.brand@linaro.org> (QEMU)`
`Tested-by: Pascal Brand <pascal.brand@linaro.org> (FVP)`

mediatek.mk
jbech-linaro: So, **MediaTek** board:
Build wise all OK, but unfortunately the `boot.img` grows from ~6MB to 41MB when switching from GCC4.9 to GCC5.3. The reason seems to be that we get quite a few more files copied into the root fs [here](https://github.com/linaro-swg/gen_rootfs/blob/master/generate-cpio-rootfs.sh#L37-L41). Due to this the flashing of the `boot.img` fails.
```
writing 'boot'...
FAILED (remote: size too large)
```

**HiKey:**
Probably too many things in my head, since I was right about that I had tried it, but I had totally forgotten about the branch ([gcc5](https://github.com/jbech-linaro/patches_hikey/tree/gcc5)) I created for the `patches_hikey` git where I made the changes needed. I've now re-verified using that particular branch and things are working as it should. I've sent a [pull request](https://github.com/linaro-swg/patches_hikey/pull/3) for this also.

My recommendation as of now is to merge the GCC5.x pull requests. The MediaTek platform won't work after this. But based on things happening and general + MTK8173 support not the best in terms of activity (neither in Linux kernel, ARM-TF nor in OP-TEE) I think we need to drop active support for mtk8173 from now on if no-one else at MediaTek are willing to work with it.

pascal-brand-st-dev: +1

jbech-linaro:
Patch rebased, tags applied and ready for merge.

PR #67, opened 2016-06-16
hikey: Use mtools to avoid sudo during build

d3zd3z:
Instead of using sudo/mount to build the FAT image, use commands from
the mtools package to copy files into the FAT image.  This eliminates
the requirement to use sudo during the build.  It also removes a small
sleep necessary to make umount work.

Signed-off-by: David Brown <david.brown@linaro.org>

### DIFF
--- a/hikey.mk
+++ b/hikey.mk
@@ -334,25 +334,15 @@ GRUBCFG = $(PATCHES_PATH)/grub/grub_uart0.cfg
 endif
 
 boot-img: linux update_rootfs edk2 grub
-	sudo -p "[sudo] Password:" true
-	if [ -d .tmpbootimg ] ; then sudo rm -rf .tmpbootimg ; fi
-	mkdir -p .tmpbootimg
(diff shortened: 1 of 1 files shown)


jenswi-linaro:

jforissier:
Nice!

jbech-linaro:

jbech-linaro:
I tried this and was first surprised that it still asked for sudo. But after looking at what's happening I found another Makefile that uses sudo. I'm not sure whether the same tricks can be applied to that [Makefile](https://github.com/96boards/l-loader/blob/master/Makefile#L18) in l-loader?.

Anyhow, this particular patch works, so I'll go a head and merge it.

d3zd3z:
The l-loader sudo is even easier, just remove it. As long as you don't have root-owned files sitting around, it will build just fine without the sudo.

jforissier:
@d3zd3z good suggestion, works fine here too without sudo. I've created https://github.com/96boards/l-loader/pull/12.

jbech-linaro:
Thanks @jforissier !

PR #69, opened 2016-07-08
qemu_v8: add First build script for QEMU ARMv8-A

matt2048:
QEMU and OP-TEE on ARMv8 is working

Signed-off-by: Matt Ma <matt.ma@linaro.org>

### DIFF
--- a/qemu_v8.mk
+++ b/qemu_v8.mk
@@ -1,0 +1,275 @@
+################################################################################
+# Following variables defines how the NS_USER (Non Secure User - Client
+# Application), NS_KERNEL (Non Secure Kernel), S_KERNEL (Secure Kernel) and
+# S_USER (Secure User - TA) are compiled
+################################################################################
+override COMPILE_NS_USER   := 64
+override COMPILE_NS_KERNEL := 64
+override COMPILE_S_USER    := 64
+override COMPILE_S_KERNEL  := 64
+
+-include common.mk
+
+################################################################################
+# Paths to git projects and various binaries
+################################################################################
+ARM_TF_PATH			?= $(ROOT)/arm-trusted-firmware
+
+EDK2_PATH			?= $(ROOT)/edk2
+EDK2_BIN			?= $(EDK2_PATH)/QEMU_EFI.fd
+
+QEMU_PATH			?= $(ROOT)/qemu
+
+SOC_TERM_PATH			?= $(ROOT)/soc_term
+STRACE_PATH			?= $(ROOT)/strace
+
+DEBUG = 1
+
+################################################################################
+# Targets
+################################################################################
+all: arm-tf qemu soc-term linux strace update_rootfs
+all-clean: arm-tf-clean busybox-clean linux-clean \
+	optee-os-clean optee-client-clean qemu-clean \
+	soc-term-clean check-clean strace-clean
+
+-include toolchain.mk
+
+################################################################################
+# ARM Trusted Firmware
+################################################################################
+ARM_TF_EXPORTS ?= \
+	CFLAGS="-O0 -gdwarf-2" \
+	CROSS_COMPILE="$(CCACHE)$(AARCH64_CROSS_COMPILE)"
+
+ARM_TF_FLAGS ?= \
+	BL32=$(OPTEE_OS_BIN) \
+	BL33=$(EDK2_BIN) \
+	ARM_TSP_RAM_LOCATION=tdram \
+	PLAT=qemu \
+	DEBUG=0 \
+	LOG_LEVEL=50 \
+	ERROR_DEPRECATED=1 \
+	BL32_RAM_LOCATION=tdram \
+	SPD=opteed
+
+arm-tf: optee-os edk2
+	$(ARM_TF_EXPORTS) $(MAKE) -C $(ARM_TF_PATH) $(ARM_TF_FLAGS) all fip
+
+arm-tf-clean:
+	$(ARM_TF_EXPORTS) $(MAKE) -C $(ARM_TF_PATH) $(ARM_TF_FLAGS) clean
+
+# FIXME: This is just too rough, we should build this just as we're doing for
+#        FVP.
+edk2: optee-os
+ifeq ("$(wildcard $(EDK2_BIN))","")
+	mkdir -p $(EDK2_PATH)
+	wget -O $(EDK2_BIN) \
+		http://snapshots.linaro.org/components/kernel/leg-virt-tianocore-edk2-upstream/989/QEMU-KERNEL-AARCH64/RELEASE_GCC49/QEMU_EFI.fd
+endif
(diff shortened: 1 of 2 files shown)


qemu_v8.mk
```diff
@@ -0,0 +1,279 @@
...
+all-clean: arm-tf-clean busybox-clean linux-clean \
+	optee-os-clean optee-client-clean qemu-clean \
+	soc-term-clean check-clean strace-clean
+
+-include toolchain.mk
+
+################################################################################
+# ARM Trusted Firmware
+################################################################################
+# FIXME: Double check why Peter isn't using the bare metal compiler
```
jbech-linaro: We can remove this comment.

qemu_v8.mk
```diff
@@ -0,0 +1,279 @@
...
+	BL32_RAM_LOCATION=tdram \
+	SPD=opteed
+
+arm-tf: optee-os edk2
+	$(ARM_TF_EXPORTS) $(MAKE) -C $(ARM_TF_PATH) $(ARM_TF_FLAGS) all fip
+
+arm-tf-clean:
+	$(ARM_TF_EXPORTS) $(MAKE) -C $(ARM_TF_PATH) $(ARM_TF_FLAGS) clean
+
+# FIXME: This is just too rough, we should build this just as we're doing for
```
jbech-linaro: I think for now, this is good enough, but as soon as we find some time for it, we should as comment says build edk2 from source.

jbech-linaro:
Just a single line that could be removed, otherwise:

qemu_v8.mk
matt2048: OK, thanks, I have removed this comment

qemu_v8.mk
```diff
@@ -0,0 +1,278 @@
...
+#        FVP.
+edk2:
+ifeq ("$(wildcard $(EDK2_BIN))","")
+	mkdir -p $(EDK2_PATH)
+	wget -O $(EDK2_BIN) \
+		http://snapshots.linaro.org/components/kernel/leg-virt-tianocore-edk2-upstream/989/QEMU-KERNEL-AARCH64/RELEASE_GCC49/QEMU_EFI.fd
+endif
+	mkdir -p $(ARM_TF_PATH)/build/qemu/release
+	ln -sf $(OPTEE_OS_BIN) $(ARM_TF_PATH)/build/qemu/release/bl32.bin
```
vchong: minor: If user calls `make edk2` directly and `optee_os` hasn't been built, then this is a broken link, so maybe add a dependency for it: `edk2: optee-os`? Should be temporary only until edk2 is built from source since edk2 shouldn't really depend on optee-os anyway.

qemu_v8.mk
```diff
@@ -0,0 +1,278 @@
...
+################################################################################
+# xtest / optee_test
+################################################################################
+XTEST_COMMON_FLAGS += CFG_ARM32=y
+xtest: xtest-common
+
+XTEST_CLEAN_COMMON_FLAGS += CFG_ARM32=y
+xtest-clean: xtest-clean-common
+
+XTEST_PATCH_COMMON_FLAGS += CFG_ARM32=y
```
vchong: Why `CFG_ARM32`?

qemu_v8.mk
```diff
@@ -0,0 +1,278 @@
...
+
+################################################################################
+# strace
+################################################################################
+strace:
+ifneq ("$(wildcard $(STRACE_PATH))","")
+		cd $(STRACE_PATH) && \
+		./bootstrap && \
+		./configure --host=aarch64-linux-gnu CC=$(CROSS_COMPILE_NS_USER)gcc && \
+		CC=$(CROSS_COMPILE_NS_USER)gcc make
```
vchong: minor: `$(MAKE)` instead of `make`? Also, `LD=$(CROSS_COMPILE_NS_USER)ld` is not required here and above line?

qemu_v8.mk
```diff
@@ -0,0 +1,278 @@
...
+endif
+
+strace-clean:
+ifneq ("$(wildcard $(STRACE_PATH))","")
+		CC=$(CROSS_COMPILE_NS_USER)gcc LD=$(CROSS_COMPILE_NS_USER)ld \
+			make -C $(STRACE_PATH) clean && \
```
vchong: minor: `$(MAKE)` instead of `make`?

qemu_v8.mk
```diff
@@ -0,0 +1,278 @@
...
+################################################################################
+.PHONY: filelist-tee
+filelist-tee: xtest
+	@echo "# xtest / optee_test" > $(GEN_ROOTFS_FILELIST)
+	@find $(OPTEE_TEST_OUT_PATH) -type f -name "xtest" | sed 's/\(.*\)/file \/bin\/xtest \1 755 0 0/g' >> $(GEN_ROOTFS_FILELIST)
+	@echo "# TAs" >> $(GEN_ROOTFS_FILELIST)
+	@echo "dir /lib/optee_armtz 755 0 0" >> $(GEN_ROOTFS_FILELIST)
+	@find $(OPTEE_TEST_OUT_PATH) -name "*.ta" | \
+		sed 's/\(.*\)\/\(.*\)/file \/lib\/optee_armtz\/\2 \1\/\2 444 0 0/g' >> $(GEN_ROOTFS_FILELIST)
+	@echo "# Secure storage dig" >> $(GEN_ROOTFS_FILELIST)
```
vchong: minor: should really say `dir` instead of `dig`

qemu_v8.mk
matt2048: Thanks a lot, you are right, I think we can move this link to optee-os target section, I will confirm it with Joakim what is the original idea.

jbech-linaro: Victor, I think you're right, the edk2 target should be update to `edk2: optee-os`

qemu_v8.mk
jbech-linaro: Good question, I probably forgot to change to 64 when I created the file. 64 would be a better choice here, I agree with that.

qemu_v8.mk
matt2048: Thank you, yes, I have tried it, everything is ok.

qemu_v8.mk
matt2048: Yes, sorry , we can use $(MAKE) instead of make

qemu_v8.mk
matt2048: Yes, sorry, I think it is just copied from qemu.mk

qemu_v8.mk
matt2048: OK, I will update it as soon as possible.

qemu_v8.mk
matt2048: OK, I am testing it now, if it is ok, I will remove this flag

PR #72, opened 2016-07-28
hikey: fix bug: @echo: command not found

vchong:
Tested-by: Victor Chong <victor.chong@linaro.org> (hikey)

### DIFF
--- a/hikey.mk
+++ b/hikey.mk
@@ -273,14 +273,12 @@ awk '{while(match($$0,"[$$]{[^}]*}")) {var=substr($$0,RSTART+2,RLENGTH -3);gsub(
 endef
 
 filelist-all: busybox
-	cat $(GEN_ROOTFS_PATH)/filelist-final.txt | sed '/fbtest/d' > $(GEN_ROOTFS_PATH)/filelist-all.txt
-	export KERNEL_VERSION=$(call KERNEL_VERSION); \
-	export TOP=$(ROOT); export MULTIARCH=$(MULTIARCH); \
-	$(expand-env-var) <$(PATCHES_PATH)/rootfs/initramfs-add-files.txt >> $(GEN_ROOTFS_PATH)/filelist-all.txt; \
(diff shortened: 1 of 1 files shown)


hikey.mk
```diff
@@ -278,10 +278,10 @@ filelist-all: busybox
 	export TOP=$(ROOT); export MULTIARCH=$(MULTIARCH); \
 	$(expand-env-var) <$(PATCHES_PATH)/rootfs/initramfs-add-files.txt >> $(GEN_ROOTFS_PATH)/filelist-all.txt; \
```
jforissier: It seems that only this line (expand-env-var) needs `TOP` and `MULTIARCH` to be set. So you may as well remove all the line continuation characters from the subsequent lines (well not inside the `if` of course) and keep the `@`. I think it would be a better solution because it's not the first time this kind of mistake happens IIRC.

vchong: Yes thanks! Will change accordingly.

d3zd3z: As far as running commands with specific variables set, might it not be clearer to use the `env` command:
```
        env TOP=$(ROOT) MULTIARCH=$(MULTIARCH) command...
```

jforissier: @d3zd3z +1 it looks a bit better indeed

vchong: Yes, tried and it works. Will use that instead. Thanks!

vchong:
Updated. Also got rid of `export KERNEL_VERSION...` since it's not needed anymore as well.

jforissier:
It seems `MULTIARCH` is not used in `initramfs-add-files.txt` either?

vchong:
Urgh. Right, sorry. I looked at different version of `initramfs-add-files.txt`. Will fix.

vchong:
`MULTIARCH` removed.

jforissier:

vchong:
Squashed and tag applied. Phew! Thanks!

PR #73, opened 2016-07-29
qemu: Ability to switch Python version for QEMU

jbech-linaro:
On some distros /usr/bin/python points to Python 3.x and changing that
to point to Python 2.x might not be an easy thing to do. By adding the
variable PYTHON_PATH one can easily change the python version to be used
with QEMU (QEMU only supports Python 2.x).

For a distro where Python 3.x is the default one could/should call make
like this:

 $ PYTHON_PATH=/usr/bin/python2 make

Signed-off-by: Joakim Bech joakim.bech@linaro.org

### DIFF
--- a/qemu.mk
+++ b/qemu.mk
@@ -18,6 +18,8 @@ QEMU_PATH			?= $(ROOT)/qemu
 
 SOC_TERM_PATH			?= $(ROOT)/soc_term
 
+PYTHON_PATH			?= $(shell which python)
+
 DEBUG = 1
 
 ################################################################################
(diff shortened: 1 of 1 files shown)


d3zd3z:
I would expect python2 to be available on pretty much any modern system. Should we just change the references to that instead of needing to pass anything in? The python2 symlink should be available in any system modern enough to hope to be able to build OP-TEE.

jbech-linaro:
I'm closing this PR, since my patch here isn't the best way to deal with this. I think either we should do as David proposed or you simply leave it up to the user to source, so that the `$PATH` contains python2 before python3.

PR #75, opened 2016-09-14
RPi3: Merge build files to the official build.git

jbech-linaro:
I've done quite a few changes since I got this from Philip, but still, I decided to squash it to make it more coherent. So if things are wrong here, it's not only Phil that should be blamed :)

I haven't tried the JTAG / debugger related stuff, but since I know that Phil uses that and will use it at a couple of conferences in short term I though it still might be a good idea to include that.


rpi3/firmware/optee
```diff
@@ -0,0 +1,35 @@
+#!/bin/sh
```
jbech-linaro: I'm not sure if this is useful at all? I do believe Phil used it in one or another way. Me myself hasn't used it.

rpi3.mk
```diff
@@ -0,0 +1,296 @@
...
+update_rootfs: arm-tf busybox u-boot optee-client xtest filelist-tee linux
+	cat $(GEN_ROOTFS_PATH)/filelist-final.txt $(GEN_ROOTFS_PATH)/filelist-tee.txt > $(GEN_ROOTFS_PATH)/filelist.tmp
+	cd $(GEN_ROOTFS_PATH) && \
+	        $(LINUX_PATH)/usr/gen_init_cpio $(GEN_ROOTFS_PATH)/filelist.tmp | gzip > $(GEN_ROOTFS_PATH)/filesystem.cpio.gz
+
+# Creating images etc, could wipe out a drive on the system, therefore we don't
+# want to automate that in script or make target. Instead we just simply provide
+# the steps here.
+.PHONY: img-help
+img-help:
```
jbech-linaro: I remember that David had a comment about this, why not to put in the the rpi3.md file instead. Well, that could of course work. The neat thing here, is that it's readily available and it will produce instructions with correct paths etc according to where you have the project on your disk. So, I'm still in favor of keeping it as an "instruction" make target and just mentioned the target in the rpi3.md file instead.

jbech-linaro:
The `kconfigs/u-boot_rpi3.conf` seems to be empty. I'll have a check to see whether I missed something or if it simply just should be removed (at least I didn't notice any problem when testing it from scratch).

jbech-linaro:
The `kconfigs/u-boot_rpi3.conf` has never been used. I think it's a copy/paste artifact from the Juno's U-Boot config (which also is empty and also should be removed). I'll push a new patch where I've removed the file and the line pointing to it (Juno will be dealt with in a separate commit).

rpi3.mk
jforissier: LGTM

rpi3/firmware/optee
jforissier: Do you start `tee-supplicant` manually?

jbech-linaro: Yes that's what I've done. I have no strong preference to autostart it or just do it manually. For development I prefer doing it manually, since you have the ability to run gdb, valgrind etc "from start" on the binary.

jforissier: OK, then I'd rather not have this file committed, to avoid bitrot.

jbech-linaro: Also, tee-supplicant is one thing. But I don't think the checks for the driver is necessary. Since that's something that will be loaded automatically when starting tee-supplicant.

jforissier: Indeed, with the current setup the driver is not even a module

jbech-linaro: I'll remove it. Sounds like we're on the same page here. Phil can add it later on if needed.

jbech-linaro:
`head.bin`? Good question I'll need to ask Phil about it. `u-boot.env`? I think that's based on some generic U-boot environment file. The way to update u-boot env is boot up, modify your parameters, save changes and put that back into the git (that's how I've updated it). There are software tools that can modify it also. I'm no U-boot expert, but I've got the impression that booting up and make changes is the most common way to update it. Besides cleaning up the u-boot.env file from unused variables, I think that file will be rather static.

jbech-linaro:
I agree regarding u-boot.env. Do you think it would be OK to do that in another patch solely about fixing that?

d3zd3z:
I think I agree with the u-boot.env: Just commit this version now, and come up with a later patch to generate it. Does generating it create any patches.

I believe this file is just generated from a text file (or at least has a textual format).

Otherwise, looks good to me.

jforissier:
Fine by me.

jbech-linaro:
Wait a second before pushing the merge button, when cleaning all and rebuild I saw some issue. Let me double check first.

jbech-linaro:
So, apparently u-boot didn't like to merge a single file (which kind of makes sense, so I've changed it like this:

```diff
diff --git a/rpi3.mk b/rpi3.mk
index 7bcd516..af10859 100644
--- a/rpi3.mk
+++ b/rpi3.mk
@@ -96,13 +96,9 @@ arm-tf-clean:
 
 U-BOOT_EXPORTS ?= CROSS_COMPILE=$(LEGACY_AARCH64_CROSS_COMPILE) ARCH=arm64
 
-U-BOOT_DEFCONFIG_FILES := \
-       $(U-BOOT_PATH)/configs/rpi_3_defconfig
-
 .PHONY: u-boot
 u-boot:
-       cd $(U-BOOT_PATH) && \
-               scripts/kconfig/merge_config.sh $(U-BOOT_DEFCONFIG_FILES)
+       $(U-BOOT_EXPORTS) $(MAKE) -C $(U-BOOT_PATH) rpi_3_defconfig
        $(U-BOOT_EXPORTS) $(MAKE) -C $(U-BOOT_PATH) all
        cd $(U-BOOT_PATH) && cat $(RPI3_HEAD_BIN) $(U-BOOT_BIN) > $(U-BOOT_JTAG_BIN)
```

Everything rebuilt from scratch, all seems to be OK. I will squash this fix and force push again. I'll post another message when I'm done.

jbech-linaro:

```
22639 subtests of which 0 failed
67 test cases of which 0 failed
0 test case was skipped
TEE test application done!
root@RPi3:/ 
```

jforissier:
Merged, thanks!

PR #78, opened 2016-10-17
qemu-check.exp: monitor TEE core log file and detect assertion failures

jforissier:
When "make check" is run and an assertion fails in TEE core, secure
world hangs, and therefore xtest hangs too. "make check" will not
report the failure and exit until a timeout occurs (480s).
In order to detect the error condition sooner, open the TEE core log
file and add a trigger on "assertion ... failed at ...".


jbech-linaro:

PR #79, opened 2016-10-26
Enable configuring QEMU with Virtfs support

igoropaniuk:
Add support for configuring QEMU with VirtFS to conveniently share 
files between the guest QEMU VM and host operating systems.

To specify dir, that has to be mounted inside QEMU VM, edit
QEMU_VIRTFS_HOST_DIR var in qemu.mk

To mount shared folder within QEMU ARM VM, run:
$ mount -t 9p -o trans=virtio host /tmp/shared/

If Virtfs support is enabled, libattr1-dev and libcap-dev
have to be installed before configuring QEMU.

For additional details see:
http://wiki.qemu.org/Documentation/9psetup

Signed-off-by: Igor Opaniuk <igor.opaniuk@linaro.org>


jenswi-linaro:
It would be nice to have the mount command documented somewhere more than in the commit message.

qemu.mk
```diff
@@ -75,6 +88,11 @@ LINUX_DEFCONFIG_COMMON_FILES := \
 		$(LINUX_PATH)/arch/arm/configs/vexpress_defconfig \
 		$(CURDIR)/kconfigs/qemu.conf
 
+ifeq ($(QEMU_VIRTFS_ENABLE),y)
+LINUX_DEFCONFIG_FILES += \
+		$(CURDIR)/kconfigs/9pfs.conf
```
jforissier: On the same line

igoropaniuk:
@jenswi-linaro I'm going to add it to [OP-TEE/README.md](https://github.com/OP-TEE/build/blob/master/README.md) inside QEMU section

kconfigs/9pfs.conf
```diff
@@ -0,0 +1,4 @@
+CONFIG_NET_9P=y
```
jbech-linaro: I don't see any issue with always having that enabled., so I'd prefer putting those configs in the already existing conf
https://github.com/OP-TEE/build/blob/master/kconfigs/qemu.conf

qemu.mk
```diff
@@ -32,6 +32,19 @@ all-clean: bios-qemu-clean busybox-clean linux-clean optee-os-clean \
 ################################################################################
 # QEMU
 ################################################################################
+QEMU_CONFIGURE_PARAMS = --target-list=arm-softmmu --cc="$(CCACHE)gcc" \
+						--extra-cflags="-Wno-error"
+
+QEMU_VIRTFS_ENABLE		?= n
```
jbech-linaro: Add alias to `/etc/profile` in [gen_rootfs](https://github.com/linaro-swg/gen_rootfs/blob/master/etc/profile) and update the  [filelist-common](https://github.com/OP-TEE/build/blob/master/common.mk#L271) target with yet another directory (might not work with `tmp` btw, since sometimes `/tmp` is mounted using tmpfs, but another folder would do `/mnt/shared`).

```
alias mount_shared='mount -t 9p -o trans=virtio host /tmp/shared/'
```
Maybe this would be overkill, maybe just easier to add it to the README.md as Igor suggested in another comment. See this as suggestions and "not a must do thing" ;-)

igoropaniuk: What about mounting just to /mnt?

jbech-linaro: That would work, it's just that `mnt` and/or `media` on most systems today is the root folder for mounted partitions etc. I.e, there is almost always a sub-folder for the actual mount. But for this purpose and how we're using it we won't probably be interested in mounting other stuff.

igoropaniuk: what about just passing mount point to alias as an additional param:
```
alias mount_shared='mount -t 9p -o trans=virtio host'
```
Usage:
`$ mount_shared /mnt/shared`

igoropaniuk:
Also added support for QEMUv8

jforissier:
Looks good, thanks!

jbech-linaro:

igoropaniuk:
Applied tags and rebased

PR #80, opened 2016-10-26
Fix toolchain.mk ROOT path

igoropaniuk:
Change path to relative to current dir, the same as in common.mk

### DIFF
--- a/toolchain.mk
+++ b/toolchain.mk
@@ -1,7 +1,7 @@
 ################################################################################
 # Toolchains
 ################################################################################
-ROOT				?= ${HOME}/devel/optee
+ROOT					?= $(CURDIR)/..
 TOOLCHAIN_ROOT 			?= $(ROOT)/toolchains
 
 AARCH32_PATH 			?= $(TOOLCHAIN_ROOT)/aarch32


toolchain.mk
```diff
@@ -1,7 +1,7 @@
 ################################################################################
 # Toolchains
 ################################################################################
-ROOT				?= ${HOME}/devel/optee
+ROOT					?= $(shell pwd)/..
```
jforissier: `$(CURDIR)/..`

igoropaniuk: isn't it the same?

jforissier: Same value, but simpler to write. Also a bit more efficient (not that we care in this particular case).

Note that using `$(CURDIR)` or `$(shell pwd)` here works because `toolchain.mk` is in the same directory as the main Makefile. So the current directory of the make process (which is what `$(CURDIR)` and `$(shell pwd)` return) and the directory where `toolchain.mk` resides are actually the same. If `toolchains.mk` were located somewhere else you would need to take that into account or use a more complex expression, something like: `$(patsubst %/,%,$(dir $(abspath $(lastword $(MAKEFILE_LIST)))))`.

jforissier:

PR #82, opened 2016-11-12
qemu: virtfs: fix wrong variable name

jforissier:
"make QEMU_VIRTFS_ENABLE=y" does not enable VirtFS because of a bad
variable name in common.mk.

Test case:

(host) echo Hello >/tmp/hello
(host) make -j8 run QEMU_VIRTFS_ENABLE=y QEMU_VIRTFS_HOST_DIR=/tmp
(QEMU) mkdir /tmp/host
(QEMU) mount_shared /tmp/host
(QEMU) cat /tmp/host/hello
Hello

Fixes: 97d0529c1687 ("Enable configuring QEMU with Virtfs support")


igoropaniuk:
Seems to me that I've never fully cleaned QEMU build before the last changes in common.mk, because I can't figure out how it still works on my PC (it's compiling with virtfs support) :)


Reviewed-by: Igor Opaniuk <igor.opaniuk@linaro.org>

PR #84, opened 2016-12-13
dra7xx: Initial support for TI DRA7xx

glneo:
Signed-off-by: Andrew F. Davis <afd@ti.com>

### DIFF
--- a/dra7xx.mk
+++ b/dra7xx.mk
@@ -1,0 +1,167 @@
+###############################################################################
+# Following variables defines how the NS_USER (Non Secure User - Client
+# Application), NS_KERNEL (Non Secure Kernel), S_KERNEL (Secure Kernel) and
+# S_USER (Secure User - TA) are compiled
+###############################################################################
+override COMPILE_NS_USER   := 32
+override COMPILE_NS_KERNEL := 32
+override COMPILE_S_USER    := 32
+override COMPILE_S_KERNEL  := 32
+
+###############################################################################
+# Includes
+###############################################################################
+-include common.mk
+
+###############################################################################
+# Paths to git projects and various binaries
+###############################################################################
+STAGING_AREA    ?= $(ROOT)/out
+U-BOOT_PATH     ?= $(ROOT)/u-boot
+UBOOT_SPL       ?= $(U-BOOT_PATH)/u-boot-spl_HS_MLO
+UBOOT_IMG       ?= $(U-BOOT_PATH)/u-boot_HS.img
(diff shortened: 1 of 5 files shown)


dra7xx.mk
```diff
@@ -0,0 +1,145 @@
...
+# Includes
+###############################################################################
+-include common.mk
+
+###############################################################################
+# Paths to git projects and various binaries
+###############################################################################
+U-BOOT_PATH     ?= $(ROOT)/u-boot
+LINUX_IMAGE     ?= $(LINUX_PATH)/arch/arm/boot/zImage
+STAGEING_AREA   ?= $(BUILD_PATH)/dra7xx
```
jbech-linaro: I think it would be better to use `$(ROOT)/out/dra7xx` as staging area. Speaking about that word, should it be "staging" or "stageing"? I think I'm used to see "staging".

dra7xx.mk
```diff
@@ -0,0 +1,145 @@
...
+###############################################################################
+# Build FIT
+###############################################################################
+.PHONY: build-fit build-fit-clean
+
+build-fit: linux update_rootfs optee-os
+	cp $(LINUX_IMAGE) $(STAGEING_AREA)/
+	cp $(LINUX_DTBS) $(STAGEING_AREA)/
+	cp $(OPTEE_OS_BIN) $(STAGEING_AREA)/
+	MKIMAGE=$(U-BOOT_PATH)/tools/mkimage $(MAKE) -C $(STAGEING_AREA)
```
jbech-linaro: This will create `fitImage.itb`. It would be quite nice to have that populated automatically to the /boot/ folder in the rootfs (i.e, `gen_rootfs/filesystem.cpio.gz`). Not that long ago each Makefile in build.git had separate sections where you could add specific files. Now that has been placed in the common.mk file instead (target `filelist-tee-common`). I have no straight answer how that could be added after we've changed to the common file. I can see that hikey.mk still has some code, but to me it looks like most of that code shouldn't be there any longer.

dra7xx.mk
glneo: Works for me, and I think that is a spelling mistake, will fix.

dra7xx.mk
glneo: I agree, I'll populate the /boot folder automatically.

jbech-linaro:
@jforissier I don't want to be the one delaying this anyone. The changes as such are good, but because I haven't got the environment for that device working I cannot test it yet. Andrew on the other hand has tested this (I suppose).

jforissier:
Rebased and merged, thanks.

PR #86, opened 2016-12-15
Compile head.bin from head.S

jforissier:


jenswi-linaro:
Can't `rpi3/firmware/head.bin` be removed now?

jforissier:
It *is* removed by this patch ;)

jenswi-linaro:
Aha. :-)

rpi3.mk
```diff
@@ -105,6 +105,16 @@ u-boot:
 u-boot-clean:
 	$(U-BOOT_EXPORTS) $(MAKE) -C $(U-BOOT_PATH) clean
 
+head-bin: $(RPI3_HEAD_BIN)
```
jenswi-linaro: Something should depend on this the `head-bin` target

jbech-linaro: Shouldn't it be the u-boot target that has a dependency?
I.e.
```Makefile
.PHONY: u-boot                                                                                                                     
u-boot:  $(RPI3_HEAD_BIN)
...
```

jenswi-linaro:

igoropaniuk:
I was wondering, what was a profit of having separated target for head.bin, and why not to compile it inside rpi3-firmware target, but then noticed that we have two separate firmware paths (a bit confusing) `RPI3_FIRMWARE_PATH`/`RPI3_STOCK_FW_PATH`, and, actually, this `head.S` doesn't relate to this rpi3-firmware target.

PR #89, opened 2016-12-21
RPi3 parallel build fix

igoropaniuk:
Fix parallel build issue (make -jn), now it works fine
Minor: add additional command for the proper cleanup of rpi3_firmware
target.

### DIFF
--- a/rpi3.mk
+++ b/rpi3.mk
@@ -40,14 +40,15 @@ U-BOOT_PATH		?= $(ROOT)/u-boot
 U-BOOT_BIN		?= $(U-BOOT_PATH)/u-boot.bin
 U-BOOT_JTAG_BIN		?= $(U-BOOT_PATH)/u-boot-jtag.bin
 
-RPI3_FIRMWARE_PATH	?= $(BUILD_PATH)/rpi3/firmware
-RPI3_HEAD_BIN		?= $(ROOT)/out/head.bin
-RPI3_BOOT_CONFIG	?= $(RPI3_FIRMWARE_PATH)/config.txt
-RPI3_UBOOT_ENV		?= $(ROOT)/out/uboot.env
-RPI3_UBOOT_ENV_TXT	?= $(RPI3_FIRMWARE_PATH)/uboot.env.txt
-RPI3_STOCK_FW_PATH	?= $(ROOT)/rpi3_firmware
-
-OPTEE_OS_PAGER		?= $(OPTEE_OS_PATH)/out/arm/core/tee-pager.bin
+RPI3_FIRMWARE_PATH		?= $(BUILD_PATH)/rpi3/firmware
+RPI3_HEAD_BIN			?= $(ROOT)/out/head.bin
+RPI3_BOOT_CONFIG		?= $(RPI3_FIRMWARE_PATH)/config.txt
+RPI3_UBOOT_ENV			?= $(ROOT)/out/uboot.env
+RPI3_UBOOT_ENV_TXT		?= $(RPI3_FIRMWARE_PATH)/uboot.env.txt
+RPI3_STOCK_FW_PATH		?= $(ROOT)/rpi3_firmware
+RPI3_STOCK_FW_PATH_BOOT	?= $(RPI3_STOCK_FW_PATH)/boot
+RPI3_STOCK_ZIP_BOOT		?= firmware-$(RPI3_FIRMWARE_FILE)/boot
+OPTEE_OS_PAGER			?= $(OPTEE_OS_PATH)/out/arm/core/tee-pager.bin
 
 LINUX_IMAGE		?= $(LINUX_PATH)/arch/arm64/boot/Image
 LINUX_DTB		?= $(LINUX_PATH)/arch/arm64/boot/dts/broadcom/bcm2710-rpi-3-b.dtb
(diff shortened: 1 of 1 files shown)


rpi3.mk
```diff
@@ -192,6 +192,7 @@ endif
 .PHONY: rpi3-firmware-clean
 rpi3-firmware-clean:
 	rm -f $(ROOT)/out/$(RPI3_FIRMWARE_FILE).$(RPI3_FIRMWARE_FILE_EXT)
+	rm -rf $(RPI3_STOCK_FW_PATH)
```
jenswi-linaro: Don't use `rm -rf` please. What if `RPI3_STOCK_FW_PATH` holds an unexpected value?

igoropaniuk: I totally agree with you, but I'm not sure how to avoid it here and properly cleanup RPI3_STOCK_FW_PATH dir, which is created here (last mv command).
```
ifeq ("$(wildcard $(ROOT)/out/$(RPI3_FIRMWARE_FILE).$(RPI3_FIRMWARE_FILE_EXT))","")
	echo "Downloading Raspberry Pi 3 firmware ..."
	mkdir -p $(ROOT)/out
	wget $(RPI3_FIRMWARE_URL)/$(RPI3_FIRMWARE_FILE).$(RPI3_FIRMWARE_FILE_EXT) -O $(ROOT)/out/$(RPI3_FIRMWARE_FILE).$(RPI3_FIRMWARE_FILE_EXT)
	unzip -a $(ROOT)/out/$(RPI3_FIRMWARE_FILE).$(RPI3_FIRMWARE_FILE_EXT) -d $(ROOT)
	mv $(ROOT)/firmware-$(RPI3_FIRMWARE_FILE) $(RPI3_STOCK_FW_PATH)
```

jenswi-linaro: Sorry, I don't have any solution, but `rm -rf` is not an option.
Hoping that no one will notice or care about the leftovers is better than `rm -rf` IMO.

jforissier: @igoropaniuk maybe you can remove each file one by one and then the directory if it's empty, as in https://github.com/jforissier/optee_client/commit/6432f5194822d7da586d736e8733d7b78401bdec

igoropaniuk: @jforissier 
I think in this case it's a bad idea :)
```
$ find ../rpi3_firmware/ | wc -l
4507
```

jbech-linaro: Deleting folders from make is a headache. If it's just a single folder, then one can specify the files and just delete the files. But when it's like in the "firmware" here where we have a bunch of folders it's a headache.

If you insist using `rm -rf` putting something more than just a single variable and having the variable **first** in the path and you should be pretty safe. i.e., `rm -rf $(ROOT)/rpi3_firmware`. If you're really paranoid you could even rename the folder (sha1 of the current name) to be absolutely sure that there is no collision and say `rm -rf $(ROOT)/rpi3_firmware_b8eee5117faa675da731b22c7472f3d1a29b14cf` and then it's just as safe as the security of some secure protocols.

Having that said, be careful so it doesn't end up like [this](http://www.independent.co.uk/life-style/gadgets-and-tech/news/man-accidentally-deletes-his-entire-company-with-one-line-of-bad-code-a6984256.html).

igoropaniuk:
I hope now it's pretty safe :)

rpi3.mk
```diff
@@ -184,15 +185,41 @@ rpi3-firmware:
 ifeq ("$(wildcard $(ROOT)/out/$(RPI3_FIRMWARE_FILE).$(RPI3_FIRMWARE_FILE_EXT))","")
 	echo "Downloading Raspberry Pi 3 firmware ..."
 	mkdir -p $(ROOT)/out
+	mkdir -p $(RPI3_STOCK_FW_PATH)/boot
 	wget $(RPI3_FIRMWARE_URL)/$(RPI3_FIRMWARE_FILE).$(RPI3_FIRMWARE_FILE_EXT) -O $(ROOT)/out/$(RPI3_FIRMWARE_FILE).$(RPI3_FIRMWARE_FILE_EXT)
-	unzip -a $(ROOT)/out/$(RPI3_FIRMWARE_FILE).$(RPI3_FIRMWARE_FILE_EXT) -d $(ROOT)
-	mv $(ROOT)/firmware-$(RPI3_FIRMWARE_FILE) $(RPI3_STOCK_FW_PATH)
+	unzip -aj $(ROOT)/out/$(RPI3_FIRMWARE_FILE).$(RPI3_FIRMWARE_FILE_EXT) \
+	'$(RPI3_STOCK_ZIP_BOOT)/bootcode.bin' \
```
jenswi-linaro: This list occurs when removing also, perhaps it should go into a variable to save the repetition.

jbech-linaro: Or even expand it to also include folders instead of a single variable name and then something like this in the clean rule:
```make
 +		rm -f $(ROOT)/rpi3_firmware/boot/*'; \
 +		rmdir $(RPI3_STOCK_FW_PATH_BOOT); \
 +		rmdir $(RPI3_STOCK_FW_PATH); \
```

jenswi-linaro: I think it's quite nice to list all the file that are to removed, it shows that you're not blindly removing files.

igoropaniuk: @jenswi-linaro 
> This list occurs when removing also, perhaps it should go into a variable to save the repetition.

The main issue that $(RPI3_STOCK_ZIP_BOOT) != $(RPI3_STOCK_FW_PATH_BOOT), so if we are going to use a variable with list of files we have to append proper path in some foreach loop and call unzip for each file; or save "unzip" commandline to another variable and costruct it by appending each file, and then invoke it. Anyway, it will just add additional complexity.

jbech-linaro: Regardless what, don't over engineer this, the end goal is to delete a folder. As long as we avoid doing `rm -rf $(VARIABLE)` (i.e. a single variable) or `rm -rf $(FOO)/mypath/$(BAR)` (variable at the end and thereby sensitive to "`..`"), then I'm happy. I.e, I'm fine with what you Igor are proposing, I'm fine with having a long list of names as Jens suggest and I'm also fine with `rm -rf $(VARIABLE)/some_odd_path_name`.

igoropaniuk:
Rebased and applied tags

PR #96, opened 2017-01-05
hikey*.mk: Remove unsafe 'rm -rf' commands

vchong:
If OPTEE_TEST_OUT_PATH is redefined to some critical path/patterns, important
files might be deleted, so remove these commands which are not present on
other platforms anyway.

Also a minor change in common.mk from 'rm -rf' to 'rm -f' to remove
unnecessary recursion.

### DIFF
--- a/hikey.mk
+++ b/hikey.mk
@@ -209,7 +209,6 @@ xtest: xtest-common
 # "make clean" in xtest: fails if optee_os has been cleaned previously
 .PHONY: xtest-clean
 xtest-clean: xtest-clean-common
-	rm -rf $(OPTEE_TEST_OUT_PATH)
 
 .PHONY: xtest-patch
 xtest-patch: xtest-patch-common
--- a/common.mk
+++ b/common.mk
@@ -137,7 +137,7 @@ busybox-clean-common:
 
 busybox-cleaner-common:
 	rm -rf $(GEN_ROOTFS_PATH)/build
-	rm -rf $(GEN_ROOTFS_PATH)/filelist-final.txt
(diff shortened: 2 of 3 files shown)


vchong:
Is it better to do `rm -rf $(OPTEE_TEST_PATH)/out` instead since it seems like https://github.com/OP-TEE/optee_test/blob/master/ta/Makefile#L50-L51 doesn't really do anything?

Or submit another PR to add a clean target to https://github.com/OP-TEE/optee_test/blob/master/ta/ta_common.mk like https://github.com/linaro-swg/hello_world/blob/master/ta/Makefile#L17-L28?

hikey.mk
```diff
@@ -209,7 +209,6 @@ xtest: xtest-common
 # "make clean" in xtest: fails if optee_os has been cleaned previously
 .PHONY: xtest-clean
 xtest-clean: xtest-clean-common
-	rm -rf $(OPTEE_TEST_OUT_PATH)
```
igoropaniuk: I think it's better to change to:
```
rmdir --ignore-fail-on-non-empty $(OPTEE_TEST_OUT_PATH)
```
It's going to handle both cases: remove empty dir, or leave it in case if it's not empty(in case OPTEE_TEST_OUT_PATH is redefined to some critical path/patterns) 
What do you think?

vchong: Thanks! There's no harm changing it to that but it's essentially a no ops since it seems like the dir will not be empty. xtest-clean-common removes all the files but leaves all the subdirs intact. Plus all the other platforms don't have an extra command for xtest-clean, so might as well try to keep things inline with the others?

vchong:
Yep.

PR #100, opened 2017-01-10
hikey: Update HiKey Debian for 2.3.0 release

jbech-linaro:
- Use a newer boot image (pre-built 16.06 UEFI image)
- Remove sudo requirement when using burn-boot
- Add udev rules help, similar to the hikey.mk file
- Step up OPTEE_PKG_VERSION to 2.3-0

Known issues:
- On HiKey 4G devices it seems like the partition table is to small to
  be able fit the newer 16.06 Debian root fs. It might be that we can
  tweak the partition table a bit, but for the moment we will keep using
  the 16.03 Debian root fs.

- When running this on the device
     $ dpkg --force-all -i /tmp/linux-image-*.deb
  there will be errors like this:
     blockdev: ioctl error on BLKROSET: Operation not permitted
  According to the comments in:
     https://github.com/OP-TEE/optee_os/issues/1234
  that message only shows up on the 16.03 root fs and it seems to only
  show up when running later kernel versions. However, if you reboot the
  device, everything is working, i.e, it will boot up your new kernel
  etc. We need to investige further to find the root cause for the error
  message and to implement a proper solution without error messages.

### DIFF
--- a/hikey_debian.mk
+++ b/hikey_debian.mk
@@ -17,14 +17,14 @@ CFG_SW_CONSOLE_UART ?= 3
 # TODO: Figure out how to handle this in a better way, but we need a version
 # number with major and minor for the debian packages.
 #   <major version>.<minor version>-<package revision>
-OPTEE_PKG_VERSION ?= 2.0-1
+OPTEE_PKG_VERSION ?= 2.3-0
 
 # IP-address to the HiKey device
 IP ?= 127.0.0.1
 
 # URL to images
 SYSTEM_IMG_URL=https://builds.96boards.org/releases/reference-platform/debian/hikey/16.03/hikey-rootfs-debian-jessie-alip-20160301-68.emmc.img.gz
-BOOT_IMG_URL=https://builds.96boards.org/releases/reference-platform/debian/hikey/16.03/hikey-boot-linux-20160301-68.uefi.img.gz
+BOOT_IMG_URL=https://builds.96boards.org/releases/reference-platform/debian/hikey/16.06/hikey-boot-linux-20160629-120.uefi.img.gz
 NVME_IMG_URL=https://builds.96boards.org/releases/hikey/linaro/binaries/latest/nvme.img
 
 ################################################################################
@@ -151,7 +151,6 @@ edk2-clean: edk2-clean-common
 ################################################################################
 LINUX_DEFCONFIG_COMMON_ARCH ?= arm64
 LINUX_DEFCONFIG_COMMON_FILES ?= $(LINUX_PATH)/arch/arm64/configs/defconfig \
-				$(LINUX_PATH)/arch/arm64/configs/distro.config \
 				$(CURDIR)/kconfigs/hikey_debian.conf
 
 linux-defconfig: $(LINUX_PATH)/.config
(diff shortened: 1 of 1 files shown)


jforissier:

jbech-linaro:

jbech-linaro:
Thanks Jerome for the quick turn around time. I'm playing a bit with changing the partition table. But .. any changes there should go into the official l-loader git. So even if I might get that to work with newer root fs, I don't think we have enough time to squeeze it into the 2.3.0 release. But ... with this PR and latest on the rest ... it still works even though it looks a bit buggy.

vchong:
@jbech-linaro 

Regarding known issue 1, it should not be a partition table issue. https://github.com/96boards-hikey/l-loader/blob/master/generate_ptable.sh#L78 indicates that there should be ~3.5gb for the 4g board. I also tested with 16.06 Debian root fs and flashing is stalled as below:

PC
```
sending sparse 'system' (131068 KB)...
```
Board
```
Downloading 134213708 bytes                                                                                            
134199309 / 134213708 bytes downloaded (99%)
```

But ONLY from UEFI fastboot (jumpers 1-2 5-6 closed)! From recovery fastboot (jumpers 1-2 3-4 closed), the flashing completes successfully and boots to prompt!
```
Linux linaro-alip 4.4.0-135-arm64 #1 SMP Debian 4.4.11.linaro.135-1linarojessie1 (2016-06-28) aarch64 
````

Has anybody actually tested from UEFI fastboot on an 8g board to see if the same problem is not there? If not, I can test in a few days when mine arrives.

Also, is changing the root fs to 16.06 and forcing FROM_RECOVERY=1 a viable workaround for now? This seems like an issue with driver in hikey uefi, not build.git.

jbech-linaro:
Hi @vchong ,

> From recovery fastboot (jumpers 1-2 3-4 closed), the flashing completes successfully and boots to prompt!

Interesting, I can confirm that after the release I also did try to change the partition size. I reduced some and used a bit of the reserved, but without any success. I did't try the recovery though.

> Also, is changing the root fs to 16.06 and forcing FROM_RECOVERY=1 a viable workaround for now?
> This seems like an issue with driver in hikey uefi, not build.git.

I think we're only a few people with the old 4GB version, so if it's working on the 8GB, then it's probably better to **not** use the `FROM_RECOVERY=1` flag. Having that said, with your finding and since I believe the 4GB people are a minority, I'm fine with updating the root fs to 16.06 and mention the finding in the commit message and/or the hikey.md file.

Also a thing that has been on the todo list since the first day we get a "Debian" base up and running is the build EDK2 from source instead of picking a prebuilt binary. We're doing that for all other targets, it's just for HiKey Debian that we aren't doing so.

vchong:
Report of flashing stall on 8g board when flashing from UEFI fastboot mode but ok from recovery: https://github.com/OP-TEE/optee_os/issues/1387. Submitted https://github.com/OP-TEE/build/pull/128 to upgrade rootfs to 16.06 and add notes in hikey.md and hikey_debian.mk file to try `make recovery` if `make flash` fails.

Building EDK2 from source - still on the todo list. :s

PR #102, opened 2017-01-10
Compilation parameter order for aes-perf correction

sylvainpelissier:
Without this modification, I was not able to compile aes-perf.

### DIFF
--- a/hikey_debian.mk
+++ b/hikey_debian.mk
@@ -68,8 +68,6 @@ SYSTEM_IMG			?= $(OUT_PATH)/debian_system.img
 ROOTFS_PATH			?= $(OUT_PATH)/rootfs
 LLOADER_PATH			?= $(ROOT)/l-loader
 PATCHES_PATH			?= $(ROOT)/patches_hikey
-AESPERF_PATH			?= $(ROOT)/aes-perf
-SHAPERF_PATH			?= $(ROOT)/sha-perf
 DEBPKG_PATH			?= $(OUT_PATH)/optee_$(OPTEE_PKG_VERSION)
 DEBPKG_BIN_PATH			?= $(DEBPKG_PATH)/usr/bin
 DEBPKG_LIB_PATH			?= $(DEBPKG_PATH)/usr/lib/$(MULTIARCH)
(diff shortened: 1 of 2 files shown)


igoropaniuk:
Hi @sylvainpelissier,
Just a little notice, aes performance test is already included into xtest tool (`optee_test`)
You can invoke it `$ ./xtest --aes-perf`

hikey_debian.mk
```diff
@@ -226,7 +226,7 @@ PERF_FLAGS := CROSS_COMPILE_HOST=$(CROSS_COMPILE_NS_USER) \
 	TA_DEV_KIT_DIR=$(OPTEE_OS_TA_DEV_KIT_DIR)
 
 aes-perf: optee-os optee-client
-	$(MAKE) -C -j$(NPROC) $(AESPERF_PATH) $(PERF_FLAGS)
+	$(MAKE) -j$(NPROC) -C $(AESPERF_PATH) $(PERF_FLAGS)
```
jbech-linaro: I can't see how that ever has worked. But aes-perf hasn't been called from the `all` target. In fact both aes-perf and sha-perf has been integrated into xtest. So I think it's better to remove everything that has to do with those two in this particular makefile and just use it from xtest.

sylvainpelissier: In fact I found it very convenient to be able to build it alone for development and test.

jbech-linaro: I can't argue against that, but it's a question about maintaining it also. That was one of the reasons why we moved it to xtest and the second was to have it all in the same test application.

jforissier: +1, and in addition `-j$(NPROC)` should not have been there in the first place.
Indeed, GNU make handles the number of jobs automatically provided you always invoke sub-makefiles with `$(MAKE)` and not `make`.

So, please try `xtest --aes-perf --help`, it should work.

sylvainpelissier:
Ok I understand, I have added a commit to remove those targets.

jforissier:
`Reviewed-by: Jerome Forissier <jerome.forissier@linaro.org>`.

When you squash your commits together and add the R-b tags, please don't forget to add a comment to the commit description saying that `aes-perf` and `sha-perf` are replaced by `xtest --aes-perf` and `xtest --sha-perf`, respectively.

sylvainpelissier:
Done.

jforissier:
Hi @sylvainpelissier ,

Thanks for the update. That's not exactly how I want the commit message to be, but I can fix it myself. 

While doing that, may I change your S-o-b to be `Signed-off-by: Sylvain Pelissier <sylvain.pelissier@gmail.com>` please?  (and update the Author field to be your full name, too).

sylvainpelissier:
Yes no problem. Thank you.

jforissier:
Commit message updated and PR merged manually. Thanks!

PR #104, opened 2017-01-13
qemu: add option to enable SLiRP user networking

igoropaniuk:
See also https://github.com/linaro-swg/gen_rootfs/pull/17

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -29,6 +29,8 @@ CCACHE ?= $(shell which ccache) # Don't remove this comment (space is needed)
 QEMU_VIRTFS_ENABLE		?= n
 QEMU_VIRTFS_HOST_DIR	?= $(ROOT)
 
+# Enable SLiRP user networking
+QEMU_USERNET_ENABLE		?= n
 ################################################################################
 # Check coherency of compilation mode
 ################################################################################
@@ -196,6 +198,11 @@ QEMU_EXTRA_ARGS +=\
 	-fsdev local,id=fsdev0,path=$(QEMU_VIRTFS_HOST_DIR),security_model=none \
 	-device virtio-9p-device,fsdev=fsdev0,mount_tag=host
 endif
+
+ifeq ($(QEMU_USERNET_ENABLE),y)
+QEMU_EXTRA_ARGS +=\
+	-netdev user,id=vmnic -device virtio-net-device,netdev=vmnic
+endif
 ################################################################################
 # OP-TEE
 ################################################################################
@@ -300,6 +307,10 @@ filelist-tee-common: optee-client xtest helloworld
 		echo "file /lib/optee_armtz/8aaaf200-2450-11e4-abe2-0002a5d5c51b.ta" \
 			"$(HELLOWORLD_PATH)/ta/8aaaf200-2450-11e4-abe2-0002a5d5c51b.ta" \
 			"444 0 0" 					>> $(fl); \
+	fi
+	@if [ "$(QEMU_USERNET_ENABLE)" = "y" ]; then \
+		echo "slink /etc/rc.d/S02_udhcp_networking /etc/init.d/udhcpc 755 0 0" \
+		>> $(fl); \
 	fi
 	@echo "# Secure storage dir" 					>> $(fl)
 	@echo "dir /data 755 0 0" 					>> $(fl)
--- a/docs/qemu.md
+++ b/docs/qemu.md
@@ -4,7 +4,8 @@
 1. [Introduction](#1-introduction)
 2. [Regular build](#2-regular-build)
 3. [Install files on the device](#3-qemu-console)
-4. [Host-Guest folder sharing in QEMU/QEMUv8 configurations](#4-host-guest-folder-sharing-in-qemuqemuv8-configurations)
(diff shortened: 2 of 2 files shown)


common.mk
```diff
@@ -320,6 +329,8 @@ filelist-tee-common: optee-client xtest helloworld
 									>> $(fl)
 	@echo "slink /lib/libteec.so.1 libteec.so.1.0 755 0 0"			>> $(fl)
 	@echo "slink /lib/libteec.so libteec.so.1 755 0 0" 			>> $(fl)
+	@echo "dir /etc/udhcp 755 0 0" 			>> $(fl)
+	@echo "file /etc/udhcp/simple.script $(ROOT)/busybox/examples/udhcp/simple.script 755 0 0" >> $(fl)
```
jbech-linaro: I still think it's better to put this into the gen_rootfs, since other files from busybox are included there directly. Here we typically only include non-busybox binaries etc.

jbech-linaro:
With or without my comment addressed.

docs/qemu.md
```diff
@@ -31,19 +32,36 @@ It will stop on the QEMU console, to continue, simply
...
+build/common.mk (set `QEMU_VIRTFS_ENABLE ?= y`), adjust `QEMU_VIRTFS_HOST_DIR` and
 rebuild QEMU.
 
 To mount host folder in QEMU, simply run:
 
 ```bash
 $ mount_shared <mount_point>
 ```
+# 5. SLiRP user networking
+To enable SLiRP user networking just set `QEMU_VIRTFS_ENABLE ?= y` in `common.mk`.
```
jforissier: That's `QEMU_USERNET_ENABLE`.

docs/qemu.md
```diff
@@ -31,19 +32,36 @@ It will stop on the QEMU console, to continue, simply
...
+After booting QEMU VM, you can receive IP address via DHCP for eth0 interface by
+running:
```
jforissier: We'd rather make this completely automatic. Should be quite easy: add a file called `/etc/init.d/udhcp_networking` with the following content (untested!):
```
#!/bin/sh
#
# /etc/init.d/udhcp_networking
#
case "$1" in
    start)
        if [ /sbin/ifconfig eth0 2>/dev/null ]; then
            echo "Configuring eth0 (DHCP)..."
            /sbin/udhcpc -i eth0 -s /etc/udhcp/simple.script
            exit 0
        else
            echo "No eth0 interface found"
        fi
        ;;
    stop)
        ;;
esac
```
And add a symlink `/etc/rc.d/S02_udhcp_networking -> /etc/init.d/udhcp_networking`.

What this should hopefully do is, if the eth0 device exists (i.e., if QEMU was started with the proper network args), it will be configured using DHCP automatically.

common.mk
igoropaniuk: @jbech-linaro 
https://github.com/linaro-swg/gen_rootfs/pull/17

docs/qemu.md
igoropaniuk: @jforissier btw, you have already added similar init script
https://github.com/linaro-swg/gen_rootfs/blob/master/etc/init.d/udhcpc

jbech-linaro: Yes, even better :+1:

jforissier: Ha! Yes! I didn't even remember ;o
My script is missing the device detection part, however. It can be safely modified to add ifconfig I think.

igoropaniuk: @jforissier Check the last commit, please

docs/qemu.md
igoropaniuk: fixed

common.mk
```diff
@@ -301,6 +308,10 @@ filelist-tee-common: optee-client xtest helloworld
 			"$(HELLOWORLD_PATH)/ta/8aaaf200-2450-11e4-abe2-0002a5d5c51b.ta" \
 			"444 0 0" 					>> $(fl); \
 	fi
+	@if [ "$(QEMU_USERNET_ENABLE)" = "y" ]; then \
```
jforissier: That could probably be done unconditionally provided that `/etc/init.d/udhcpc` properly checks if `eth0` is available (which you do in your other PR, AFAICT).

igoropaniuk: I thought that it would better to avoid "No eth0 interface found" message after each boot if network was intentionally disabled.
And in this case check in ` /etc/init.d/udhcpc` can be be useful in case if someone decide to manually run `/etc/init.d/udhcpc start`. But anyway, if it's critical I can just remove it, and add slink into `filelist.txt` in `gen_rootfs`

jforissier: Yeah I have a slight preference for doing this in `gen_rootfs` but I leave it up to you. You can keep this here, I don't mind that much because it does makes sense here.

igoropaniuk:
Tags applied & rebased

igoropaniuk:
I think I've addressed all comments
@jforissier could you please merge this PR?

jforissier:
@igoropaniuk this is all good, but I'd rather merge it tomorrow after the OP-TEE release is tagged.

jforissier:
Rebased on top of master and merged. Thanks!

PR #108, opened 2017-01-18
rootfs: start tee-supplicant on boot

jforissier:
Tested-by: Jerome Forissier <jerome.forissier@linaro.org> (QEMU)

### DIFF
--- a/init.d.optee
+++ b/init.d.optee
@@ -1,0 +1,26 @@
+#!/bin/sh
+#
+# /etc/init.d/optee
+#
+# Start/stop tee-supplicant (OP-TEE normal world daemon)
+#
+case "$1" in
+    start)
+	if [ -e /bin/tee-supplicant -a -e /dev/teepriv0 ]; then
+		echo "Starting tee-supplicant..."
+		tee-supplicant&
+		exit 0
+	else
+		echo "tee-supplicant or TEE device not found"
+		exit 1
+	fi
+
+        ;;
+    stop)
+	killall tee-supplicant
+	;;
+    status)
+	cat /dev/teepriv0 2>&1 | grep -q "Device or resource busy" || not="not "
+	echo "tee-supplicant is ${not}active"
+	;;
+esac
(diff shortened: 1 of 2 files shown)


init.d.optee
```diff
@@ -0,0 +1,21 @@
...
+# Start/stop tee-supplicant (OP-TEE normal world daemon)
+#
+case "$1" in
+    start)
+	if [ -e /bin/tee-supplicant -a -e /dev/teepriv0 ]; then
+		echo "Starting tee-supplicant..."
+		tee-supplicant&
+		exit 0
+	else
+		echo "tee-supplicant or TEE device not found"
```
igoropaniuk: minor: exit 1 (or another appropriate exit code)

init.d.optee
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
+#
+# /etc/init.d/optee
+#
+# Start/stop tee-supplicant (OP-TEE normal world daemon)
+#
+case "$1" in
+    start)
+	if [ -e /bin/tee-supplicant -a -e /dev/teepriv0 ]; then
```
igoropaniuk: Based on (https://github.com/OP-TEE/optee_client/blob/master/tee-supplicant/src/tee_supplicant.c#L388):
```
if (argc == 2) {
		fd = open_dev(argv[1]);
		if (fd < 0) {
			EMSG("failed to open \"%s\"", argv[1]);
			exit(EXIT_FAILURE);
		}
```
we can support different device file names. Maybe it's better to check in the ps list?

jbech-linaro: 99.9% of the times when we're using tee-supplicant in our setup's we simply use `tee-supplicant &`. So, at this moment I don't think it's worth taking care of extra args etc, even though Igor is right in principle.

jbech-linaro:

init.d.optee
jforissier: Indeed, I'd rather keep this simple. Here we are targeting a controlled environment (our busybox-based rootfs) and we know that the TEE should be listening on `/dev/teepriv0`.

init.d.optee
jforissier: OK

init.d.optee
igoropaniuk: ok

jforissier:
Updated.

igoropaniuk:

PR #115, opened 2017-01-27
Travis enablement

jbech-linaro:
In the quest for updating to GCC6.x I wanted to investigate the ability to automate building full OP-TEE repo configurations (only build, no testing). After lots of trail and error and reading I think I've manged to come up with something that works pretty good using Travis. An example of a complete and successful build looks like this (and some of you will probably get an email when Travis has completed the build for this PR):
https://travis-ci.org/jbech-linaro/build/builds/195839345

Individual build time for each project is ~40 minutes and in total it runs for 1h 20min. I have played with enabling caches, but I don't think I've got it working as it should. Because 40 minutes is quite much time compared to local (cached) setups on my own computer. Unfortunately it's a guessing game, since there isn't that much documentation about it and you cannot (what I know) print folders etc so "see" what is in the cache. I've even played with using a repo reference, but no real success. Probably worth investigate this further when we have the base functionality in place.

There are various ways to enable when Travis should trigger builds. Pushing to branches, when creating pull request (like on optee_os) and they now also have cron job in Beta. I think I'd like to start with running a cron job on daily basis. We don't have that much activity in build.git, so by running as cron job, we will implicitly make builds containing changes in our other gits to and therefore we would know if those break the repo setups. We can of course also enable it for pull request if we think we need that and again, the low activity in build.git makes me think that it indeed would be OK to enable for pull requests.

When doing this I've also made two changes to the toolchain.mk. One patch enables parallel downloads of toolchains and prevent from re-download if already downloaded. The second patch simplifies the download a bit by wrapping it in a macro. I'm fully aware that an interrupted download will foobar this since folders will be there, but not the entire content. The likelihood that it happens is low and I'm willing to take the risk/trade-off there. Likewise I haven't bother doing any toolchains-clean target, since ... the rest of you seems to wanna do that in a more complicated way than I would have done :) So, feel free to send patches doing a nice cleanup later on.

When this has been merged, it should be easier to check the **builds** when enabling GCC6.x.

// Joakim
(I think we can update optee_os/.travis.xml with some **env** also, to simplify that file a bit also).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,62 @@
+# Travis configuration file for downloading and building OP-TEE according to the
+# instructions in build.git
+language: bash
+
+notifications:
+  email:
+    recipients:
+      - op-tee@linaro.org
+      - joakim.bech@linaro.org
+    on_success: change
+    on_failure: always
+
+dist: trusty
+sudo: required
+group: beta
+
+cache:
+  directories:
+    - $HOME/reference
+    - $HOME/.ccache
+
+git:
+  depth: 3
+
+after_script:
+  - ccache -s
+
+before_install:
+  - ccache -s
+  - sudo dpkg --add-architecture i386
+  - sudo apt-get update -qq
+  # Taken from README.md + device-tree-compiler
+  - sudo apt-get install -qq -y android-tools-adb android-tools-fastboot autoconf automake bc bison build-essential cscope curl device-tree-compiler flex ftp-upload gdisk libattr1-dev libc6:i386 libcap-dev libfdt-dev libftdi-dev libglib2.0-dev libhidapi-dev libncurses5-dev libpixman-1-dev libssl-dev libstdc++6:i386 libtool libz1:i386 make mtools netcat python-crypto python-serial python-wand unzip uuid-dev xdg-utils xterm xz-utils zlib1g-dev
+
+install: true
+
+env:
+  - $REPO_PROJ=default
+  - $REPO_PROJ=fvp
+  - $REPO_PROJ=hikey
+  - $REPO_PROJ=juno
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -0,0 +1,62 @@
...
+      - op-tee@linaro.org
+    on_success: change
+    on_failure: always
+
+dist: trusty
+sudo: required
+group: beta
+
+cache:
+  #ccache: true
```
jforissier: Uncomment this and remove `$HOME/.ccache` below?

jbech-linaro: Well, this is one of the magical things with caches, that I've been trying to get working. The thing is that if I do what you suggest, nothing really changed. **But** if I change the `language` a few lines up to `c`, then I can see that Travis picks up the `cache: ccache: true`. But, for some unknown reason I got compiler error in OP-TEE (Python Crypto) for some strange reason. See this for example: https://travis-ci.org/jbech-linaro/build/builds/195668777

So, for now, I'd like to keep `$HOME/.ccache` and instead remove the out-commented `#ccache: true` and then later try to see if it's possible to change to language `c` and resolve the issue mentioned.

jforissier: OK, sounds good.

jbech-linaro: I'll make the change when adding the tag. I'm gonna take the liberty to make yet one more change and that is to add my self to the email section. I recall reading that to receive emails one has to have the account register. So .. until we've figured out how to do that I'll have my own email there.

jbech-linaro:
Current job and results for this force push can be seen here (just started it, so ...):
https://travis-ci.org/jbech-linaro/build/builds/195915764

PR #118, opened 2017-01-31
qemu, qemu_v8: prefer gnome-terminal over xterm when available

jforissier:
gnome-terminal comes by default with Ubuntu and is much more advanced
than xterm, so use it when available.
While we're at it, move common definitions to common.mk.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -203,6 +203,42 @@ ifeq ($(QEMU_USERNET_ENABLE),y)
 QEMU_EXTRA_ARGS +=\
 	-netdev user,id=vmnic -device virtio-net-device,netdev=vmnic
 endif
+
+define run-help
+	@echo
+	@echo \* QEMU is now waiting to start the execution
+	@echo \* Start execution with either a \'c\' followed by \<enter\> in the QEMU console or
+	@echo \* attach a debugger and continue from there.
+	@echo \*
+	@echo \* To run OP-TEE tests, use the xtest command in the \'Normal World\' terminal
+	@echo \* Enter \'xtest -h\' for help.
+	@echo
+endef
+
+gnome-terminal := $(shell command -v gnome-terminal 2>/dev/null)
+xterm := $(shell command -v xterm 2>/dev/null)
+ifdef gnome-terminal
+# Note: the title option (-t) is ignored with gnome-terminal versions
+# >= 3.14 and < 3.20
+define launch-terminal
+	@nc -z  127.0.0.1 $(1) || \
+	$(gnome-terminal) -t "$(2)" -x $(SOC_TERM_PATH)/soc_term $(1) &
+endef
+else
+ifdef xterm
+define launch-terminal
+	@nc -z  127.0.0.1 $(1) || \
+	$(xterm) -title $(2) -e $(BASH) -c "$(SOC_TERM_PATH)/soc_term $(1)" &
+endef
+else
+check-terminal := @echo "Error: could not find gnome-terminal nor xterm" ; false
+endif
+endif
+
+define wait-for-ports
+	@while ! nc -z 127.0.0.1 $(1) || ! nc -z 127.0.0.1 $(2); do sleep 1; done
+endef
+
 ################################################################################
 # OP-TEE
 ################################################################################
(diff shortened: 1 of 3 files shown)


common.mk
```diff
@@ -203,6 +203,42 @@ ifeq ($(QEMU_USERNET_ENABLE),y)
...
+ifdef gnome-terminal
+# Note: the title option (-t) is ignored with gnome-terminal versions
+# >= 3.14 and < 3.20
+define launch-terminal
+	@nc -z  127.0.0.1 $(1) || \
+	$(gnome-terminal) -t "$(2)" -x $(SOC_TERM_PATH)/soc_term $(1)
+endef
+else
+ifdef xterm
+	define launch-terminal
```
vchong: nit no tab?

common.mk
```diff
@@ -203,6 +203,42 @@ ifeq ($(QEMU_USERNET_ENABLE),y)
...
+	@nc -z  127.0.0.1 $(1) || \
+	$(xterm) -title $(2) -e $(BASH) -c "$(SOC_TERM_PATH)/soc_term $(1)" &
+endef
+else
+check-terminal := @echo "Error: could not find gnome-terminal nor xterm" ; false
+endif
+endif
```
vchong: Seems like an extra endif here?

jbech-linaro:
From [this commit](https://github.com/OP-TEE/build/commit/e5425ff0898a5eabf4a6239c3e5afb6c66d569f8) which comes from a discussion on email (with Peter M IIRC), it sounded like people preferred using xterm in favor for gnome-terminal. In the first versions I had gnome-terminal as the default (and only terminal). So, even though I personally wouldn't mind using gnome-terminal as default, we've had the discussion in the past.

vchong:
Can we have a build flag to choose one or the other?

common.mk
```diff
@@ -203,6 +203,42 @@ ifeq ($(QEMU_USERNET_ENABLE),y)
...
+endef
+
+gnome-terminal := $(shell command -v gnome-terminal 2>/dev/null)
+xterm := $(shell command -v xterm 2>/dev/null)
+ifdef gnome-terminal
+# Note: the title option (-t) is ignored with gnome-terminal versions
+# >= 3.14 and < 3.20
+define launch-terminal
+	@nc -z  127.0.0.1 $(1) || \
+	$(gnome-terminal) -t "$(2)" -x $(SOC_TERM_PATH)/soc_term $(1)
```
vchong: Do we need a non-blocking `&` here also?

jbech-linaro:
> @vchong 
> Can we have a build flag to choose one or the other?

Maybe time for a `$HOME/.optee` file :wink:. And by the way Jerome, you're the maintainer, so ... do what you think is best.

common.mk
jforissier: Yup

common.mk
jforissier: No.
```
ifdef
else
  ifdef
  else
  endif
endif
```

common.mk
jforissier: Indeed! Tricky.

jforissier:
@jbech-linaro thanks for reminding me about #18. The issues discussed there were: (1) gnome-terminal blocks if not run in the background and is the first instance, and (2) gnome-terminal is not always available. The solution chosen was to replace `gnome-terminal` with `xterm`, but the current PR is better in that it addresses both issues while still preferring `gnome-terminal` by default. Or did I miss something?

jbech-linaro:
I guess you're right @jforissier , but I think the comment was that xterm is more common than gnome-terminal and xterm is "standalone", but I think gnome-terminal has dependencies to other gnome packages. But ... I'm fine with you changes, it makes saving to log-files a bit more smooth too.

common.mk
vchong: Ah, ok. I was reading:

```
ifdef
else ifdef
else
endif
endif #the extra one
```

Thanks!

vchong:
`Reviewed by: Victor Chong <victor.chong@linaro.org>`

jforissier:
Merged, thanks!

PR #120, opened 2017-02-07
Final piece of the toolchain upgrade (gcc6.x)

jbech-linaro:
This is the patch that step up default toolchain in build.git to GCC6.x. Note that I had to add a dummy patch for Travis so that it was actually checking out GCC6.x when running the Travis job. That should not be included when merging this. I'll remove that when adding R-B tags etc. However, my push looked good on my local branch, see -> https://travis-ci.org/jbech-linaro/build/builds/199189080                                                   
                                                                                                                
So, hopefully this should be it.

### DIFF
--- a/faq.md
+++ b/faq.md
@@ -65,9 +65,6 @@ the using the ARM toolchain. Please install the [OP-TEE pre-requisties] and this
 time try to ensure that you are using GCC for ARM (for more information, please
 see [Issue#846]).
 
-### I can't get OP-TEE to build using GCC 6.x?
-GCC 6.x isn't currently supported, please see [Issue#1200] for more information.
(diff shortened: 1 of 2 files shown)


jenswi-linaro:

jforissier:
Nitpicking: you could as well mention the actual version in the commit description, i.e., GCC 6.2 rather than 6.x. Anyway, that's good!

jbech-linaro:
I'll change to the correct version number as proposed when rebasing and adding RB.

jbech-linaro:
Travis was happy with the first patch set, so now I've added the tags, removed the "temporary" commit, changed to GCC6.2.

PR #125, opened 2017-02-24
travis: qemu_v8: fetch dtc submodule to avoid issues with libfdt-dev

jforissier:
If the libfdt-packe is missing or is too old, QEMU reports the
following error:

ERROR: DTC (libfdt) version >= 1.4.2 not present. Your options:
         (1) Preferred: Install the DTC (libfdt) devel package
         (2) Fetch the DTC submodule, using:
             git submodule update --init dtc

It looks like the version of libfdt-dev that is currently deployed on
travis-ci.org is too old. So, let's do (2).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,5 +58,7 @@ script:
   # in the makefile
   - if [ $REPO_PROJ == "fvp" ]; then mkdir -p $HOME/$REPO_PROJ/Foundation_Platformpkg; fi
   - cd $HOME/$REPO_PROJ && repo init -u https://github.com/OP-TEE/manifest.git -m $REPO_PROJ.xml </dev/null && repo sync -j2 --no-clone-bundle --no-tags --quiet
+  # Fetch a local copy of dtc+libfdt to avoid issues with a possibly outdated libfdt-dev
+  - if [ $REPO_PROJ == "qemu_v8" ]; then cd $HOME/$REPO_PROJ/qemu && git submodule update --init dtc; fi
   - cd $HOME/$REPO_PROJ && ln -s $HOME/reference/toolchains .
   - cd $HOME/$REPO_PROJ/build && make all -j2


.travis.yml
```diff
@@ -58,5 +58,7 @@ script:
   # in the makefile
   - if [ $REPO_PROJ == "fvp" ]; then mkdir -p $HOME/$REPO_PROJ/Foundation_Platformpkg; fi
   - cd $HOME/$REPO_PROJ && repo init -u https://github.com/OP-TEE/manifest.git -m $REPO_PROJ.xml </dev/null && repo sync -j2 --no-clone-bundle --no-tags --quiet
+  # Fetch a local copy of dtc+libfdt to avoid issues with a possibly outdated libfdt-dev
+  - if [ $REPO_PROJ == "qemu_v8" ]; then cd $HOME/$REPO_PROJ/qemu && git submodule update --init dtc; fi
```
igoropaniuk: Minor: maybe it's better just to fetch all git submodules recursively by default (as more generic solution)?
`git submodule update --init --recursive`

jbech-linaro: Should be OK, but remember that we have to consider two things here. We're using bandwidth from Travis and they have a hard limit of 50 minutes per job. So anything we can do that lowers build time is something we should try to do.

jbech-linaro:

.travis.yml
jforissier: I disagree, because:
1. All submodules in qemu/ are at the top-level, in other words: `--recursive` would be a no-op
2. More importantly, fetching more submodules than needed just brings more opportunities for failures because not all repositories are on the same server. And it's quite infuriating to see a build fail due to something that's not even needed ;)

igoropaniuk: @jforissier
`--recursive` is performed not only in the submodules of the current repo, but also in any nested submodules inside those submodules (and so on). I haven't checked if pixman or dtc has their own submodules, but that's why I prefer to use `--recursive` just to avoid checking this :)

But, anyway, I totally agree with other points, so generic solution is redundant

PR #127, opened 2017-03-03
qemu-check.exp: do not start tee-supplicant

jforissier:
Since commit 1333db4d959b ("rootfs: start tee-supplicant on boot"),
there is no need to start tee-supplicant from qemu-check.exp.


vchong:

PR #128, opened 2017-03-03
hikey_debian: Update rootfs to 16.06

vchong:
Also add notes to try 'make recovery' if 'make flash' fails

### DIFF
--- a/docs/hikey.md
+++ b/docs/hikey.md
@@ -78,7 +78,10 @@ $ make system-img
 ```
 
 The `make flash` step will tell you how you should set the jumpers on the board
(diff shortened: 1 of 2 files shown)


jforissier:
Nitpicking: the word "also" in a commit message is often a sign that the commit could (should) be split ;o)
Since this is just documentation / URL update (i.e., not code), it doesn't matter much.

vchong:
Yea was trying to get away with being a bit lazy but you caught me. :P
Commit split and tags added.

jforissier:
Thanks Victor ;)

PR #130, opened 2017-03-10
qemu_v8: Update broken link

vchong:
Change link from build 716 to latest to prevent it from becoming outdated.

### DIFF
--- a/qemu_v8.mk
+++ b/qemu_v8.mk
@@ -65,7 +65,7 @@ edk2: optee-os
 ifeq ("$(wildcard $(EDK2_BIN))","")
 	mkdir -p $(EDK2_PATH)
 	wget -O $(EDK2_BIN) \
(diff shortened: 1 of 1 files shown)


jbech-linaro:
Fix is OK, but I wonder why Travis still fails and shows the old link!?

jforissier:
@jbech-linaro `.travis.yml` downloads `manifest.git` using `repo`, but then, where is the `build/` directory updated to point to the one that should be verified? It seems that this step is missing.

jbech-linaro:
You're right Jerome, now it comes back to me. The reason I did like that was thought by doing, so you would only have to download the three toolchains once for all subprojects. But in fact, I still think it will downloads all three for all 8 builds, so I think it's better to simply remove that and just rely on the build.git from repo. That's the first thing, the second thing is that we should bring down the patch in question and apply it. In any case, I think it's safe to just merge Victors patch.

jforissier:
Right, @vchong please apply Joakim's R-b and I'll merge this patch.

vchong:
Tag added and ok to merge.
Thanks!

PR #133, opened 2017-03-29
qemu: fix linux kernel config regarding SDP support

etienne-lms:
The first 1Mbyte of the secure RAM is reserved to the ARM-TF on
the qemu ARMv8 setup. To support both QEMU ARMv7 and QEMU ARMv8
with the same kernel configuration and with the secure data path
support, decrease the SDP memory pool to 2Mbyte instead of 3Mbyte,
starting its base address after the ARM-TF reserved RAM.


jbech-linaro:
2MB is sufficient from what we know as of now? (I guess 3MB was just a guess too?)

etienne-lms:
@jbech-linaro, 2MB is sufficient to test SDP for regressions.
To experience SDP on a specific usecase... it will depend on the usecase. Secure video playback will likely require far more. secure audio stream should be ok. secure dummy processing is fine with few kByte. Will the qemu be used to test secure video playback?
To increase SDP pool size or change its location, one can still modify its linux config and its op-tee config, without needing to change the implementations.

jbech-linaro:
> To experience SDP on a specific usecase... it will depend on the usecase.

Yepps, we're on the same page here.

etienne-lms:
Updated: SDP memory pool moved to end of the QEMU secure RAM and increased to 4Mbyte.

Dependencies on: https://github.com/linaro-swg/bios_qemu_tz_arm/pull/11 and https://github.com/OP-TEE/optee_os/pull/1447

jenswi-linaro:

etienne-lms:
rebased, tag applied.
should be merged together with https://github.com/linaro-swg/bios_qemu_tz_arm/pull/11 and https://github.com/OP-TEE/optee_os/pull/1447

PR #138, opened 2017-04-06
Travis and QEMU v8 fixes

jbech-linaro:
This fixes the build issue for v8 and also makes use of the actual PR/push commit/changes in build.git when running Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ notifications:
   email:
     recipients:
       - op-tee@linaro.org
-      - joakim.bech@linaro.org
     on_success: change
     on_failure: always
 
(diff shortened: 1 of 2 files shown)


jenswi-linaro:
First two commits:
The last "travis: use commit from push/pr branch"

jbech-linaro:
So yet again the v8 build fails and after checking the log it seems like on a pull request, the `https://github.com/$TRAVIS_REPO_SLUG` and `$TRAVIS_BRANCH` points to the OP-TEE/optee_os.git and master. I.e, the destination git and branch and not the source git and branch. I can probably fix that by also taking `$TRAVIS_PULL_REQUEST_SLUG` into account as stated [here](https://docs.travis-ci.com/user/environment-variables/). However, I'd like to do that as a separate fix. After merging this three patches here, the v8 branch will/should work again and that would give us a "green" light for the release.

Issue #146, opened 2017-05-21
OP-TEE build failure for HiKey board


ykbs0903:
Hi, 
I have encountered an error during OP-TEE build (for HiKey debian)

After "#make all run" command in "_7.5 Build the solution_" section (https://github.com/OP-TEE/build),
i got the message below.

**< build log >**
...
...
...
make[3]: Entering directory '/root/devel/optee/optee_test/ta/socket'
  CC      /root/devel/optee/build/../optee_test/out/ta/socket/ta_entry.o
  CC      /root/devel/optee/build/../optee_test/out/ta/socket/user_ta_header.o
  CPP     /root/devel/optee/build/../optee_test/out/ta/socket/ta.lds
  LD      /root/devel/optee/build/../optee_test/out/ta/socket/873bcd08-c2c3-11e6-a937-d0bf9c45c61c.elf
  OBJDUMP /root/devel/optee/build/../optee_test/out/ta/socket/873bcd08-c2c3-11e6-a937-d0bf9c45c61c.dmp
  OBJCOPY /root/devel/optee/build/../optee_test/out/ta/socket/873bcd08-c2c3-11e6-a937-d0bf9c45c61c.stripped.elf
  SIGN    /root/devel/optee/build/../optee_test/out/ta/socket/873bcd08-c2c3-11e6-a937-d0bf9c45c61c.ta
make[3]: Leaving directory '/root/devel/optee/optee_test/ta/socket'
make[2]: Leaving directory '/root/devel/optee/optee_test/ta'
make[1]: Leaving directory '/root/devel/optee/optee_test'
make -C /root/devel/optee/build/../hello_world HOST_CROSS_COMPILE="/usr/bin/ccache /root/devel/optee/build/../toolchains/aarch64/bin/aarch64-linux-gnu-" TA_CROSS_COMPILE="/usr/bin/ccache /root/devel/optee/build/../toolchains/aarch32/bin/arm-linux-gnueabihf-" TA_DEV_KIT_DIR=/root/devel/optee/build/../optee_os/out/arm/export-ta_arm32 TEEC_EXPORT=/root/devel/optee/build/../optee_client/out/export
make[1]: Entering directory '/root/devel/optee/hello_world'
make -C host CROSS_COMPILE="/usr/bin/ccache /root/devel/optee/build/../toolchains/aarch64/bin/aarch64-linux-gnu-"
make[2]: Entering directory '/root/devel/optee/hello_world/host'
[...]

PR #147, opened 2017-05-21
README: Remove `run` target

vchong:
Only QEMU, FVP and MTK8173 have a `run` target but all *.md files
refer to README Section 7.5 which calls for `make all run`. Remove
this target and add it to mtk8173.md instead. qemu.md and fvp.md
already mentions the run target so don't need to be added again.

Fixes: https://github.com/OP-TEE/build/issues/146

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ symlink the `Makefile` to the correct device specific makefile, that means that
 you simply start the build by running:
 
 ```bash
-$ make all run
+$ make all
 ```
 This step will also take some time, but you can speed up subsequent builds by
 enabling [ccache] (again see Tips and Tricks).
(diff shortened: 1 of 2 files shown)


jbech-linaro:
For this, I'd recommend to keep `run` for software targets, i.e., `QEMU v7/v8` and `FVP` and for the rest rename current `run` to `flash` instead. But based on you comment, it seems like only `MTK8173` needs a change (besides the `README.md` that you have already changed).

README.md
```diff
@@ -163,7 +163,7 @@ symlink the `Makefile` to the correct device specific makefile, that means that
 you simply start the build by running:
 
 ```bash
-$ make all run
+$ make all
```
jforissier: `all` should be the default target so `make` should be enough.

Related to this: I think we need another patch to `s/all-clean/clean/` because `all-clean` is just stupid. Everyone expects to find a `clean` target.

vchong:
Replaced by https://github.com/OP-TEE/build/pull/148. Comments addressed there. Seems easier to do that than to continue here.

PR #156, opened 2017-07-03
qemu_v8: recent QEMU requires libfdt 1.4.2 or higher

etienne-lms:
Ubuntu 1.4.04 LTS, 16.04 LTS and even 17.04 provide the libfdt in
version 1.4.0. This version is too old for QEMU that requires libfdt
in version 1.4.2 or higher.

This change insures DTC package is locally clone so that QEMU builds
over it rather than using the libfdt support and tools from the system.

Suggested-by: Joakim Bech <joakim.bech@linaro.org>

### DIFF
--- a/qemu_v8.mk
+++ b/qemu_v8.mk
@@ -65,7 +65,11 @@ arm-tf-clean:
 # QEMU
 ################################################################################
 qemu:
-	cd $(QEMU_PATH); ./configure --target-list=aarch64-softmmu\
+	cd $(QEMU_PATH); \
+		dpkg --compare-versions 1.4.2 le \
+		    `dpkg-query --showformat='$${Version}' --show libfdt-dev` || \
+		    git submodule update --init dtc && \
+		./configure --target-list=aarch64-softmmu \
 			$(QEMU_CONFIGURE_PARAMS_COMMON)
 	$(MAKE) -C $(QEMU_PATH)
 
--- a/qemu_v8.mk
+++ b/qemu_v8.mk
@@ -66,9 +66,7 @@ arm-tf-clean:
 ################################################################################
 qemu:
 	cd $(QEMU_PATH); \
-		dpkg --compare-versions 1.4.2 le \
-		    `dpkg-query --showformat='$${Version}' --show libfdt-dev` || \
-		    git submodule update --init dtc && \
+		git submodule update --init dtc && \
 		./configure --target-list=aarch64-softmmu \
 			$(QEMU_CONFIGURE_PARAMS_COMMON)
 	$(MAKE) -C $(QEMU_PATH)
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,8 +53,6 @@ script:
   # in the makefile.
   - if [ $REPO_PROJ == "fvp" ]; then mkdir -p $HOME/$REPO_PROJ/Foundation_Platformpkg; fi
   - cd $HOME/$REPO_PROJ && repo init -u https://github.com/OP-TEE/manifest.git -m $REPO_PROJ.xml </dev/null && repo sync -j2 --no-clone-bundle --no-tags --quiet
-  # Fetch a local copy of dtc+libfdt to avoid issues with a possibly outdated libfdt-dev
-  - if [ $REPO_PROJ == "qemu_v8" ]; then cd $HOME/$REPO_PROJ/qemu && git submodule update --init dtc; fi
   # Here we are using Travis environment variables to select the correct branch
   # based on either a pull request or a normal push.
   - |
--- a/qemu.mk
+++ b/qemu.mk
@@ -53,7 +53,9 @@ bios-qemu-clean:
 	$(call bios-qemu-common) clean
 
 qemu:
-	cd $(QEMU_PATH); ./configure --target-list=arm-softmmu\
+	cd $(QEMU_PATH); \
+		git submodule update --init dtc && \
+		./configure --target-list=arm-softmmu \
 			$(QEMU_CONFIGURE_PARAMS_COMMON)
 	$(MAKE) -C $(QEMU_PATH)
 


qemu_v8.mk
```diff
@@ -65,7 +65,11 @@ arm-tf-clean:
 # QEMU
 ################################################################################
 qemu:
-	cd $(QEMU_PATH); ./configure --target-list=aarch64-softmmu\
+	cd $(QEMU_PATH); \
+		dpkg --compare-versions 1.4.2 le \
```
jbech-linaro: Do we even need to compare etc? It seems like QEMU defaults to the locally checked out version if there is any, otherwise it will use the hosts version. So, I guess line 71 only is sufficient.

jbech-linaro:
Also, with this change (working), you could also remove the lines doing the same in .travis.yml.

qemu_v8.mk
etienne-lms: Ok, always rely on the cloned dtc package.

etienne-lms:
By the way, I wonder if the same change should be required on qemu_virt (default.mk/default_stable.mk). These currently rely on old qemu and we should upgrade it some day.

jbech-linaro:
> By the way, I wonder if the same change should be required on qemu_virt (default.mk/default_stable.mk).

Yeah, I think that is better safe than sorry. Anyhow, for the current patches.

qemu_v8.mk
```diff
@@ -65,7 +65,9 @@ arm-tf-clean:
 # QEMU
 ################################################################################
 qemu:
-	cd $(QEMU_PATH); ./configure --target-list=aarch64-softmmu\
+	cd $(QEMU_PATH); \
+		git submodule update --init dtc && \
```
jforissier: I'd rather avoid changing the source tree in the build step, if possible. Could this be done before, such as in `manifest.xml`?

jbech-linaro: I'm not sure if that can be done by repo. I some other cases in the past, I've added a check for the existence of a file created by some command to avoid doing the same operation over and over again. Having that said, by doing so it would still be a step/check done in the build step.

igoropaniuk: @jbech-linaro 
I'll probably agree with Jerome,
It could be done by adding default attribute to the manifest (based on info from [manifest-format.txt](https://gerrit.googlesource.com/git-repo/+/master/docs/manifest-format.txt#171))
```
<default sync-s="true"/>
```
And to sync git reps + all submodules:
```
repo sync --fetch-submodules
```

etienne-lms: Tested: adding `sync-s="true"` to the qemu project only seems to do the job through command `repo sync`. I.e, in qemu_v8.xml:
```patch
-	<project remote="qemu" path="qemu" name="qemu.git" />
+	<project remote="qemu" path="qemu" name="qemu.git" sync-s="true" />
```

jbech-linaro: Doesn't seem that optimal to have it there. Another option would be to write a simple "init" script, but I'm not super keen on doing that. Just running the make is quite nice when you have got the code. What about an "init" target to the make file? The "qemu" target would depend on that for example?

jforissier: @jbech-linaro do you mean it will also clone submodules that are not needed? It is true and indeed it's a problem because some of them are quite big (u-boot). Still, having `make` start by downloading stuff kind of makes me sick... ;) so let's try something else. What about:

```diff
        <!-- QEMU -->
-       <project remote="qemu" path="qemu" name="qemu.git" />
+       <project remote="qemu" path="dtc" name="dtc.git" />
+       <project remote="qemu" path="qemu" name="qemu.git">
+               <linkfile src="../dtc" dest="dtc" />
+       </project>
```
?

etienne-lms: qemu already creates the sub-directory **./dtc**. The linkfile directive above may fail.
Aside this, I am fine with this solution.

jforissier: qemu should not create dtc/ if `git submodule` is not called? I can double-check this.

jbech-linaro: This also works, it's a trade-off. I think your proposal is better than having `sync-s="true"`, since with your proposal, it's only a symlink that needs to be created if you clean a lot.

Odd, that you are in favor of using repo and I'm not :)

Related, I think there is something fishy with the QEMU-v8 setup in general. Meanwhile doing other stuff today I've built QEMU-v8 in the background on different computers with different manifests (latest, old_stable, 2.5.0-rc1 etc, with and without Liangs ARM-TF rebase) and it behaves very weird. It hangs almost immediately on both computers. Having that said, I have had some test runs working. So I'm staring to wonder if there is an issue showing up from time to time. On my side it seems to fail more often. Probably time for gdb ... or printfs.

jforissier:
> Odd, that you are in favor of using repo and I'm not :)

Troll! :D

qemu_v8.mk
etienne-lms: A have the same issues as @jbech-linaro: with this DTC issue resoved, optee/qemu_v8 tests are failing. Hangs during optee inits or hangs when exercising xtest.

jforissier: @etienne-lms 
> qemu already creates the sub-directory ./dtc. The linkfile directive above may fail.

You are right. Everything seems to work fine with this simple change instead:
```diff
        <!-- QEMU -->
        <project remote="qemu" path="qemu" name="qemu.git" />
+       <project remote="qemu" path="qemu/dtc" name="dtc.git" />
```

igoropaniuk: @jforissier 

```patch
        <!-- QEMU -->
-       <project remote="qemu" path="qemu" name="qemu.git" />
+       <project remote="qemu" path="dtc" name="dtc.git" />
+       <project remote="qemu" path="qemu" name="qemu.git">
+               <linkfile src="../dtc" dest="qemu/dtc" />
+       </project>
```

This will probably brake qemu git local rep, at least if someone decide to update dtc submodule manually within qemu rep. After syncing this manifest file, have you tried to do this? I'm curious what will happen
```
git submodule update --init dtc
```

jforissier: @igoropaniuk yeah ignore my first proposal, it's broken because `repo sync` errors out due to `qemu/dtc` exists already.
FYI, I tried what you are asking with my second proposal (https://github.com/OP-TEE/build/pull/156#discussion_r125310828) and it's alright:
```
$ cd eqmu
$ git submodule update --init dtc
Submodule 'dtc' (git://git.qemu-project.org/dtc.git) registered for path 'dtc'
Submodule path 'dtc': checked out '558cd81bdd432769b59bff01240c44f82cfb1a9d'
```

jbech-linaro:
> edited for info: using the qemu version from tag optee-2.4.0 (qemu commit 5fe2339e6b09...), i can run optee and its tests for the qemu_armv8 setup.

I think I've managed to git bisect to the QEMU commit where it starts failing. I've been booting up QEMU and starting xtest between 5-10 times before marking it either as good or bad and I ended up with this:
```patch
$ git bisect bad                                                                                                                                                                                                                                                                 
e75449a346bf558296966a44277bfd93412c6da6 is the first bad commit                                                                                                                                                                                                                 
commit e75449a346bf558296966a44277bfd93412c6da6                                                                                                                                                                                                                                  
Author: Emilio G. Cota <cota@braap.org>                                                                                                                                                                                                                                          
Date:   Fri Apr 28 14:59:23 2017 -0400                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                 
    target/aarch64: optimize indirect branches                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                 
    Measurements:                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                 
    [Baseline performance is that before applying this and the previous commit]                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                 
    -                                    NBench, aarch64-softmmu. Host: Intel i7-4790K @ 4.00GHz
...
diff --git a/target/arm/translate-a64.c b/target/arm/translate-a64.c                                                                  
index ab61d96099..860e279658 100644                                
--- a/target/arm/translate-a64.c 
+++ b/target/arm/translate-a64.c 
@@ -11367,8 +11367,7 @@ void gen_intermediate_code_a64(ARMCPU *cpu, TranslationBlock *tb)                                             
             gen_a64_set_pc_im(dc->pc);                            
             /* fall through */  
         case DISAS_JUMP:        
-            /* indicate that the hash table must be used to find the next TB */                                                      
-            tcg_gen_exit_tb(0); 
+            tcg_gen_lookup_and_goto_ptr(cpu_pc);                  
             break;              
         case DISAS_TB_JUMP:     
         case DISAS_EXC:       
```
The patch contains just a oneliner, but without out understanding of QEMU it is hard to say anything about it.

I've also played with running latest and just revert that particular patch and since I did that I simply cannot repeat issue (just hangs otherwise).

Maybe @pm215 would like to have a look or know what it could be about? To reproduce, simply checkout a [QEMU v8 OP-TEE environment](https://github.com/OP-TEE/build#6-manifests) (and the DTC fix that this PR is about is also needed) and [run xtest](https://github.com/OP-TEE/build#79-run-xtest).



@jforissier , If this indeed is the bad commit I found, then I think we should use the commit prior to that in our manifest when tagging, i.e, this one:
```
e78722368c (HEAD, refs/bisect/good-e78722368c721f3c5b8109ed525adac1653ae97b) target/aarch64: optimize cross-page direct jumps in softmmu
```

Finally it would be good if someone else could test it, to see if you agree with me here. It was a bit tricky to rebase since it was working a few times even with the bad patch. Not often, but it happened.

etienne-lms:
While bisecting and playing with the qemu history, I went to another faulty commit:
*  7e6478e7d4f2c4b  "Check the return value of fcntl in qemu_set_cloexec"
Reverting this commit from qemu latest head... does not help optee tests.

I think there are 2 unrelated issues. One that prevent input console from being functional, and one that makes the xtest hanging at some place.


The faulty commit found by @jbech-linaro (e75449a346bf5 "target/aarch64: optimize indirect branches", committed 5th of June) is not far from the one I found, but on another branch (it is hard to follow the history through all these merges).


And.... reverting both the commit found by Joakim (e75449a) and the one pointed above (7e6478e7d4f) from the latest merged head (82d76dc7fc19a) make the optee xtest functional again. Don't ask me why.

Yet, here is the most recent merge commit I found (from the 7th of June) over which I successfully ran the optee env/test:
* 0db1851becbefe "Merge remote-tracking branch 'remotes/vivier/tags/m68k-for-2.10-pull-request' into staging"

pm215:
I'm told that commit e75449a346bf55 was indeed an upstream regression, but it should also now be fixed upstream, by commit 8da54b2507c. I'm not sure why commit 7e6478e7d4 (the fcntl one) would cause problems unless you're actually triggering the new asserts that it adds...

etienne-lms:
> I'm not sure why commit 7e6478e7d4 (the fcntl one) would cause problems

From my end user view, the effect seems that the console prints the log but fails the get input characters. And that reverting it restores input console. I cannot tell more.

jforissier:
Replace by https://github.com/OP-TEE/manifest/pull/57.

pm215:
On 4 July 2017 at 13:05, Jérôme Forissier <notifications@github.com> wrote:
>
> Closed #156.
>

Does that mean you've resolved the issues with regressions with
recent QEMU, or is that still an outstanding thing we need to
look into on our side?

thanks
-- PMM

jforissier:
@pm215 
> Does that mean you've resolved the issues with regressions with recent QEMU, or is that still an outstanding thing we need to look into on our side?

No, I'm closing this because QEMU regressions is not the topic of this PR. @jbech-linaro or @etienne-lms would you please open a separate issue? Thanks.

jbech-linaro:
> Does that mean you've resolved the issues with regressions with
> recent QEMU, or is that still an outstanding thing we need to
> look into on our side?

@pm215 , I guess we still need to do some additional testing. The issue itself here was about adding a newer libfdt version to our QEMU setup and when we did that we faced this other issue when running latest from upstream. So one issue became two. I'll have a look at the commits you mention and see if I can come to any conclusion.

jbech-linaro:
> @jbech-linaro or @etienne-lms  would you please open a separate issue? Thanks.

Yes, will do that if needed.

PR #158, opened 2017-07-04
fvp: use gicv3 legacy mode

lackan:
The default value of FVP_USE_GIC_DRIVER is FVP_GICV3,
now we set it to FVP_GICV3_LEGACY.

Signed-off-by: Guanchao Liang <liang.guanchao@linaro.org>


jenswi-linaro:

jbech-linaro:
Tested together with https://github.com/linaro-swg/arm-trusted-firmware/pull/6
`Tested-by: Joakim Bech <joakim.bech@linaro.org>` (FVP)

jforissier:
Care to say *why* we do that in the commit log? I suppose it is because ARM-TF has changed its default behavior and therefore this is needed to be able to use https://github.com/linaro-swg/arm-trusted-firmware/tree/optee_arm-tf_2017-06-22 without making any change in OP-TEE.

lackan:
@jforissier OK, I will detail the reason.

lackan:
Update.

jenswi-linaro:
Didn't the new GIC mode require changes in UEFI also?

lackan:
@jenswi-linaro Yes, if we use GICV3, UEFI should also change. You mean UEFI should also add to the commit log?

jenswi-linaro:
@lackan Yes, please mention UEFI in the commit message also. Without that you could get the impression that it would have been easier to just update OP-TEE.

jbech-linaro:
I can't recall that I did any changes for UEFI, but it all worked fine ... I'm confused.

jenswi-linaro:
Yes, that's also the point with `FVP_USE_GIC_DRIVER=FVP_GICV3_LEGACY`. It preserves the previous behavior, without it other components would need to be updated, OP-TEE OS, UEFI and DTB (I assume).

jbech-linaro:
Ah, yes of course, don't know what I was thinking.

lackan:
Thanks @jenswi-linaro @jbech-linaro

PR #164, opened 2017-07-14
rpi3: update jtag debugging details

igoropaniuk:
Add Bus Bluster config/usage details
Add physical memory map


jforissier:
LGTM but could some RPi3 user double check this please? @jbech-linaro maybe? Thanks.

jbech-linaro:
Yes, I can give it a try in the near future ... and speaking about the changes here. In our guide we are referring to the the OpenOCD fork from Sequitur Labs. Things have changed since I wrote that guide, today there is support for ARMv8-A in the official OpenOCD fork, so we should change accordingly. I don't say that we need to do it in this PR, but the sooner the better.

docs/rpi3.md
```diff
@@ -330,13 +330,25 @@ that we have also connected a USB FTDI to UART cable to a few more pins.
...
+```
+
 We are using the [Sequitur Labs OpenOCD] fork, simply clone that to your
 computer and then building is like a lot of other software, i.e.,
 ```bash
 $ ./bootstrap
 $ ./configure
 $ make
 ```
+In case if a jtag debugger which needs legacy ft2332 support is going to be
```
etienne-lms: 'In case if' looks redontant.

igoropaniuk:
Deprecated

PR #173, opened 2017-08-09
Add hikey960 doc and update hikey doc

vchong:


jforissier:
Since you're updating `hikey.md`: I think there is a mistake in https://github.com/OP-TEE/build/blob/master/docs/hikey.md#3-supported-hikey-boards. Where does the "Kirin 620" name come from? To the best of my knowledge, this is the name of the SoC, not the board. The board is still "HiKey". So I would suggest to remove the first column.
Also in the same table, it might be useful to add "Green solder mask" and "Black solder mask" in the comments (CircuitCo boards are green, LeMaker ones are black -- well as far as I know!).

vchong:
`hikey.md` updated per comment.

jforissier:
Thanks. 

vchong:
Squashed and tag added. Thanks!

PR #175, opened 2017-08-24
qemu: update ION configuration for SDP tests on linux 4.12

etienne-lms:
Ion driver in Linux 4.12 no more provides a dummy device driver for
test purpose. Instead, a generic ion device driver is loaded when
ion is enabled. One shall enable the expected ion heaps from the
linux kernel configuration. Here we enable only the unmapped heap.

For testing a SDP memory pool, we still use the same test setup where
the SDP memory pool location is hard coded into the Linux configuration.

**Do not merge until optee relies on k4.12 with updated ION drivers.**


jforissier:
I suppose this PR is obsolete since https://github.com/OP-TEE/build/pull/192 is merged?

etienne-lms:
Almost obsolete. The ION_DUMMY config should be removed for cleanup (no expect bug).

PR #155, opened 2017-06-29
Use config.txt to enable JTAG instead of U-Boot

jogarock:
See the Issue I filed in optee_os: https://github.com/OP-TEE/optee_os/issues/1634

Any merge should be synchronized with the PR I opened in linaro-swg/u-boot: https://github.com/linaro-swg/u-boot/pull/1

PR #177, opened 2017-08-25
rpi3: update jtag debugging details

igoropaniuk:
1. Add Bus Bluster config/usage details
2. The Raspberry Pi 3 allows to enable JTAG on the GPIO via the
enable_jtag_gpio=1 flag in the config.txt, which is more prefarable.
Switch to this method
3. Add physical memory map
4. Change information about usage of Sequitur labs OpenOCD fork to
the uptsteam OpenOCD, as it already has ARMv8 support

Relates: https://github.com/linaro-swg/u-boot/pull/2
Deprecates: https://github.com/OP-TEE/build/pull/164
https://github.com/OP-TEE/build/pull/155
```
Signed-off-by: Igor Opaniuk <igor.opaniuk@linaro.org>
Tested-by: Igor Opaniuk <igor.opaniuk@linaro.org>
Suggested-by: Joseph Pantoga <jp@redballoonsecurity.com>
```


jbech-linaro:
I think this is all good, but since since you are touching lots of different things it would probably make more sense to split up the patches into several patches instead.

igoropaniuk:
@jbech-linaro yeap, but all these changes relate to debugging/jtag usage, so I thought why not to have everything in one commit

And some information from Joseph's suggested changes contradicts from with my changes, so that was the second reason to squash everything into one patch and remove redundant changes

igoropaniuk:
Applied tags & rebased. Ready for merge

PR #185, opened 2017-09-08
hikey_debian: Copy CAs to the proper location

vchong:
Commit 584efe5 ("examples: use optee_examples rep") copies CAs to a wrong
location.


vchong:
Tag added. Thanks!

PR #186, opened 2017-09-11
Make include directive of toolchain.mk and  common.mk compulsory

vchong:


vchong:
@jforissier Is this worth merging?

jbech-linaro:
I think it is used everywhere, so why not, but on the other hand has there been any issue with it as it is?

vchong:
Not really, not unless someone (un)intentionally go and remove the included files and doesn't know what to do next. I suppose that's rare so we can probably drop this.

jforissier:
Yeah, one way or the other is OK, really, but still, the version without "-" is more correct. So I don't mind merging this at all.

vchong:
Let's do the more correct thing then. ; )
Tag added. Thanks!

PR #187, opened 2017-09-11
hikey+debian: Update instructions for latest repos

vchong:


vchong:
Depends on https://github.com/OP-TEE/manifest/pull/74.

hikey.mk
```diff
@@ -15,6 +15,9 @@ CFG_SW_CONSOLE_UART ?= 3
 # eMMC flash size: 8 or 4 GB [default 8]
 CFG_FLASH_SIZE ?= 8
 
+# URL to images
+NVME_IMG_URL=https://builds.96boards.org/releases/hikey/linaro/binaries/latest/nvme.img
```
jenswi-linaro: Can we get a fixed version for this?

hikey.mk
```diff
@@ -266,72 +292,124 @@ grub-flags := CC="$(CCACHE)gcc" \
...
 .PHONY: lloader-clean
-lloader-clean:
-	$(MAKE) -C $(LLOADER_PATH) clean
+lloader-clean: lloader-bin-clean lloader-ptbl-clean
 
 ################################################################################
 # nvme image
 ################################################################################
 .PHONY: nvme
 nvme: prepare
```
jenswi-linaro: Wouldn't it be preferable to use the `$(NVME_IMG)` target instead to avoid downloading this file again and again?

hikey.mk
vchong: nvme.img is essentially a never-changing empty file (all 0s) used to clear the nvme partition. As such, it might be better to keep it at 'latest' to avoid a specific version (links) getting archived or removed.

hikey.mk
vchong: The `ifeq` condition below the target avoids further downloads. Also to keep format/syntax/style consistent across all `hikey*.mk` and similar targets like `system-img`.

hikey.mk
jforissier: Then the command should be: `dd if=/dev/zero of=tutu bs=1K count=128`, not `wget ...`.

BTW, it's weird, I have an old `nvme.img` which is far from being all 0s:
```
$ strings nvme.img 
SWVERSI
T8300
BOARDID
MACADDR
WEBPWD
IMEI
IMEI
RSV6
RSV7
RSV8
...
```

I wonder what its purpose is.

hikey.mk
jforissier: OK, perhaps this format was not a good choice. The downside is that `make nvme.img` won't work. Such a cleanup might be the object of a separate PR, though.

hikey.mk
vchong: Can change to `dd` if that's preferred?

`nvme` partition is used to store uefi non-volatile variables, so your copy might be for a time when some of these variables were required on boot, but not anymore.

jforissier: Thanks for the info.
I definitely prefer `dd`, and if you could add a comment saying 'nvme partition is used to store UEFI non-volatile variables` that would be great!

vchong: You're welcome!
I pushed a separate commit for changing `wget` to `dd` and adding the comment. Please check if it's ok. Thanks!

vchong: On 2nd thought, maybe a separate commit is not required. Lemme know if so so I can squash it later.

jforissier:
For "hikey+debian: Build nvme.img instead of downloading a prebuilt":

IMO there are too many unrelated changes in the first commit. Hard to tell what it does, and the commit comment does not help.

vchong:
Ok. I'll break it out into smaller pieces wherever possible.

vchong:
Replaced by #191.

PR #189, opened 2017-09-14
qemu: bios load OP-TEE image v2

jenswi-linaro:
Bios loads OP-TEE image v2 binaries instead of the v1 format.

Depends on https://github.com/linaro-swg/bios_qemu_tz_arm/pull/13

### DIFF
--- a/qemu.mk
+++ b/qemu.mk
@@ -15,6 +15,7 @@ override COMPILE_S_KERNEL  := 32
 ################################################################################
 BIOS_QEMU_PATH			?= $(ROOT)/bios_qemu_tz_arm
 QEMU_PATH			?= $(ROOT)/qemu
+BINARIES_PATH			?= $(ROOT)/out/bin
 
 SOC_TERM_PATH			?= $(ROOT)/soc_term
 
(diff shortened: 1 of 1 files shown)


qemu.mk
```diff
@@ -41,13 +41,15 @@ define bios-qemu-common
...
-		BIOS_SECURE_BLOB=$(OPTEE_OS_BIN) \
 		PLATFORM_FLAVOR=virt
 endef
 
 bios-qemu: update_rootfs optee-os
+	ln -sf $(OPTEE_OS_HEADER_V2_BIN) .
+	ln -sf $(OPTEE_OS_PAGER_V2_BIN) .
+	ln -sf $(OPTEE_OS_PAGEABLE_V2_BIN) .
+	ln -sf $(LINUX_PATH)/arch/arm/boot/zImage .
+	ln -sf $(GEN_ROOTFS_PATH)/filesystem.cpio.gz rootfs.cpio.gz
```
etienne-lms: This fills the **build/** directory with the optee images.
Could we do as the qemu/v8, store bins in a `$(BIOS_QEMU_PATH)/build/release` and run qemu for the later?

jenswi-linaro: Except that we don't have any such path here as we're not building ARM-TF.
How about putting it in `out/` or `out/bin` instead?

etienne-lms: out/ or out/bin are fine. Thanks.

jenswi-linaro:
Update

etienne-lms:
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` (/!\ dependency)

jenswi-linaro:
Tag applied

PR #154, opened 2017-06-10
hikey_debian: Update build for 2.5.0 release candidate

vchong:
Depends on https://github.com/OP-TEE/manifest/pull/55

PR #191, opened 2017-09-15
hikey+debian: Update instructions for latest repos - v2

vchong:
Replaces #187


vchong:
Depends on OP-TEE/manifest#74

hikey.mk
```diff
@@ -133,26 +133,40 @@ busybox-cleaner: busybox-clean-common busybox-cleaner-common
...
-	EDK2_VARS += EDK2_MACROS="-DSERIAL_BASE=0xF8015000"
+	EDK2_BUILDFLAGS += -DSERIAL_BASE=0xF8015000
 endif
 
 define edk2-call
-	GCC49_AARCH64_PREFIX=$(LEGACY_AARCH64_CROSS_COMPILE) \
-	$(MAKE) -j1 -C $(EDK2_PATH) \
-		-f HisiPkg/HiKeyPkg/Makefile $(EDK2_VARS)
+	$(EDK2_TOOLCHAIN)_$(EDK2_ARCH)_PREFIX=$(LEGACY_AARCH64_CROSS_COMPILE) \
+	build -n `getconf _NPROCESSORS_ONLN` -a $(EDK2_ARCH) \
```
jforissier: Not a big deal, but this `build` script (or function? I wasn't even able to find it in the edk2 sources!) will break proper parallel build. First problem: you are hard-coding _NPROCESSORS_ONLN, so it becomes impossible to build in a single thread (useful for troubleshooting). Second problem: you will likely end up with more build threads than specified on the command line. For instance `make -j2` should never run more than 2 subprocesses.

My suggestion: if there is a way to invoke $(MAKE) directly it's much better, i.e., simply remove the `-j1` and `make` will know the proper number of sub-processes it is allowed to spawn by communicating with the parent `make` process.

hikey.mk
```diff
@@ -292,20 +292,28 @@ grub-flags := CC="$(CCACHE)gcc" \
 	TARGET_RANLIB="$(AARCH64_CROSS_COMPILE)ranlib" \
 	TARGET_STRIP="$(AARCH64_CROSS_COMPILE)strip"
 
+GRUB_MODULES += boot chain configfile echo efinet eval ext2 fat font gettext \
+                gfxterm gzio help linux loadenv lsefi normal part_gpt \
+                part_msdos read regexp search search_fs_file search_fs_uuid \
+                search_label terminal terminfo test tftp time
```
jforissier: You are adding more modules here, are they all needed?

hikey.mk
```diff
@@ -372,12 +372,35 @@ atf-fb-clean:
...
-	$(MAKE) -C $(LLOADER_PATH) BL1=$(ARM_TF_PATH)/build/hikey/$(ARM_TF_BUILD)/bl1.bin CROSS_COMPILE="$(CCACHE)$(AARCH32_CROSS_COMPILE)" PTABLE_LST=linux-$(CFG_FLASH_SIZE)g
+.PHONY: lloader-bin
+lloader-bin: arm-tf atf-fb
+	cd $(LLOADER_PATH) && \
+		ln -sf $(ARM_TF_PATH)/build/hikey/$(ARM_TF_BUILD)/bl1.bin && \
+		ln -sf $(ATF_FB_PATH)/build/hikey/$(ATF_FB_BUILD)/bl1.bin fastboot.bin && \
+		$(AARCH32_CROSS_COMPILE)gcc -c -o start.o start.S && \
+		$(AARCH32_CROSS_COMPILE)ld -Bstatic -Tl-loader.lds -Ttext 0xf9800800 start.o -o loader && \
+		$(AARCH32_CROSS_COMPILE)objcopy -O binary loader temp && \
+		python gen_loader_hikey.py -o l-loader.bin --img_loader=temp --img_bl1=bl1.bin --img_ns_bl1u=fastboot.bin
```
jforissier: That is seriously ugly and fragile, too.
Can't you fix the lloader Makefile instead and keep the previous code here?

hikey_debian.mk
```diff
@@ -359,12 +359,35 @@ system-img-cleaner:
 ################################################################################
 # l-loader
 ################################################################################
-lloader: arm-tf
-	$(MAKE) -C $(LLOADER_PATH) BL1=$(ARM_TF_PATH)/build/hikey/$(ARM_TF_BUILD)/bl1.bin CROSS_COMPILE="$(CCACHE)$(AARCH32_CROSS_COMPILE)" PTABLE_LST=linux-$(CFG_FLASH_SIZE)g
+.PHONY: lloader-bin
+lloader-bin: arm-tf atf-fb
```
jforissier: Same comment here of course.

hikey_debian.mk
```diff
@@ -205,6 +205,7 @@ LINUX_CLEAN_COMMON_FLAGS += ARCH=arm64
 
 .PHONY: linux-clean
 linux-clean: linux-clean-common
+	rm -f $(ROOT)/linux-*
```
jforissier: That is a bit dangerous. Can you be slightly more specific (less wildcarding)?

jforissier:
With my comments addressed:

hikey.mk
vchong: Iiuc, `build` is actually `edk2/BaseTools/Source/Python/build/build.py`, so there's no way to invoke $(MAKE) directly.

There used to be a Makefile but not anymore. The Makefile for Arm Platform (`fvp.mk`, `qemu_v8.mk`) still exists in `edk2/ArmPlatformPkg/Scripts/Makefile`, but it's just a wrapper for `build` too, and so both these .mk files also hard-code `-n 6` in their `edk2-call`.

If we don't specify the number of threads, i.e not using `-n` option, and if I'm reading `build.py` and related code properly, it defaults to a single threaded build, which means it breaks proper parallel build too? Which means safer but slower?

In any case, I've tried various combinations of 1 and _NPROCESSORS_ONLN for both the `edk2` target and `edk2-call`, and this is the only one that never fails to build, and fast too! Even `-j1` in `edk2:` and `-n 1` in `edk2-call`, which used to be stable previously, failed for some reason.

hikey.mk
vchong: Ok. I'll submit a PR to the l-loader repo and make the necessary changes here once it's merged.

hikey_debian.mk
vchong: These are the files that need to go:
```
linux-4.9.0-optee-rpb_4.9.0-optee-rpb-1_amd64.changes
linux-4.9.0-optee-rpb_4.9.0-optee-rpb-1.debian.tar.gz
linux-4.9.0-optee-rpb_4.9.0-optee-rpb-1.dsc
linux-4.9.0-optee-rpb_4.9.0-optee-rpb.orig.tar.gz
linux-firmware-image-4.9.0-optee-rpb_4.9.0-optee-rpb-1_amd64.deb
linux-headers-4.9.0-optee-rpb_4.9.0-optee-rpb-1_amd64.deb
linux-image-4.9.0-optee-rpb_4.9.0-optee-rpb-1_amd64.deb
linux-libc-dev_4.9.0-optee-rpb-1_amd64.deb
```

so is below ok?
```
rm -f $(ROOT)/linux-*.changes
rm -f $(ROOT)/linux-*.tar.gz
rm -f $(ROOT)/linux-*.dsc
rm -f $(ROOT)/linux-*.deb
```

vchong: Actually, what about `rm -f $(ROOT)/linux-*optee*.*`? Seems more efficient and safer too.

jforissier: Agreed.

hikey.mk
vchong: We've had this discussion before. : P
https://github.com/OP-TEE/build/pull/154#discussion_r121326578

jforissier: Not applicable, this is hikey.mk not hikey_debian.mk, we have a very simple, limited environment and I bet you a beer that not all modules are useful. Can you tell, for each of them, what it used for and/or what error happens if removed?

hikey.mk
jforissier: Sounds good!

hikey.mk
jforissier: OK, I'm ready to accept that as a valid answer because I'm really not eager to go and try to understand the edk2 build system :O

hikey.mk
vchong: Ok. I'll revert back to the original list of modules.

hikey.mk
vchong: @jforissier l-loader Makefile PR has been merged and instructions here updated. Please ack before I rebase, squash and tag.

vchong: Link to l-loader PR: https://github.com/96boards-hikey/l-loader/pull/5

vchong: Sorry, please hold. Some corrections required.

vchong: Ok. All system go again. Thanks!

jforissier:
How can the last commit work on HiKey? https://github.com/OP-TEE/manifest/blob/master/hikey.xml uses the master branch of https://github.com/96boards-hikey/l-loader and I don't see the new targets (hikey-clean etc.) in the Makefile :(

HiKey960 looks OK (https://github.com/OP-TEE/manifest/blob/master/hikey960.xml).

vchong:
This PR depends on https://github.com/OP-TEE/manifest/pull/74 which uses testing/hikey960_v1.2 branch of l-loader.

jforissier:
OK, then I'll trust you and merge this + https://github.com/OP-TEE/manifest/pull/74 simultaneously.
[rant]
Probably not the first time I say this, but: the fact that we use two Gits (build + manifest) is really a PITA. It makes it difficult to try pull requests because one has to manually chase the dependencies. The consequence of this is, although I like to make sure PRs don't break HiKey (my development platform of choice), I will likely not bother in the future because it's just too error-prone and time-consuming. Well testing is probably not the primary role of the maintainer anyway...
[/rant]

jforissier:

jforissier:
@vchong add Ack and squash fixup commits, thanks!

jbech-linaro:
> [rant]
> Probably not the first ....

So, what would you suggest instead? I'm open for suggestions. As long as we have something that is easy (?) to work with, fast to setup (--reference) and works for all our platforms I'm all ears.

jforissier:
@jbech-linaro as I said in the past, I think git submodules would suit our workflow much better than repo.

What I'm thinking about is: one single main project, submodules for all dependencies, one branch per platform, one tag per platform for each OP-TEE release (or not tag if the platform was not tested!).
Not the best time for me to try that, since it's release time ;-) but maybe later I can do it and let you try various use cases by yourself. A simple prototype with two platforms (QEMU and HiKey) should be enough.

jbech-linaro:
> but maybe later I can do it and let you try various use cases by yourself. A simple prototype with two platforms (QEMU and HiKey) should be enough.

@jforissier Sounds like a plan, let's try to look into that after the v2.6.0 release and then we can evaluate if this if better than our current solution.

vchong:
Rebased, squashed and tagged. Thanks!

PR #194, opened 2017-09-16
Benchmark support for HiKey6220 and TI SoCs(am57xx, dra7xx)

igoropaniuk:
To be able to run benchmarks on TI platforms "OP-TEE Benchmark" patch         
for linux optee driver should be applied manually (as they are using their own linux fork), `CONFIG_OPTEE_BENCHMARK=y` flag should be added to `ti_config_fragments/baseport.cfg`.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -24,6 +24,9 @@ LIBYAML_LIB_PATH		?= $(BENCHMARK_APP_PATH)/libyaml/out/lib
 CFG_TEE_CORE_LOG_LEVEL		?= 3
 
 # default disable latency benchmarks (over all OP-TEE layers)
+# To be able to run benchmarks on TI platforms "OP-TEE Benchmark" patch 
+# for linux optee driver should be applied manually, CONFIG_OPTEE_BENCHMARK=y
+# flag should be added to ti_config_fragments/baseport.cfg.
 CFG_TEE_BENCHMARK			?= n
 
 CCACHE ?= $(shell which ccache) # Don't remove this comment (space is needed)
--- a/hikey.mk
+++ b/hikey.mk
@@ -51,6 +51,11 @@ PATCHES_PATH			?=$(ROOT)/patches_hikey
 ################################################################################
 # Targets
 ################################################################################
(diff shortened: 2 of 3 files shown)


common.mk
```diff
@@ -24,6 +24,9 @@ LIBYAML_LIB_PATH		?= $(BENCHMARK_APP_PATH)/libyaml/out/lib
 CFG_TEE_CORE_LOG_LEVEL		?= 3
 
 # default disable latency benchmarks (over all OP-TEE layers)
+# To be able to run benchmarks on TI platforms "OP-TEE Benchmark" patch
```
jbech-linaro: Sentence is a bit hard to follow, do you mean something like:
`# To be able to run benchmarks on TI platforms, the "OP-TEE Benchmark" patch(es) in Linux kernel needs to be applied manually by setting CONFIG_OPTEE_BENCHMARK=y`

igoropaniuk: Actually I meant that two steps should be done:
1) Cherry-pick OP-TEE benchmark patches from our linaro-swg/linux rep, and apply it (as in TI manifests their own linux fork is used)
2) Manually add `CONFIG_OPTEE_BENCHMARK=y` to `ti_config_fragments/baseport.cfg.`

jbech-linaro: > Cherry-pick OP-TEE benchmark patches from our linaro-swg/linux rep, and apply it (as in TI manifests their own linux fork is used)

I think we instead should put the benchmark patches on the LSK branch that TI are using 
`branch: v4.9/topic/optee (git://git.linaro.org/kernel/linux-linaro-stable.git)`.

igoropaniuk: @etienne-lms 
So I guess I should take a break for changes that enable benchmarks on TI SoCs(remove this commit), so we can just add support for HiKey620. What do you think?

And when benchmark patches will be upstreamed to LSK branch, then we will return to TI benchmark enabling patches here in `built.git`

jbech-linaro: Igor, let's sync up on this on our next 1:1.

igoropaniuk: @jbech-linaro sure, will add it to 1:1 doc.

PR #196, opened 2017-09-20
qemu: fix "make check" to use OP-TEE image v2

jforissier:
Commit 37ab51d ("qemu: bios load OP-TEE image v2") has updated the run
and run-only targets, but not make check. Fix this.


jbech-linaro:

PR #201, opened 2017-10-13
qemu: use 4 virtual CPUs (-smp 4) for "make check"

jforissier:
Now that OP-TEE supports several cores in QEMU, use 4 of them when
running "make check". This is hopefully more likely to catch any
concurrency bug.

### DIFF
--- a/qemu-check.exp
+++ b/qemu-check.exp
@@ -42,7 +42,7 @@ info "Starting QEMU..."
 open "serial1.log" "w+"
 spawn -open [open "|tail -f serial1.log"]
 set teecore $spawn_id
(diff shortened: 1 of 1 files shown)


jbech-linaro:
Seems good to me, the only thing is that I recall Jens mention something about it was better having 2 cores instead of 4, but I cannot remember what it was about. Anyhow ...

jenswi-linaro:
The QEMU-virt port is configured to support up to 4 CPUs so that part is OK. Using 4 cores instead of 2 cores will mean even more overhead, but that's perhaps not a problem.

jforissier:
About overhead: indeed, with `-smp 4` the tests tend to run a bit more slowly than without `-smp`, but not by much and it's OK IMO.

PR #203, opened 2017-10-18
qemu: build a statically linked version of QEMU

fboudra:
Reduce the dependencies on the host environment.
It allows to reuse easily the binary built in other enviroment like
ramdisk and other CI environments.

Signed-off-by: Fathi Boudra <fathi.boudra@linaro.org>

### DIFF
--- a/qemu.mk
+++ b/qemu.mk
@@ -59,7 +59,7 @@ bios-qemu-clean:
 	$(call bios-qemu-common) clean
 
 qemu:
-	cd $(QEMU_PATH); ./configure --target-list=arm-softmmu\
+	cd $(QEMU_PATH); ./configure --target-list=arm-softmmu --static\
 			$(QEMU_CONFIGURE_PARAMS_COMMON)
 	$(MAKE) -C $(QEMU_PATH)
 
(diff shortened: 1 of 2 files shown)


jbech-linaro:
Travis seems to be happy with this, so I guess it is a local "Arch Linux" thing, but I get

```make
make[1]: Leaving directory '/mnt/sshd/devel/optee_projects/qemu/bios_qemu_tz_arm'
cd /mnt/sshd/devel/optee_projects/qemu/build/../qemu; ./configure --target-list=arm-softmmu --static\
                --cc="/usr/bin/ccache gcc" --extra-cflags="-Wno-error"

ERROR: sizeof(size_t) doesn't match GLIB_SIZEOF_SIZE_T.
       You probably need to set PKG_CONFIG_LIBDIR
       to point to the right pkg-config files for your
       build target

make: *** [Makefile:62: qemu] Error 1
```
When testing this patch. Since I guess this is a local configuration error, I'm still happy with the patch.

fboudra:
ci.linaro.org is happy with it too. I've applied the patch in the build temporary and used it in lava so it's functional.

jforissier:

PR #204, opened 2017-10-19
mtk8173: Disable unused features in Linux kernel

jbech-linaro:
On the mtk8173 device there is partition which is ~25MB in size and that
area contains both Linux kernel and the root fs. When we stepped up from
Linux kernel 4.9 to 4.12 we got too big to fit in that area and
therefore we have tweaked the config and removed a couple of things not
being used in our setup.


jforissier:

jbech-linaro:

PR #205, opened 2017-10-20
fvp: update path to simulator executable

jenswi-linaro:
The path to the executable in FVP Foundation model has changed, update
the path accordingly.


jforissier:

jenswi-linaro:
Tag applied

jforissier:
About the Travis build failure: I suspect that QEMU is not the proper revision (https://github.com/OP-TEE/manifest/blob/master/qemu_v8_stable.xml#L18 vs. https://github.com/OP-TEE/manifest/blob/master/qemu_v8.xml#L36), but I did not checked further.

PR #208, opened 2017-10-27
juno: upload all the device tree blobs

fboudra:
It's assumed we have a Juno R0 board. It isn't always the case. If someone
has another revision, he'll prefer to use the right device tree for his
board.

Signed-off-by: Fathi Boudra <fathi.boudra@linaro.org>

### DIFF
--- a/juno.mk
+++ b/juno.mk
@@ -160,4 +160,6 @@ flash:
 	$(FTP-UPLOAD) $(ARM_TF_PATH)/build/juno/release/fip.bin
 	$(FTP-UPLOAD) $(ROOT)/linux/arch/arm64/boot/Image
 	$(FTP-UPLOAD) $(ROOT)/linux/arch/arm64/boot/dts/arm/juno.dtb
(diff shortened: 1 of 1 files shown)


jenswi-linaro:
Looks good, but it's not clear to me how the correct dtb is selected. Some magic in u-boot?

fboudra:
@jenswi-linaro Depending on your boot method, you can define which dtb to use in U-Boot. You can also generate your own recovery image and flash in NOR flash. Though the NOR flash approach is black magic to me, because I don't need to specify which dtb to take, it just takes the right one.

jbech-linaro:

jbech-linaro:
Strange that the v8-build on Travis gives the Arm-TF GIC errors again (`g0_interrupt_num`), Jerome removed that flag in https://github.com/OP-TEE/build/pull/206, but still I can see in the build log that it is set.

jenswi-linaro:

jforissier:
Merged, thanks.

PR #211, opened 2017-11-15
FVP: Use official EDK2 and use boot image instead of semihosting

jbech-linaro:
This will move away from the old EDK2 landing team release and instead use latest and greatest from the official EDK2 tree(s). With this we are using a proper boot-image instead of relying on semihosting (which isn't supported in EDK2 anymore, which has only been on the Linaro landing-team branches).

To try this out one needs the commits in https://github.com/OP-TEE/manifest/pull/84. I.e., both this PR and the one in manifest.git needs to be merged at the same time.

If one want to try this on a local forest (just to show that this **can** be done with repo also, although I personally prefer [checking out the PR's from GitHub directly](https://help.github.com/articles/checking-out-pull-requests-locally/) instead, since that is just a oneliner), then one can:
```bash
$ repo init -u https://github.com/jbech-linaro/manifest.git -b fvp_grub -m fvp.xml
```

Create a folder `.repo/local_manifests/` and in there a file `foo.xml` containing:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<manifest>
        <remote name="jbech" fetch="https://github.com/jbech-linaro" />

        <remove-project path="build" name="build.git" />
        <project remote="jbech" path="build" name="build.git" revision="update_edk2">
                <linkfile src="fvp.mk" dest="build/Makefile" />
        </project>
</manifest>
```
And then sync again:
```bash
$ repo sync
```

Building and running are the same as before.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -207,27 +207,21 @@ linux-cleaner-common: linux-defconfig-clean
 ################################################################################
 # EDK2 / Tianocore
 ################################################################################
-# Make sure edksetup.sh only will be called once and that we don't rebuild
-# BaseTools again and again.
-$(EDK2_PATH)/Conf/target.txt:
-	set -e && cd $(EDK2_PATH) && source edksetup.sh && \
-	$(MAKE) -j1 -C $(EDK2_PATH)/BaseTools
-
 .PHONY: edk2-common
-edk2-common: $(EDK2_PATH)/Conf/target.txt
-	set -e && cd $(EDK2_PATH) && source edksetup.sh && \
-	$(call edk2-call)
+edk2-common:
+	export WORKSPACE=$(ROOT) && \
+	export PACKAGES_PATH=$(EDK2_PATH):$(ROOT)/edk2-platforms && \
+	source $(EDK2_PATH)/edksetup.sh && \
+	$(MAKE) -j1 -C $(EDK2_PATH)/BaseTools && \
+	$(call edk2-call) all
 
 .PHONY: edk2-clean-common
 edk2-clean-common:
-	set -e && cd $(EDK2_PATH) && source edksetup.sh && \
-	$(call edk2-call) clean && \
-	$(MAKE) -j1 -C $(EDK2_PATH)/BaseTools clean
-	rm -rf $(EDK2_PATH)/Build
-	rm -rf $(EDK2_PATH)/Conf/.cache
-	rm -f $(EDK2_PATH)/Conf/build_rule.txt
-	rm -f $(EDK2_PATH)/Conf/target.txt
-	rm -f $(EDK2_PATH)/Conf/tools_def.txt
+	export WORKSPACE=$(ROOT) && \
+	export PACKAGES_PATH=$(EDK2_PATH):$(ROOT)/edk2-platforms && \
+	source $(EDK2_PATH)/edksetup.sh && \
+	$(MAKE) -j1 -C $(EDK2_PATH)/BaseTools clean && \
+	$(call edk2-call) cleanall
 
 ################################################################################
 # QEMU / QEMUv8
--- a/fvp.mk
+++ b/fvp.mk
@@ -18,7 +18,8 @@ include common.mk
 ################################################################################
 ARM_TF_PATH		?= $(ROOT)/arm-trusted-firmware
 EDK2_PATH		?= $(ROOT)/edk2
-EDK2_BIN		?= $(EDK2_PATH)/Build/ArmVExpress-FVP-AArch64/RELEASE_GCC49/FV/FVP_AARCH64_EFI.fd
+EDK2_BIN		?= $(ROOT)/Build/ArmVExpress-FVP-AArch64/RELEASE_GCC49/FV/FVP_AARCH64_EFI.fd
+EDK2_PLATFORMS_PATH	?= $(ROOT)/edk2-platforms
 FOUNDATION_PATH		?= $(ROOT)/Foundation_Platformpkg
 ifeq ($(wildcard $(FOUNDATION_PATH)),)
 $(error $(FOUNDATION_PATH) does not exist)
@@ -27,13 +28,14 @@ GRUB_PATH		?= $(ROOT)/grub
 GRUB_CONFIG_PATH	?= $(BUILD_PATH)/fvp/grub
 OUT_PATH		?= $(ROOT)/out
 GRUB_BIN		?= $(OUT_PATH)/bootaa64.efi
+BOOT_IMG		?= $(OUT_PATH)/boot-fat.uefi.img
 
 ################################################################################
 # Targets
 ################################################################################
-all: arm-tf edk2 grub linux optee-os optee-client xtest optee-examples
-clean: arm-tf-clean busybox-clean edk2-clean grub-clean optee-os-clean \
-	optee-client-clean optee-examples-clean
+all: arm-tf boot-img edk2 grub linux optee-os optee-client xtest optee-examples
+clean: arm-tf-clean boot-img-clean busybox-clean edk2-clean grub-clean \
+	optee-os-clean optee-client-clean optee-examples-clean
 
 
 include toolchain.mk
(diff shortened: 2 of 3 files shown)


common.mk
```diff
@@ -207,27 +207,21 @@ linux-cleaner-common: linux-defconfig-clean
...
+	export PACKAGES_PATH=$(EDK2_PATH):$(ROOT)/edk2-platforms && \
+	source $(EDK2_PATH)/edksetup.sh && \
+	$(MAKE) -j1 -C $(EDK2_PATH)/BaseTools && \
+	$(call edk2-call) all
 
 .PHONY: edk2-clean-common
 edk2-clean-common:
+	export WORKSPACE=$(ROOT) && \
+	export PACKAGES_PATH=$(EDK2_PATH):$(ROOT)/edk2-platforms && \
 	set -e && cd $(EDK2_PATH) && source edksetup.sh && \
```
etienne-lms: minor: `set -e && cd $(EDK2_PATH)` needed?  sync with `source $(EDK2_PATH)/edksetup.sh` above?

fvp.mk
```diff
@@ -146,19 +155,72 @@ filelist-tee: filelist-tee-common
...
+
+$(GRUB_PATH)/configure: $(GRUB_PATH)/configure.ac
+	cd $(GRUB_PATH) && ./autogen.sh
+
+$(GRUB_PATH)/Makefile: $(GRUB_PATH)/configure
+	cd $(GRUB_PATH) && ./configure --target=aarch64 --enable-boot-time $(grub-flags)
+
+.PHONY: grub
+grub: $(GRUB_PATH)/Makefile | $(OUT_PATH)
+	$(MAKE) -C $(GRUB_PATH); \
```
etienne-lms: `&&` instead of `;`?

fvp.mk
```diff
@@ -146,19 +155,72 @@ filelist-tee: filelist-tee-common
...
+		--output=$(GRUB_BIN) \
+		--config=$(GRUB_CONFIG_PATH)/grub.cfg \
+		--format=arm64-efi \
+		--directory=grub-core \
+		--prefix=/boot/grub \
+		$(GRUB_MODULES)
+
+.PHONY: grub-clean
+grub-clean:
+	if [ -e $(GRUB_PATH)/Makefile ]; then $(MAKE) -C $(GRUB_PATH) clean; fi
```
etienne-lms: maybe some `@` here and there.

common.mk
jbech-linaro: You're meaning the `set -e`? Yeah that is unneccesary with all the `&&` being used. Is that what you had in mind?

etienne-lms: it is just a minor comment to align `edk2-clean-common` sequence on `edk2-common` when can be similar.

fvp.mk
jbech-linaro: Agree, will fix.

fvp.mk
jbech-linaro: I will randomly put some `@` there :smile: ... if grub wasn't using autotools, then we could just simply run `make -C $(GRUB_PATH) clean` :wink:

etienne-lms: I'm blessed! :cry:

common.mk
vchong: Also, don't see `WORKSPACE` and `PACKAGES_PATH` used anywhere else. Are they required for the build to build?

fvp.mk
```diff
@@ -18,23 +18,35 @@ include common.mk
...
+OUT_PATH		?= $(ROOT)/out
+GRUB_BIN		?= $(OUT_PATH)/bootaa64.efi
+BOOT_IMG		?= $(OUT_PATH)/boot-fat.uefi.img
 
 ################################################################################
 # Targets
 ################################################################################
-all: arm-tf edk2 linux optee-os optee-client xtest optee-examples
-clean: arm-tf-clean busybox-clean edk2-clean optee-os-clean \
+all: arm-tf boot-img edk2 grub linux optee-os optee-client xtest optee-examples
```
vchong: `boot-img` depends on `linux grub update_rootfs` so can remove `linux` and `grub` from `all:`.

fvp.mk
```diff
@@ -18,23 +18,35 @@ include common.mk
...
+clean: arm-tf-clean busybox-clean edk2-clean grub-clean optee-os-clean \
```
vchong: Add `boot-img-clean` here?

fvp.mk
```diff
@@ -146,19 +155,72 @@ filelist-tee: filelist-tee-common
...
+		search_label terminal terminfo test tftp time
+
+$(GRUB_PATH)/configure: $(GRUB_PATH)/configure.ac
+	cd $(GRUB_PATH) && ./autogen.sh
+
+$(GRUB_PATH)/Makefile: $(GRUB_PATH)/configure
+	cd $(GRUB_PATH) && ./configure --target=aarch64 --enable-boot-time $(grub-flags)
+
+.PHONY: grub
+grub: $(GRUB_PATH)/Makefile | $(OUT_PATH)
```
vchong: What does the `|` do?

fvp.mk
```diff
@@ -146,19 +155,72 @@ filelist-tee: filelist-tee-common
...
+	if [ -e $(GRUB_PATH)/Makefile ]; then $(MAKE) -C $(GRUB_PATH) clean; fi
+	rm -f $(GRUB_BIN)
+	rm -f $(GRUB_PATH)/configure
+
+
+################################################################################
+# Boot Image
+################################################################################
+.PHONY: boot-img
+boot-img: arm-tf linux busybox grub update_rootfs
```
vchong: `arm-tf` and `busybox` not needed here.

fvp.mk
```diff
@@ -146,19 +155,72 @@ filelist-tee: filelist-tee-common
...
+boot-img-clean:
+	rm -f $(BOOT_IMG)
+
+################################################################################
 # Run targets
 ################################################################################
 # This target enforces updating root fs etc
 run: all
-	$(MAKE) update_rootfs
+	$(MAKE) boot-img
```
vchong: `all` already builds `boot-img` so this is not needed. Or if staying with the previous structure, keep `boot-img` here and remove it from `all:` above.

fvp.mk
jbech-linaro: That is probably true and actually yesterday I remove every single prereq/target and started from scratch, because I think we have dependencies that are not working as they should. But didn't submit that patch yet, I wanted to wait with that and do that as a separate "clean pre-req / target" thing. We have some platforms where the -jX flag doesn't work on the first run (RPI3).

fvp.mk
jbech-linaro: Yes! Thanks!

fvp.mk
jbech-linaro: That will create the out path only when necessary (i.e., avoid updating it). It's the so called [order-only-prerequisites](https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html). If you aren't using that, then it will always run the `$(OUT_PATH)` target which will update the timestamp on that folder and therefore could/will trigger unnecessary (re-)builds. [Here](https://www.cmcrossroads.com/article/making-directories-gnu-make) is a pretty good example and explanation of this.

fvp.mk
jbech-linaro: True I guess, can't recall why I added it, will clean it up, thanks!

fvp.mk
jbech-linaro: Yeps, I think you're right, I'll try to remove it.

common.mk
jbech-linaro: I guess so, the `build` script in UEFI seems to use those.

jbech-linaro:
Comments addressed! If you're happy with it, I'd like to have an RB tag.

fvp.mk
vchong: Very good to know thanks!

vchong:
Sorry for the delay.

jbech-linaro:
Squashed, rebased and tag(s) applied (should be) ready for merge if I can get a RB for https://github.com/OP-TEE/manifest/pull/84 too! Thanks!

jbech-linaro:
The dependency: https://github.com/OP-TEE/manifest/pull/84 also is also ready for merge now.

jbech-linaro:
Thanks @jforissier !

PR #212, opened 2017-11-21
[RFC] optee_client builds through autotools

etienne-lms:
Build optee_client from its subdir out/.
Install optee_client in subdir out/export/.

Depends on https://github.com/OP-TEE/optee_client/pull/110.


etienne-lms:
Closed, cmake is preferred.

PR #214, opened 2017-11-25
qemu v8: Use default path from edk2-platforms

jbech-linaro:
With the introduction of edk2-platforms ([manifest/PR85](https://github.com/OP-TEE/manifest/pull/85)) we need to update the default edk2-platforms path to the built firmware.

### DIFF
--- a/qemu_v8.mk
+++ b/qemu_v8.mk
@@ -97,6 +97,10 @@ busybox-cleaner: busybox-cleaner-common
 ################################################################################
 # EDK2 / Tianocore
 ################################################################################
+define edk2-env
+	export WORKSPACE=$(EDK2_PATH)
+endef
+
 define edk2-call
 	GCC49_AARCH64_PREFIX=$(LEGACY_AARCH64_CROSS_COMPILE) \
 		$(MAKE) -j1 -C $(EDK2_PATH) \
--- a/common.mk
+++ b/common.mk
@@ -209,15 +209,15 @@ linux-cleaner-common: linux-defconfig-clean
 ################################################################################
 .PHONY: edk2-common
 edk2-common:
-	export WORKSPACE=$(ROOT) && \
-	export PACKAGES_PATH=$(EDK2_PATH):$(ROOT)/edk2-platforms && \
+	$(call edk2-env) && \
+	export PACKAGES_PATH=$(EDK2_PATH):$(EDK2_PLATFORMS_PATH) && \
 	source $(EDK2_PATH)/edksetup.sh && \
 	$(MAKE) -j1 -C $(EDK2_PATH)/BaseTools && \
 	$(call edk2-call) all
 
 .PHONY: edk2-clean-common
 edk2-clean-common:
-	export WORKSPACE=$(ROOT) && \
+	$(call edk2-env) && \
 	export PACKAGES_PATH=$(EDK2_PATH):$(ROOT)/edk2-platforms && \
 	source $(EDK2_PATH)/edksetup.sh && \
 	$(MAKE) -j1 -C $(EDK2_PATH)/BaseTools clean && \
--- a/fvp.mk
+++ b/fvp.mk
@@ -18,8 +18,8 @@ include common.mk
 ################################################################################
 ARM_TF_PATH		?= $(ROOT)/arm-trusted-firmware
 EDK2_PATH		?= $(ROOT)/edk2
-EDK2_BIN		?= $(ROOT)/Build/ArmVExpress-FVP-AArch64/RELEASE_GCC49/FV/FVP_AARCH64_EFI.fd
 EDK2_PLATFORMS_PATH	?= $(ROOT)/edk2-platforms
+EDK2_BIN		?= $(EDK2_PLATFORMS_PATH)/Build/ArmVExpress-FVP-AArch64/RELEASE_GCC49/FV/FVP_AARCH64_EFI.fd
 FOUNDATION_PATH		?= $(ROOT)/Foundation_Platformpkg
 ifeq ($(wildcard $(FOUNDATION_PATH)),)
 $(error $(FOUNDATION_PATH) does not exist)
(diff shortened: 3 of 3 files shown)


qemu_v8.mk
```diff
@@ -16,7 +16,7 @@ include common.mk
 ARM_TF_PATH		?= $(ROOT)/arm-trusted-firmware
 
 EDK2_PATH		?= $(ROOT)/edk2
-EDK2_BIN		?= $(EDK2_PATH)/Build/ArmVirtQemuKernel-AARCH64/DEBUG_GCC49/FV/QEMU_EFI.fd
+EDK2_BIN		?= $(ROOT)/Build/ArmVirtQemuKernel-AARCH64/DEBUG_GCC49/FV/QEMU_EFI.fd
```
jforissier: So, we have `$(ROOT)/build` which contains source files (build.git), and `$(ROOT)/Build` which is the output of `edk2` and `edk2-platforms`? That seems a bit confusing to me. Is there no way to keep EDK2 files under `edk2/build`?

jbech-linaro: That is confusing and a bit annoying, but there seems to be no `O=...` equivalent when building EDK2, not what I've found at least.

etienne-lms: port it to autotools!

jforissier: Keep `EDK2_BIN` unchanged and set `WORKSPACE=$(EDK2_PATH)` in `common.mk` 
(https://github.com/tianocore/tianocore.github.io/wiki/Multiple_Workspace#environment-variables says "WORKSPACE: location of the Build directory").

jbech-linaro: Ah, that seems to work fine, thanks for looking that up @jforissier . Btw, this requires a change in `fvp.mk` too. I've changed that in the fix I just submitted. Haven't tested it on FVP yet, but I'm going to do that now.

jforissier: Sounds good

jforissier:

qemu_v8.mk
jbech-linaro: Unfortunately this doesn't seem to work when building FVP.  With full path it complains that it cannot find the dsc-file and if I remove the full path, like this:
```diff
-               -t "GCC49" -p $(EDK2_PLATFORMS_PATH)/Platform/ARM/VExpressPkg/ArmVExpress-FVP-AArch64.dsc -b RELEASE
+               -t "GCC49" -p Platform/ARM/VExpressPkg/ArmVExpress-FVP-AArch64.dsc -b RELEASE
```
Then it seems to find the dsc-file and compiles some stuff, but ends with:
```bash
GenFds.py...
 : error F003: Output file for RAW section could not be found for Platform/ARM/VExpressPkg/AcpiTables/AcpiTables.inf

#########


build.py...
 : error 7000: Failed to execute command
        GenFds -f /mnt/sshd/devel/optee_projects/fvp/edk2-platforms/Platform/ARM/VExpressPkg/ArmVExpress-FVP-AArch64.fdf --conf=/mnt/sshd/devel/optee_projects/fvp/edk2/Conf -o /mnt/sshd/devel/optee_projects/fvp/edk2/Build/ArmVExpress-FVP-AArch64/RELEASE_GCC49 -t GCC49 -b RELEASE -p /mnt/sshd/devel/optee_projects/fvp/edk2-platforms/Platform/ARM/VExpressPkg/ArmVExpress-FVP-AArch64.dsc -a AARCH64 -D "EFI_SOURCE=/mnt/sshd/devel/optee_projects/fvp/edk2/EdkCompatibilityPkg" -D "EDK_SOURCE=/mnt/sshd/devel/optee_projects/fvp/edk2/EdkCompatibilityPkg" -D "TOOL_CHAIN_TAG=GCC49" -D "TOOLCHAIN=GCC49" -D "TARGET=RELEASE" -D "FAMILY=GCC" -D "WORKSPACE=/mnt/sshd/devel/optee_projects/fvp/edk2" -D "EDK_TOOLS_PATH=/mnt/sshd/devel/optee_projects/fvp/edk2/BaseTools" -D "ARCH=AARCH64" -D "ECP_SOURCE=/mnt/sshd/devel/optee_projects/fvp/edk2/EdkCompatibilityPkg" [/mnt/sshd/devel/optee_projects/fvp/edk2]

- Failed -
Build end time: 10:54:03, Nov.27 2017
Build total time: 00:00:28
```

I suspect this is working on QEMU because we (incorrectly) are doing a shortcut by calling `make` directly in QEMU, but (correctly) using the `build` script when building for FVP. Also, the `*.dsc` file for QEMU lives in `edk2` while the the `dsc` for FVP lives in edk2-platforms. 

I've been testing just a clean edk2/edk2-platforms according to the instructions here:
https://github.com/tianocore/edk2-platforms

Running the https://github.com/tianocore/edk2-platforms#using-uefi-tools-helper-scripts with `$WORKSPACE` set to `edk2` will also make that script fail. So ... I don't really know what to do here.

jbech-linaro: Exporting `WORKSPACE` to `edk2-platforms` instead of `edk2` does seems to work for FVP.

jbech-linaro:
@jforissier I finally came up with something that seems to be working on both QEMU v8 and FVP without adding a toplevel `Build` folder, let me know if this looks good to you and then I'll squash and rebase.

jforissier:
@jbech-linaro looks good to me, although it smells a bit fishy :( My R-b still applies. Thanks!

jbech-linaro:
> although it smells a bit fishy :(

Totally agree, building UEFI is tricky and it doesn't help with the big python scripts doing lots of checks meanwhile compiling, it's really hard to understand what is happening and what is going wrong when things (_are_ and) aren't working. In the end this makes our way of building it pretty fragile. It feels like it is a bit pure luck that it works as good as it does. But I think we are one step closer to how it is supposed to be built now at least. I'll apply the R-B and squash.

jbech-linaro:
I've double checked on both QEMU v8 and FVP after squashing and things are indeed still working :)

PR #220, opened 2017-12-15
docs: update debug.md to match the new trace output

jbech-linaro:
This is the updates needed here due to the trace changes in https://github.com/OP-TEE/optee_os/pull/2011.

### DIFF
--- a/docs/debug.md
+++ b/docs/debug.md
@@ -415,106 +415,110 @@ $ echo 0 > options/graph-time
 
 When OP-TEE encounters a serious error condition, it prints diagnostic
 information to the secure console. The message contains a call stack if
-CFG_UNWIND=y (enabled by default).
+`CFG_UNWIND=y` (enabled by default).
 The following errors will trigger a dump:
 - Data or prefetch abort exception in the TEE core (kernel mode) or in a TA
-(user mode),
-- When a user-mode Trusted Application panics, either by calling TEE_Panic()
-directly or due to some error detected by the TEE Core Internal API,
+  (user mode),
+- When a user-mode Trusted Application panics, either by calling `TEE_Panic()`
+  directly or due to some error detected by the TEE Core Internal API,
 - When the TEE core detects a fatal error and decides to hang the system
-because there is no way to proceed safely (core panic).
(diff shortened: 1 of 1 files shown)


docs/debug.md
```diff
@@ -415,107 +415,111 @@ $ echo 0 > options/graph-time
...
-The above dump was triggered by the TA trying to write to a global const
-variable. Such variables are stored in the `.rodata` section of the ELF binary,
-and loaded into a read-only section of memory.
-
-OP-TEE provides a helper script called `symbolize.py` to facilitate the
-analysis of such issues. It is located in the OP-TEE OS source tree, and is
-also copied to the TA development kit. Whenever you are confronted with an error
-message reporting a serious error and showing a "Call stack:" line, you may use
-the symbolize script.
+  read or write permission fault).
```
etienne-lms: Alignment issues also?

docs/debug.md
```diff
@@ -415,107 +415,111 @@ $ echo 0 > options/graph-time
...
+D/TC:0 tee_ta_close_session:402 tee_ta_close_session(0xe07be98)
+D/TC:0 tee_ta_close_session:421 Destroy session
+D/TC:0 tee_ta_close_session:447 Destroy TA ctx
+```
+
+The above dump was triggered by the TA when entering an irrecoverable error
+ending up in a `TEE_Panic(0)` call.
+
+OP-TEE provides a helper script called `symbolize.py` to facilitate the analysis
+of such issues. It is located in the OP-TEE OS source tree, and is also copied
```
etienne-lms: remove coma.

s/copied to/installed with/ ?

docs/debug.md
jbech-linaro: Yes, your are probably right.

docs/debug.md
jbech-linaro: It's Jerome's wording and I think copied to is fine here, it's not that we are running something like "make install" (so I'll keep the current wording).

etienne-lms:
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>`

PR #228, opened 2018-02-09
Qemu/armv7: use Arm-TF as secure bootloader

etienne-lms:
These changes allows to build arm-tf for qemu_virt (cortex-a15), build bios_qemu_tz_arm to act as a non secure bootloader and boot the qemu from arm-tf BL1/BL2.

Main makefile currently rely on directive `USE_ATF=y/n` to enable boot from arm-tf.
When booting qemu from arm-tf, one should set `CFG_QEMU_ATF_MAILBOX=y` for Qemu to wakeup secondary cores (case `QEMU_SMP>1` or `CFG_TEE_CORE_NB_CORE>1`).

If the arm-tf boot becomes to standard qemu boot scheme, `USE_ATF=y` will be implicit, and `CFG_QEMU_ATF_MAILBOX=y` will be default enable from the op-tee platform. 

Requires:
  https://github.com/linaro-swg/bios_qemu_tz_arm/pull/15
  https://github.com/OP-TEE/optee_os/pull/2138

### DIFF
--- a/qemu_v8.mk
+++ b/qemu_v8.mk
@@ -28,14 +28,9 @@ DEBUG = 1
 ################################################################################
 # Targets
 ################################################################################
-ifeq ($(CFG_TEE_BENCHMARK),y)
-all: benchmark-app
-clean: benchmark-app-clean
-endif
-all: arm-tf edk2 qemu soc-term linux update_rootfs optee-examples
-clean: arm-tf-clean busybox-clean edk2-clean linux-clean \
-	optee-os-clean optee-client-clean qemu-clean \
-	soc-term-clean check-clean strace-clean optee-examples-clean
+all: arm-tf qemu soc-term linux buildroot
+clean: arm-tf-clean edk2-clean linux-clean optee-os-clean qemu-clean \
+	soc-term-clean check-clean buildroot-clean
 
 include toolchain.mk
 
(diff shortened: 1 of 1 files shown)


qemu.mk
```diff
@@ -157,12 +217,38 @@ update_rootfs: update_rootfs-common
...
 .PHONY: run
 # This target enforces updating root fs etc
 run: all
 	$(MAKE) run-only
 
 .PHONY: run-only
+
+ifeq ($(USE_ATF),y)
+run-only: get-binaries
+get-binaries:
```
jenswi-linaro: Should have `.PHONY` attribute for `get-binaries`

qemu.mk
```diff
@@ -157,12 +217,38 @@ update_rootfs: update_rootfs-common
...
+	ln -sf $(ARM_TF_OUT)/bl1.bin $(BINARIES_PATH)/bl1.bin
+	ln -sf $(ARM_TF_OUT)/bl2.bin $(BINARIES_PATH)/bl2.bin
+	ln -sf $(ROOT)/out/bios-qemu/bios.bin $(BINARIES_PATH)/bl33.bin
+	ln -sf $(OPTEE_OS_HEADER_V2_BIN) $(BINARIES_PATH)/bl32.bin
+	ln -sf $(OPTEE_OS_PAGER_V2_BIN) $(BINARIES_PATH)/bl32_extra1.bin
+	ln -sf $(OPTEE_OS_PAGEABLE_V2_BIN) $(BINARIES_PATH)/bl32_extra2.bin
+
+BIOS_PATH := $(BINARIES_PATH)/bl1.bin
+else
+BIOS_PATH := $(ROOT)/out/bios-qemu/bios.bin
```
jenswi-linaro: Do we need to keep the old bios hack?

etienne-lms: Hmm, no. Actually, this p-r is almost a rfc. This is why I kept both. But in the end, we should use only 1 scheme. I'll update the p-r, discarding the old boot scheme support.

etienne-lms:
This needs an update.... on going.

etienne-lms:
Dependencies updated:
  https://github.com/linaro-swg/bios_qemu_tz_arm/pull/15
  https://github.com/OP-TEE/optee_os/pull/2138

Well... there is still an added dependency: one shall clone the arm-trusted-firmware to get it built!

PR #229, opened 2018-02-18
qemu_v8: fix missing edk2 'cleanall' target

goatshriek:
The edk2-clean-common target in common.mk uses a cleanall target, which does
not exist in the makefile used by edk2-call in qemu_v8.mk. Adding a specific
target for edk2-clean in qemu_v8.mk that uses 'clean' instead of 'cleanall'
allows the clean target to complete successfully.

Signed-off-by: Joel Anderson <jeander@vt.edu>

### DIFF
--- a/qemu_v8.mk
+++ b/qemu_v8.mk
@@ -112,7 +112,12 @@ endef
 
 edk2: edk2-common
 
-edk2-clean: edk2-clean-common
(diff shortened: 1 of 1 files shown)


goatshriek:
Okay, thanks! I'll look to your pull request to solve the issue.

PR #237, opened 2018-03-08
hikey, hikey960: make bl2.bin available to l-loader

jforissier:
The latest version of l-loader apparently needs bl2.bin in addition to
bl1.bin. Fixes the following error:

 make[2]: Entering directory '/home/optee/devel/hikey/l-loader'
 /usr/bin/ccache /home/optee/devel/hikey/build/../toolchains/aarch32/bin/arm-linux-gnueabihf-gcc  -c -o start.o start.S
 make[2]: *** No rule to make target 'bl2.bin', needed by 'l-loader.bin'.  Stop.

### DIFF
--- a/hikey.mk
+++ b/hikey.mk
@@ -367,6 +367,7 @@ atf-fb-clean:
 lloader: arm-tf atf-fb
 	cd $(LLOADER_PATH) && \
 		ln -sf $(ARM_TF_PATH)/build/hikey/$(ARM_TF_BUILD)/bl1.bin && \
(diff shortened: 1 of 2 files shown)


jenswi-linaro:

jforissier:
Well, it looks like things are not as simple. With this change, the build is OK but "make recovery" hangs.

Since I don't see any clear documentation in l-loader on how to update the build easily to get the same behavior as before, I'll just submit a PR to our manifest.git to pin the l-loader version to an older, non-broken one. I guess we don't care much whether we're using the latest and greatest l-loader.

PR #239, opened 2018-03-14
Support legacy compiler

jenswi-linaro:
Adds support for legacy compiler
gcc-linaro-4.9-2015.02-3-x86_64_arm-linux-gnueabihf

To compile using this compiler instead on ARMv7 systems supply
COMPILE_LEGACY=y to make.


toolchain.mk
```diff
@@ -35,7 +41,7 @@ define dltc
 endef
 
 .PHONY: toolchains
-toolchains: aarch32 aarch64 aarch64-legacy
+toolchains: aarch32 aarch64 aarch64-legacy aarch32-legacy
```
jforissier: Quick note: IMO it would be better to download only the necessary toolchain(s), either `aarch32` or `aarch32-legacy` depending on `COMPILE_LEGACY`. But this should be left for a separate PR, especially since we're already downloading both 32- and 64-bit toolchains when QEMU (v7) although the 64-bit compiler will never get used.

jenswi-linaro: Agree with that.

PR #240, opened 2018-03-14
qemu-check.exp: fix expected prompts

jforissier:
Since commit eed314d11475 ("qemu: build using buildroot"), the login
and the shell prompts of the root FS have changed. Adjust our "make
check" accordingly.


jenswi-linaro:

PR #250, opened 2018-04-26
hikey960: build using buildroot

jenswi-linaro:

Depends on https://github.com/OP-TEE/manifest/pull/108

This has only been test compiled, not run on hardware.
@jforissier, would you mind giving it a try?

### DIFF
--- a/hikey960.mk
+++ b/hikey960.mk
@@ -10,6 +10,8 @@ COMPILE_S_KERNEL  ?= 64
 
 # Normal and secure world console UART: 6 (v2 or newer board) or 5 (v1 board)
 CFG_CONSOLE_UART ?= 6
+# Needed by buildroot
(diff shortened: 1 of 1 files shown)


jforissier:
> @jforissier, would you mind giving it a try?

Sure. It could take some time however, because I need to clone the whole setup and I have a slow network access here.

jforissier:
First of all, https://github.com/OP-TEE/build/pull/251 and https://github.com/OP-TEE/manifest/pull/109 are needed or `make flash` fails.
With that, the board boots up to kernel initialization but then no login prompt is shown. The console keeps saying "can't open /dev/ttyAMA0: No such file or directory".

jenswi-linaro:
Please test again with the added patch

jforissier:
`Tested-by: Jerome Forissier <jerome.forissier@linaro.org> (HiKey960)`

jforissier:
And also...

PR #252, opened 2018-04-26
qemu_virt: boot from arm-tf and allow smp test

etienne-lms:
This build arm-trusted-firmware for qemu cortex-a15.
bios_qemu_tz_arm is still used as the primary non-secure boot stage.

Depends on https://github.com/OP-TEE/manifest/pull/110.

### DIFF
--- a/qemu.mk
+++ b/qemu.mk
@@ -14,10 +14,9 @@ include common.mk
 # Paths to git projects and various binaries
 ################################################################################
 ARM_TF_PATH			?= $(ROOT)/arm-trusted-firmware
+BINARIES_PATH			?= $(ROOT)/out/bin
 BIOS_QEMU_PATH			?= $(ROOT)/bios_qemu_tz_arm
 QEMU_PATH			?= $(ROOT)/qemu
-BINARIES_PATH			?= $(ROOT)/out/bin
-
 SOC_TERM_PATH			?= $(ROOT)/soc_term
 
 DEBUG = 1
@@ -25,9 +24,9 @@ DEBUG = 1
 ################################################################################
 # Targets
 ################################################################################
-all: arm-tf bios-qemu qemu soc-term linux
-clean: arm-tf-clean bios-qemu-clean linux-clean optee-os-clean \
-	qemu-clean soc-term-clean check-clean buildroot-clean
+all: arm-tf bios-qemu buildroot linux optee-os qemu soc-term
+clean: arm-tf-clean bios-qemu-clean buildroot-clean linux-clean optee-os-clean \
+	qemu-clean soc-term-clean check-clean
 
 include toolchain.mk
 
@@ -84,7 +83,7 @@ define bios-qemu-common
 		PLATFORM_FLAVOR=virt
 endef
 
-bios-qemu: buildroot optee-os linux
+bios-qemu:
 	$(call bios-qemu-common)
 	ln -sf $(ROOT)/out/bios-qemu/bios.bin $(BINARIES_PATH)
 
--- a/qemu.mk
+++ b/qemu.mk
@@ -155,6 +155,8 @@ run: all
 	ln -sf $(ROOT)/out-br/images/rootfs.cpio.gz $(BINARIES_PATH)/
 	$(MAKE) run-only
 
+QEMU_SMP ?= 1
+
 .PHONY: run-only
 run-only:
 	$(call check-terminal)
@@ -165,6 +167,7 @@ run-only:
 	(cd $(BINARIES_PATH) && $(QEMU_PATH)/arm-softmmu/qemu-system-arm \
 		-nographic \
 		-serial tcp:localhost:54320 -serial tcp:localhost:54321 \
+		-smp $(QEMU_SMP) \
 		-s -S -machine virt -machine secure=on -cpu cortex-a15 \
 		-d unimp  -semihosting-config enable,target=native \
 		-m 1057 \
@@ -181,7 +184,6 @@ ifneq ($(TIMEOUT),)
 check-args += --timeout $(TIMEOUT)
 endif
 
-QEMU_SMP ?= 1
 check: $(CHECK_DEPS)
 	cd $(BINARIES_PATH) && \
 		export QEMU=$(ROOT)/qemu/arm-softmmu/qemu-system-arm && \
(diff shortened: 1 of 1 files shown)


qemu.mk
```diff
@@ -41,13 +84,8 @@ define bios-qemu-common
 endef
 
 bios-qemu: buildroot optee-os linux
```
jenswi-linaro: The dependencies aren't needed any longer

etienne-lms: ok for dropping dep on optee-os.
buildroot and linux are still booted by bios but i think both beps can be discarded at build time. can you confirm?

jenswi-linaro: Agree. There's other targets that depend on linux and buildroot

jenswi-linaro:

jforissier:
> Note: once merged, one who syncs with latest **build.git** but did not clone arm-tf (or synced through the update manifest (OP-TEE/manifest#110) will get:
> ```
> make: *** /path/to/workenv/build/../arm-trusted-firmware: No such file or directory.  Stop.
> ``` 
> without further info. Are you ok with this?


@etienne-lms that is to be expected, and one of the reasons why I say `repo` is totally broken (it doesn't keep the changes to the Makefiles in the same commit as the changes in the manifest whici is fundamentally wrong).

Anyway, the error message seems pretty clear to me so I don't think we can do better (short of using git submodules that is ;-)).

jbech-linaro:
> Note: once merged, one who syncs with latest build.git but did not clone arm-tf (or synced through the update manifest (OP-TEE/manifest#110) will get:
> 
> make: *** /path/to/workenv/build/../arm-trusted-firmware: No such file or directory.  Stop.
> without further info. Are you ok with this?
> 
> @etienne-lms that is to be expected, and one of the reasons why I say repo is totally broken (it doesn't keep the changes to the Makefiles in the same commit as the changes in the manifest which is fundamentally wrong).
> 
> Anyway, the error message seems pretty clear to me so I don't think we can do better (short of using git submodules that is ;-)).

You can always try `repo sync` and for some odd reason the totally broken tool works for millions(?) of Android developers :)

jforissier:
> You can always try repo sync 

That's the problem, `repo sync` and the build may break at some point (for instance, after this PR is merged but manifest is not yet merged), then do `repo sync` again and it may work (after both PRs are merged). A proper tool should not allow that. Or is it our development model?
If I merge the manifest PR first, will it avoid this nasty window? (I don't think so).

> and for some odd reason the totally broken tool works for millions(?) of Android developers :)

Maybe it "mostly works" and they just accept it as it is. Or their workflow doesn't expose the flaws in the tool. I don't know.

jbech-linaro:
> Maybe it "mostly works" and they just accept it as it is. Or their workflow doesn't expose the flaws in the tool. I don't know.

Android uses the same setup as we are, but with tons of more gits. All source code gits (>600 gits) + the build git where you initiate the build system (`source build/envsetup.sh && lunch`). So they face the exact same problem and depending on how you are using your tools you can end up with out-of-sync gits no matter how your are doing things. If you merge a single git with any dependency to another git you are in this situation, repo doesn't solve that, git submodules doesn't solve that, clearcase doesn't solve that or not any other tool that I'm aware of. The only way to "work around" that is to wait until some integrator have given you a tag where he knows that things depending on each other has been merged. That type of tag is a repo manifest with only SHA-1's or git submodule only tracking SHA-1 (which is the only way to do things with git submodules IIRC).

The way you insist that submodule solves this is that it simply lags behind, i.e., if you commit something in optee_os.git, you still get a working build if you checkout the latest "optee_build (submodule thing)". Average Joe on the other hand, he might use git pull directly in optee_os and **boom** broken! But if he waits until you have merged also the dependency, then he can once again run "sync" on his submodule tree and things are fine again. But there is still this window where things can break. Can we agree on this?

If we do the same with repo using a stable tag, what happens? The exact same things will happen. Average Joe will break optee_os if he updates on his own directly in the git. But a new user still get a working setup with a stable setup. But when the manifest has been updated, even average Joe will get back to a working state. So exactly the same situation between repo and submodules. Agree?

I think what is being mixed up here is that we are using a IMHO nice feature in repo, that you can track upstream branches mixed with stable tags. I.e, **you** have a choice to either use stable tags **or** use the mix. It's up to you as a developer to decide. Latest and greatest with risk of broken things or just use the stable versions. I cannot see how a choice can be seen as a bad thing. I know you can manually checkout latest on submodules also, but once again, then we're back in the same situation things can break if there are dependencies? Agree on this one (proactive guess, no ...)?

jforissier:
>  That type of tag is a repo manifest with only SHA-1's or git submodule only tracking SHA-1 (which is the only way to do things with git submodules IIRC).

git submodules has a branch attribute **in addition** to the SHA1. So, the default behavior is to always checkout a known good configuration, that is true, while still allowing the user to easily update the submodules to their latest version, all this with no change upstream.
With repo on the other hand, you decide upstream whether you are tracking branches (then there is no "known good" configuration recorded anywhere), or pinning SHA1 but in the latter case you loose the branch name.

> But there is still this window where things can break. Can we agree on this?

Yes, but with git submodules this window is the result of a deliberate action of the user. He knows which submodules he has just updated, and it is easy to see what has changed compared to the known stable version (for instance `git diff --submodule` will shows a nice list of commits for each submodule that has changed and `git diff --submodule=diff` will show full diffs).
Compare with `repo`: the window is comprised between `repo sync` and `repo sync`. And when it's broken, how do you diff against the previous known good state (before the first "repo sync")?

> I think what is being mixed up here is that we are using a IMHO nice feature in repo, that you can track upstream branches mixed with stable tags.

I agree on this. If we were to use SHA1 only, the behavior would be quite similar to git submodules. Except that as I said, when doing so, you lose which branches should be tracked to get the latest stuff.

Back to the specific problem of dependencies between Makefiles and sub-projects. We could pin the version of build.git in our manifests. It would close the "build vs. manifest update" window. Not sure it's worth the effort however, because it seems that current `repo` users are happy with the current situation and it would certainly not make me switch to `repo` anyway ;-)

jbech-linaro:
> Back to the specific problem of dependencies between Makefiles and sub-projects. We could pin the version of build.git in our manifests. It would close the "build vs. manifest update" window. Not sure it's worth the effort however, because it seems that current repo users are happy with the current situation 

We will not change it, this works.

> and it would certainly not make me switch to repo anyway ;-)

Can't say I'm surprised. I can use either one, for me it is same same but different.

PR #253, opened 2018-04-27
qemu: fix directory creation for output binaries and check target

etienne-lms:
Fixes: 46ddff2b339b ("qemu: boot from arm-tf")


etienne-lms:
Sorry, this must be merge to get Travis tests back functional for all optee_os P-Rs.

qemu.mk
```diff
@@ -179,9 +181,8 @@ ifneq ($(filter check,$(MAKECMDGOALS)),)
 CHECK_DEPS := all
 endif
 
-check-args := --bios $(ROOT)/out/bios-qemu/bios.bin
```
jforissier: Shouldn't that be `bl1.bin` like in this patch: https://github.com/OP-TEE/build/commit/46ddff2b339b9457a37b3bd58b29e4a4e06dc10f#diff-e0bd85e5484a3cb990cdad4e8a7bcea8L129?

jenswi-linaro:

qemu.mk
jforissier: With this line replaced by: `check-args := --bios $(BINARIES_PATH)/bl1.bin` (and the `+=` kept below):

(have you actually tried `make check`?)

etienne-lms: Yes I tried `make check` and failed to get something out of it. I was even wondering whether we maintain this part...  Note: see https://github.com/OP-TEE/build/pull/254.

etienne-lms: That said, should we really set `check-args := --bios $(BINARIES_PATH)/bl1.bin`?
Note: in the qemu_v8.mk, the bios is not 'checked'.

etienne-lms:
> The mkdir additions do not seem to be related to the commit mentioned in the commit comment (Fixes ...), are they? => split in two patches.

The commit header specifies "create dir + fix target". But I can split in 2 commits.

I have another change again in the qemu, to align **qemu.mk** and **qemu_v8.mk** but I was waiting to submit that through a seperated P-R. Ok like that? (or rather 1 P-R for all qemu changes I want to push?)

etienne-lms:
I pushed with restored bios path in `check` target, as you Jérôme tested.
R-b and tested-by tags applied.

jforissier:
I still don't think the `Fixes:` tag is applicable to the first commit. And there should be no blank line between Fixes: and the other tags.

etienne-lms:
Well the 1st commit really fixes an issue. The 1st failing travis result i read this afternoon was a failure to create a symbolic link due to this directory not being created.
Thanks for the description of `check` :^)

jforissier:
I'm not saying it does not fix an issue. I'm saying the issue it fixes is not introduced by the commit you are mentioning.

etienne-lms:
Yes it does. The referred (merged) change includes the following for the `arm-tf` make target:
```
arm-tf: optee-os bios-qemu
	$(ARM_TF_EXPORTS) $(MAKE) -C $(ARM_TF_PATH) $(ARM_TF_FLAGS) all fip
	ln -sf $(ARM_TF_OUT)/bl1.bin $(BINARIES_PATH)
	ln -sf $(ARM_TF_OUT)/bl2.bin $(BINARIES_PATH)
	ln -sf $(OPTEE_OS_HEADER_V2_BIN) $(BINARIES_PATH)/bl32.bin 
```

Assuming it's a build from scratch, binary directory `$(BINARIES_PATH)` does not exits, first processed target is likely be `arm-tf` (see the order of dependencies for `all`) so the 2 first `ln -sf ...` above creates symlinks to files for `$(BINARIES_PATH)`. The next command fails as `$(BINARIES_PATH)` is a symlink to a file, not a directory (or equiv).

jforissier:
> Yes it does

OK, you're right. It fixes this one plus some older code, indeed.
You may add my:

...to both commits. Thanks!

etienne-lms:
thanks (for helping me to get my bugs fixed :)

etienne-lms:
tag applied

PR #256, opened 2018-05-02
qemu: make check: create symlink to rootfs.cpio.gz

jforissier:
Now that Arm Trusted Firmware is used as the secure boot loader
and bios_qemu_tz_arm is used as the non-secure bootloader, all binaries
are expected to be found in the $(BINARIES) directory. Symbolic links
are created by various targets in the qemu.mk.
The 'check' target should create a link to rootfs.cpio.gz, similar to
the 'run' target.

Fixes: 46ddff2b339b9 ("qemu: boot from arm-tf")

### DIFF
--- a/qemu.mk
+++ b/qemu.mk
@@ -187,6 +187,7 @@ check-args += --timeout $(TIMEOUT)
 endif
 
 check: $(CHECK_DEPS)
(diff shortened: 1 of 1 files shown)


jbech-linaro:

PR #260, opened 2018-05-09
am32xx am57xx dra7xx: build using buildroot

jenswi-linaro:
Changes for the TI platforms to use buildroot instead. I'm not able to test this myself. @igoropaniuk, @glneo, can you help?

To build this:
```
repo init -u https://github.com/jenswi-linaro/manifest.git -m dra7xx.xml -b remaining-br
cd build
git remote add jenswi https://github.com/jenswi-linaro/build.git
git fetch jenswi
git checkout jenswi/remaining-br
```
then build and flash as usual.

### DIFF
--- a/am43xx.mk
+++ b/am43xx.mk
@@ -8,6 +8,9 @@ override COMPILE_NS_KERNEL := 32
 override COMPILE_S_USER    := 32
 override COMPILE_S_KERNEL  := 32
 
+# Need to set this before including common.mk
+BUILDROOT_GETTY_PORT ?= ttyS0
+
 ###############################################################################
 # Includes
 ###############################################################################
@@ -28,7 +31,6 @@ FIT_MAKEFILE    ?= $(BUILD_PATH)/ti/Makefile
 OPTEE_PLATFORM  ?= ti-am43xx
 U-BOOT_CONFIG   ?= am43xx_hs_evm_defconfig
 CONFIG_TYPE     ?= ti_sdk_am4x_debug
-BUSYBOX_TARGET  ?= am43xx
 
 ###############################################################################
 # Include common to TI builds
--- a/am57xx.mk
+++ b/am57xx.mk
@@ -8,6 +8,9 @@ override COMPILE_NS_KERNEL := 32
 override COMPILE_S_USER    := 32
 override COMPILE_S_KERNEL  := 32
 
+# Need to set this before including common.mk
+BUILDROOT_GETTY_PORT ?= ttyS2
+
 ###############################################################################
 # Includes
 ###############################################################################
@@ -29,7 +32,6 @@ OPTEE_PLATFORM  ?= ti-am57xx
 U-BOOT_CONFIG   ?= am57xx_hs_evm_defconfig
 # using the same configs as for DRA7xx
 CONFIG_TYPE     ?= ti_sdk_dra7x_debug
-BUSYBOX_TARGET  ?= dra7xx
 
 ###############################################################################
 # Include common to TI builds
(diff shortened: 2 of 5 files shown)


jenswi-linaro:
Depends on https://github.com/OP-TEE/manifest/pull/113

glneo:
The patch subject is wrong, s/am32xx/am43xx

glneo:
The documentation will need to be updated [here.](https://github.com/OP-TEE/build/blob/master/docs/ti.md#31-sd-card-boot)

As this currently uses:
```<repo directory>/gen_rootfs/filesystem.cpio.gz```
will need to be:
```<repo directory>/out-br/images/rootfs.cpio.gz```

**Comment by jenswi-linaro** [comment]
Sorry, I'll update subject when doing the final commit.
I'm off for some vacation a couple of days now, I'm back on Monday.

**Comment by glneo** [comment]
Okay, one more thing before I forget, the rootfs doesn't include the fitImage, this is added previously here:
https://github.com/OP-TEE/build/blob/master/ti/ti-common.mk#L126

**Comment by glneo** [comment]
I am getting:
```
Starting logging: OK
Initializing random number generator... done.
Starting tee-supplicant...
Starting network: OK
can't open /dev/ttyAMA0: No such file or directory
can't open /dev/ttyAMA0: No such file or directory
can't open /dev/ttyAMA0: No such file or directory
can't open /dev/ttyAMA0: No such file or directory
...
```

The default with buildroot seems to be ttyAMA0, we use ttyS0/2.

With [gen_rootfs](https://github.com/linaro-swg/gen_rootfs/blob/master/etc/inittab-dra7xx) we had just launched a shell on both those serial ports, as different boards used different ports.

**Comment by glneo** [comment]
After manually fixing the above looks good:

```
15851 subtests of which 0 failed
70 test cases of which 0 failed
0 test case was skipped
TEE test application done!
```

One last question, why buildroot? The old way was much faster to build and didn't require so many external downloads and tools that we don't seem to use anyway.

**Comment by jbech-linaro** [comment]
> One last question, why buildroot?

Main reason for this is that it started to become a bit challenging to add new normal world user space tools to the root fs (in our particular case we needed OpenSSL). With Buildroot you can just enable whatever developer tool you need (valgrind, strace, ltrace, gdb etc...). But, it's definitely true that it is much slower to build now. My HiKey automatic testing went from (full cycle, build, flash, xtest) ~6-7 minutes to ~13 minutes and that it only because of buildroot. Hopefully we can improve that a bit in the future.

**Comment by glneo** [comment]
Thanks for the explanation. That works for me, saving development time and effort is much more important than a few saved minutes of build time.

**Comment by jenswi-linaro** [comment]
@glneo, thanks for testing. I think I've done the required changes.

Please note
```BUILDROOT_GETTY_PORT ?= ttyS0
```
for each TI platform. I guess at least one of them should have a different value.

glneo:
am57xx.mk should have ttyS2

glneo:
Now I'm getting:
```
‘/mnt/hdd/optee/build/../u-boot/u-boot-spl_HS_MLO’ -> ‘/boot/MLO’
install: cannot create regular file ‘/boot/MLO’: Permission denied
```
Looks like $(BUILDROOT_TARGET_ROOT) is not defined for me.

To test build, you can use [this project](https://github.com/glneo/dummy-secdev):
```
$ git clone https://github.com/glneo/dummy-secdev.git
$ export TI_SECURE_DEV_PKG=/location/of/git/repo/dummy-secdev
```
After that, building this project should work.

jenswi-linaro:
Thanks, BUILDROOT_TARGET_ROOT fixed now also.

glneo:
Everything seems good now, thanks for the platform improvements,

Acked-by: Andrew F. Davis <afd@ti.com>

jenswi-linaro:
Tag applied

PR #263, opened 2018-05-23
Remove legacy toolchain

jbech-linaro:
This is a follow-up to the email thread with Ard about running newer toolchains with the GCC49 profile in UEFI.

I've done build testing on all platforms mentioned (and travis did build test for Juno). Then I have been running xtest on QEMU v8, FVP and RPi3 (which is about U-boot ...) all without regressions.


jenswi-linaro:
How about AOSP, isn't gcc 4.9 use there? Or that's still dealt with in a different way?

jforissier:
Nice!

jenswi-linaro:

jbech-linaro:

jforissier:
I think those things are not needed now and should be removed:
https://github.com/OP-TEE/build/blob/master/common.mk#L115-L125
https://github.com/OP-TEE/build/blob/master/common.mk#L189-L193
https://github.com/OP-TEE/build/blob/master/br-ext/configs/toolchain-aarch32-legacy

@jenswi-linaro?

jenswi-linaro:
https://github.com/OP-TEE/build/blob/master/common.mk#L115-L125 and
https://github.com/OP-TEE/build/blob/master/common.mk#L189-L193
should be cleaned up
https://github.com/OP-TEE/build/blob/master/br-ext/configs/toolchain-aarch32-legacy
can be removed

PR #267, opened 2018-06-05
buildroot: add host-python-pycrypto dependency

lorc:
Currently sign.py uses system python to sign TAs. Problem is
that buildroot can build own host-python. This python lacks
pycrypto, so sign.py fails.

This patch adds dependency on host-python-pycrypto, so buildroot
will always build python with needed module.

Signed-off-by: Volodymyr Babchuk <vlad.babchuk@gmail.com>


jenswi-linaro:

lorc:
Added tag.

PR #268, opened 2018-06-11
Rpi3: build using Buildroot

jbech-linaro:
Depends on https://github.com/OP-TEE/manifest/pull/116.

```bash
U-Boot 2016.03-00725-gbc6d93cf5e (Jun 11 2018 - 21:07:15 +0200)
...
I/TC:  OP-TEE version: 3.1.0-141-ga0a7d92c #1 Mon Jun 11 19:07:19 UTC 2018 aarch64
...
[    0.000000] Linux version 4.6.3-01139-g7d976fc52ae3 (jbech@debby) (gcc version 6.2.1 20161016 (Linaro GCC 6.2-2016.11) ) #1 SMP Mon Jun 11 21:16:51 CEST 2018
...
buildroot login: root
# xtest
...
16055 subtests of which 0 failed
74 test cases of which 0 failed
0 test case was skipped
TEE test application done!
#

### DIFF
--- a/rpi3.mk
+++ b/rpi3.mk
@@ -27,14 +27,14 @@ U-BOOT_PATH		?= $(ROOT)/u-boot
 U-BOOT_BIN		?= $(U-BOOT_PATH)/u-boot.bin
 U-BOOT_RPI_BIN		?= $(U-BOOT_PATH)/u-boot-rpi.bin
 
-RPI3_FIRMWARE_PATH		?= $(BUILD_PATH)/rpi3/firmware
-RPI3_HEAD_BIN			?= $(ROOT)/out/head.bin
-RPI3_BOOT_CONFIG		?= $(RPI3_FIRMWARE_PATH)/config.txt
-RPI3_UBOOT_ENV			?= $(ROOT)/out/uboot.env
-RPI3_UBOOT_ENV_TXT		?= $(RPI3_FIRMWARE_PATH)/uboot.env.txt
-RPI3_STOCK_FW_PATH		?= $(ROOT)/firmware
+RPI3_FIRMWARE_PATH	?= $(BUILD_PATH)/rpi3/firmware
+RPI3_HEAD_BIN		?= $(ROOT)/out/head.bin
+RPI3_BOOT_CONFIG	?= $(RPI3_FIRMWARE_PATH)/config.txt
+RPI3_UBOOT_ENV		?= $(ROOT)/out/uboot.env
+RPI3_UBOOT_ENV_TXT	?= $(RPI3_FIRMWARE_PATH)/uboot.env.txt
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: rpi3.mk [hunk 1]
```diff
@@ -44,9 +47,9 @@ ifeq ($(CFG_TEE_BENCHMARK),y)
 all: benchmark-app
 clean: benchmark-app-clean
 endif
-all: arm-tf optee-os optee-client xtest u-boot u-boot-rpi-bin\
+all: arm-tf buildroot optee-os optee-client xtest u-boot u-boot-rpi-bin\
```
jenswi-linaro: builtroot is managing `optee-client`, `optee-examples` and `xtest`, no need for that here or below in the clean

rpi3.mk
```diff
@@ -44,9 +47,9 @@ ifeq ($(CFG_TEE_BENCHMARK),y)
...
 	linux update_rootfs optee-examples
-clean: arm-tf-clean busybox-clean u-boot-clean u-boot-rpi-bin-clean \
+clean: arm-tf-clean buildroot-clean busybox-clean u-boot-clean u-boot-rpi-bin-clean \
```
jenswi-linaro: `busybox-clean` can be removed also

rpi3.mk
jbech-linaro: I'll remove, what about the targets further down in the same file?
```Makefile
optee-client: optee-client-common 
... and friends
```
I can see that they are still in qemu.mk etc.

jenswi-linaro: Those should be removed too.
They where left by mistake in qemu.mk

rpi3.mk
```diff
@@ -44,11 +47,10 @@ ifeq ($(CFG_TEE_BENCHMARK),y)
 all: benchmark-app
```
jenswi-linaro: All the benchmark-app stuff can be removed too.

jbech-linaro: Right, but when doing that I start to wonder how do we propagate the `CFG_TEE_BENCHMARK` flag?

jbech-linaro: In other words, shouldn't the CFG... be moved to surround this?
https://github.com/OP-TEE/build/blob/master/common.mk#L218-L219

jbech-linaro: Never mind ... spotted it now further down 2 seconds after posting the comments :)

jenswi-linaro: https://github.com/OP-TEE/build/blob/d496f98386525b0379116707de56883057469e8d/common.mk#L223-L225

jenswi-linaro:

jbech-linaro:
Dependencies ready, squashed, rebased and tag(s) applied (should be) ready for merge! Thanks!

PR #290, opened 2018-08-17
buildroot: tee-supplicant: use -d (daemonize) option and print OK/FAIL

jforissier:
Init scripts normally print OK or FAIL when they start daemons. Update
our tee-supplicant script to do just that. The -d option to tee-supplicant
is used so that when the tee-supplicant command returns, we  know for sure
if initialization went well or not.


jforissier:
Depends on https://github.com/OP-TEE/optee_client/pull/132.

jforissier:
Merged via https://github.com/OP-TEE/build/pull/291.

PR #292, opened 2018-08-21
qeu: buildroot: always enable networking

jforissier:

### DIFF
--- a/docs/qemu.md
+++ b/docs/qemu.md
@@ -45,12 +45,14 @@ To mount host folder in QEMU, simply run:
 ```bash
 $ mount -t 9p -o trans=virtio host <mount_point>
 ```
-# 5. SLiRP user networking
-To enable SLiRP user networking just set `QEMU_USERNET_ENABLE ?= y` in `common.mk`.
-After booting QEMU VM, eth0 will automatically receive IP address via DHCP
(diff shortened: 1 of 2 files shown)


docs/qemu.md
```diff
@@ -45,12 +45,5 @@ To mount host folder in QEMU, simply run:
 ```bash
 $ mount -t 9p -o trans=virtio host <mount_point>
 ```
-# 5. SLiRP user networking
-To enable SLiRP user networking just set `QEMU_USERNET_ENABLE ?= y` in `common.mk`.
-After booting QEMU VM, eth0 will automatically receive IP address via DHCP
-
-*Important* Take into account that ICMP doesn't work in SLiRP mode,
```
jenswi-linaro: Is this note still relevant?

jforissier: It is. Well, `ping localhost` works, but pinging external hosts does not.

jforissier:
@jenswi-linaro almost no additional delay to boot with this patch. IP configuration takes no more than 1 second.

jenswi-linaro:
Please keep the description of networking. With that:

jforissier:
@jenswi-linaro I have slightly reworded the doc and applied your R-b. Thanks.

PR #293, opened 2018-08-21
buildroot: add support for debugging of Normal World applications

jforissier:
Build with GDBSERVER=y to enable remote debugging of Normal World
applications. gdbserver is added to the root filesystem.


jbech-linaro:
Nice! Haven't tested it, but looks good!

jbech-linaro:
> ... so I guess we'd rather clean this up in a separate PR moving to reStructuredText as we've discussed already (then we can use the .. contents:: directive).

Well inline with my proposal for the themed hacking session :)

docs/qemu.md
```diff
@@ -53,6 +53,38 @@ gateway to the host network [SLiRP].
...
+# gdbserver :12345 xtest 4002
+Process xtest created; pid = 654
+Listening on port 12345
+
+```
+Back on your PC, in another terminal, start GDB and connect to the target:
+```
+$ ../out-br/host/bin/arm-buildroot-linux-gnueabihf-gdb
+(gdb) set sysroot ../out-br/host/arm-buildroot-linux-gnueabihf/sysroot
+(gdb) target remote :12345
```
igoropaniuk: btw, maybe a bit off-topic (just my curiosity), but have you tried to debug using both gdbservers mutually (the one integrated into QEMU, and another running as a user app inside QEMU), for example, if you need to debug both CA and TA.

In case if the breakpoint is hit in the external gdbserver (integrated into QEMU), does another gdb instance loose connection to the internal one (by timeout)

jforissier: Not tried but each server has its own TCP port so they should be pretty much independent.

PR #294, opened 2018-08-22
buildroot: add teeclnt group and test user account

jforissier:
Restrict execution of TEE client applications to members of the teeclnt
group. A test user is created and added to this group.
Having two groups (tee and teeclnt) allows to better isolate
tee-supplicant from the client applications.


jforissier:
@jenswi-linaro is this what you had in mind?

jenswi-linaro:
Yes

PR #299, opened 2018-08-28
fvp: use DT file foundation-v8-gicv3-psci.dtb

jforissier:
The Device Tree configuration we need for the Foundation Platform is now
available upstream (v4.18) as foundation-v8-gicv3-psci.dts (it only lacks
the optee node and memory reservations). Let's use this new file instead
of a hacked version of foundation-v8.dts modified for PSCI and GiCv3.


jforissier:
Depends on Linux v4.18 so this should be merged together with an update to manifest.git.

jenswi-linaro:

jforissier:
For the record, there is a regression in the secure data path test (`xtest 1014`, tested on QEMU with `make CFG_SECURE_DATA_PATH=y run`). The problem is caused by Linux commit [54ef5b9db767](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=54ef5b9db767) ("staging: android: ion: Initialize dma_address of new sg list").

Holding off this PR until the issue is addressed.

PR #300, opened 2018-08-31
buildroot: optee_client: add ion group and set permissions on /dev/ion

jforissier:
If /dev/ion exists, it should be set up so that the test user can access
it. Otherwise, some tests will fail, such as xtest 1014 (Secure Data
Path). Similar to what is done for /dev/tee*, this commit configures
permissions based on a specific group: ion.


jenswi-linaro:

PR #302, opened 2018-09-06
Update toolchain.mk

JohnyDev42:
This option is required for insecure connection

### DIFF
--- a/toolchain.mk
+++ b/toolchain.mk
@@ -22,7 +22,7 @@ define dltc
 	@if [ ! -d "$(1)" ]; then \
 		mkdir -p $(1); \
 		echo "Downloading $(3) ..."; \
-		curl -s -L $(2) -o $(TOOLCHAIN_ROOT)/$(3).tar.xz; \
+		curl -s -L -k $(2) -o $(TOOLCHAIN_ROOT)/$(3).tar.xz; \
 		tar xf $(TOOLCHAIN_ROOT)/$(3).tar.xz -C $(1) --strip-components=1; \
 	fi
 endef


jenswi-linaro:
I think this is OK, but then we should check that the sha1sums of downloaded binaries matches expected values. We should check the sha1sums regardless of this switch though.

ArunSahu42:
Is threre any way to check the sha1sums of the downloaded binaries

jbech-linaro:
Right, that seems to be what @ryankumar is facing in https://github.com/OP-TEE/optee_os/issues/2521. But I think what Jens mentioned is that if we introduce the `-k`, then we should check that that downloaded toolchain is what we expect, so in psuedo code you do:
```c
toolchain_sha256sum = "abababababababab...." # this is pre-calculated and hardcoded in here
curl ... toolchain.xz
if (sha256sum(toolchain.xz) != toolchain_sha256sum)
    print ("Toolchain hash mismatch")
    return error
```
Something like that is what should be done in `toolchain.mk`.

ArunSahu42:
I am not getting why this sha256sum is required...if I will download this toolchain insecurely does it effect my sha256sum and why ?

jenswi-linaro:
Man in the middle...

wamserma:
Hm, I think this PR could be closed w/o merging.

jbech-linaro:
> Hm, I think this PR could be closed w/o merging.

Agree and we haven't seen anyone having this issue for quite some time now. If a reader of this still think this is an issue, then please open the ticket again. I'm closing it for now. Thanks!

PR #313, opened 2018-10-05
qemu, qemu_v8: set QEMU_SMP ?= 2

jforissier:
Use two vCPUs by default instead of one. This configuration is more
likely to trigger race conditions when testing.


jenswi-linaro:

PR #316, opened 2018-10-25
buildroot: use QEMU overlay only when platform is QEMU or QEMUv8

jforissier:
common.mk mistakenly uses the QEMU overlay (which currently contains
only an init script for udhcpc) on all platforms. This is incorrect and
causes problems on RPi3 at least [1].

Fixes: commit 751b35bb0b84 ("qemu: buildroot: fix networking")
Links: [1] https://github.com/OP-TEE/optee_os/issues/2478#issuecomment-432818199


jenswi-linaro:

PR #317, opened 2018-10-29
hikey960: add config fragments required for upstream kernel v4.19

jforissier:
Plain upstream kernel v4.19 now supports HiKey960 well enough, so we do
not need to use the 96boards repository anymore [1]. Setting a few config
symbols is needed however, otherwise the kernel can not boot our
buildroot-generated root FS. This commit introduces the required symbols
so that we can later drop the 96boards branch (when we rebase our optee
branch [2] onto v4.19). Note that having these symbols enabled is OK with
the current branch, too.

Link: [1] https://github.com/96boards-hikey/linux
Link: [2] https://github.com/linaro-swg/linux/tree/optee


jforissier:
Rebased optee linux branch is here for the moment: https://github.com/jforissier/linux/tree/wip/optee-v4.19

jenswi-linaro:

jbech-linaro:
> Plain upstream kernel v4.19 now supports HiKey960 well enough, so we do
> not need to use the 96boards repository anymore [1].

Nice, thanks!

jforissier:
@jbech-linaro IBART fails to clone the Gits?

jbech-linaro:
> @jbech-linaro IBART fails to clone the Gits?

Let me check, it could be that the delta is big and therefore require more time. Having that said, I have a cronjob running nightly so if I haven't missed including kernel, then the delta shouldn't be that big. Anyhow, I'll check.

jbech-linaro:
Can't say what went wrong. I showed timeout in the logs. So I re-synced the mirror before re-running and now it seems like it is working again.

PR #318, opened 2018-11-09
qemu: remove MULTI_CONSOLE=0 to support TF-A v2.0

jforissier:
Adds support for Arm Trusted Firmware-A version v2.0.

The qemu platform in TF-A v2.0 does not acccept MULTI_CONSOLE=0 anymore,
so remove it. The current version v1.5 is quite happy without it already.


jenswi-linaro:

PR #319, opened 2018-11-09
fvp: use GICv3 to support TF-A v2.0

jforissier:
Adds support for Arm Trusted Firmware-A version v2.0.

TF-A v2.0 does not acccept FVP_USE_GIC_DRIVER=FVP_GICV3_LEGACY anymore.
Therefore, use FVP_GICV3 instead. This is the default value with v2.0, and
is also supported by earlier versions (such as 1.5 which is the one we are
using currently).

OP-TEE is configured accordingly (CFG_ARM_GICV3=y).

~~** Note ** Depends on an optee_os patch, see PR 2548.~~ That was PR 2560, merged.


jenswi-linaro:

PR #321, opened 2018-11-19
Enable Plan9 sharing and home passthrough by default

HernanGatta:
This PR enables Plan9 sharing by default on QEMU builds for both ARMv7 and ARMv8. Plan9 sharing is used to mount host directories inside a QEMU guest. This, in turn, would be used by developers of OE enclaves to copy their programs into the emulated environment.

This PR also makes it so that when a user runs `make run`, QEMU automatically passes their home folder into the guest as read-only by default.


HernanGatta:
Sorry, created this against the wrong base.

PR #336, opened 2019-01-24
Fix the DEBUG build for QEMU_V8

Hmtir:
When the DEBUG config was set, the optee_os was still built with
DEBUG=0 because the flag was hardcoded. As a result the optimization
was still O2. This is fixed by using the global DEBUG flag instead of
the hardcoded value of zero.

Signed-off-by: Hashem Tatari <hashem.tatari@exset.com>

### DIFF
--- a/qemu_v8.mk
+++ b/qemu_v8.mk
@@ -133,7 +133,7 @@ linux-cleaner: linux-cleaner-common
 # OP-TEE
 ################################################################################
 OPTEE_OS_COMMON_FLAGS += PLATFORM=vexpress-qemu_armv8a CFG_ARM64_core=y \
-			 DEBUG=$(DEBUG) CFG_PM_DEBUG=0
+			 DEBUG=$(DEBUG)
 optee-os: optee-os-common
 
 OPTEE_OS_CLEAN_COMMON_FLAGS += PLATFORM=vexpress-qemu_armv8a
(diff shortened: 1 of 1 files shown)


qemu_v8.mk
```diff
@@ -133,7 +133,7 @@ linux-cleaner: linux-cleaner-common
 # OP-TEE
 ################################################################################
 OPTEE_OS_COMMON_FLAGS += PLATFORM=vexpress-qemu_armv8a CFG_ARM64_core=y \
-			 DEBUG=0 CFG_PM_DEBUG=0
+			 DEBUG=$(DEBUG) CFG_PM_DEBUG=0
```
etienne-lms: by the way, `CFG_PM_DEBUG` is a deprecated thing, should be removed.

Hmtir: Yes I was not sure about that that's why I didn't change it.
I will also remove it then.

etienne-lms:
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` with jforissier's comment addressed.
@Hmtir, can you update the P-R? It looks fine and is valuable.

jbech-linaro:
Closing the PR, since I've manually merged the patches (after fixing Jerome's comment and adding the tags).

PR #337, opened 2019-01-24
Update the launch_terminal definition in case of using LAUNCH_TERMINAL

Hmtir:
Add quotes around the soc_term and the port number to make it compatible with other terminals like terminator.

To use terminator it is enough to set the flag like:
$ export LAUNCH_TERMINAL="terminator -e"
$ make run-only

Note in this case the title of the terminals cannot be set properly

Signed-off-by: Hashem Tatari <hashem.tatari@exset.com>

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -336,10 +336,12 @@ define run-help
 	@echo
 endef
 
+# Note: Using the LAUNCH_TERMINAL environment variable, it is not currently possible to set 
+# different titles for the terminals because there is no any common way among all the terminals
 ifneq (, $(LAUNCH_TERMINAL))
 define launch-terminal
 	@nc -z  127.0.0.1 $(1) || \
-		$(LAUNCH_TERMINAL) $(SOC_TERM_PATH)/soc_term $(1) &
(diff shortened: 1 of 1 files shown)


jbech-linaro:
Haven't tested it, but looks OK, so as long as it doesn't break our current setup's I'm happy with this.

jforissier:
@jbech-linaro I have checked that the original use case mentioned in the commit that introduced `LAUNCH_TERMINAL` [1] still works OK with the quotes.

[1] commit 07580605eabd ("common.mk: allow user to override launch-terminal in make run")

@Hmtir please fix the spelling mistake in the commit subject (`LAUNCH_TERMINLAL`) and wrap lines in the commit message at 75 columns. With that:

Hmtir:
> @Hmtir please fix the spelling mistake in the commit subject (`LAUNCH_TERMINLAL`) and wrap lines in the commit message at 75 columns. With that:
Done

jforissier:
@Hmtir I was expecting you to also add the Ack/Review tags ;-) sorry for not being clear.
Anyway, no big deal, I have added them and merged the patch manually.
Thanks for the contribution!

PR #340, opened 2019-01-28
toolchain: Update to GCC82 with TUI enabled

### LINKED ISSUE #334: Update toolchains to (re-)enable TUI support
In the current toolchain (`gcc-arm-8.2-2018.08...`) the TUI mode is disabled which in turns make it harder to use GDB (see https://bugs.linaro.org/show_bug.cgi?id=4130 for details). Arm have just confirmed that they've re-enabled that in their recent builds [1] and I have cross checked that it is indeed working. So, my suggestion is that after the 3.4.0 release we just simply update `toolchain.mk` with the most current version, i.e. [1].

[1] 8.2-2019.01 --> https://developer.arm.com/open-source/gnu-toolchain/gnu-a/downloads

jbech-linaro:
This updates to a more recent version of GCC/GDB which has TUI mode enabled again. However, there are a couple of things that needs to be merged before we can continue with this patch set. I.e., **hold off merging this until**:
1. The Buildroot [patch upstream](http://patchwork.ozlabs.org/patch/1031920/) has been merged.
2. And we have updated manifest.git with that new Buildroot patch.

Fixes #334.


jbech-linaro:
Just gave this a try running latest with Buildroot, the build issue is still there, i.e. I haven't bothered pushing more for getting my Buildroot patch merged that would fix the build issue, since it was hard to figure out what the Buildroot maintainers actually wanted. It's probably already too late to get it merged in time for the OP-TEE 3.5.0 release, so we can just as well let this PR sit a month or two more.

jbech-linaro:
I've force pushed two new patches instead of closing this and creating an almost identical PR. I've tested this on both Qemu v7/v8 and I've tested it with both the new gcc8.3 as well as the old gcc 8.2 (2018.09). Both seems to work fine.

However, there is a dependency to https://github.com/OP-TEE/manifest/pull/135, since all this needs a more recent Buildroot version. So, merge order is https://github.com/OP-TEE/manifest/pull/135 and then this one. I'll update with a new comment when I believe this one is ready to be merged.

jbech-linaro:
With https://github.com/OP-TEE/manifest/pull/135 just being merged I've done a successful builds with Qemu v7/v8 and HiKey. For Qemu v7 I've both built and run xtest successfully. So, I need a review on this and once done, I think we can finally merge this and have a default toolchain with TUI mode enabled in GDB again.
(I've restarted IBART build, in queue right now, but I guess the job should end successfully here).

jforissier:

But I don't think the 4.18 -> 4.19 change is compatible with the previous compiler. You should remove this sentence from the commit message.

jbech-linaro:
> But I don't think the 4.18 -> 4.19 change is compatible with the previous compiler. You should remove this sentence from the commit message.

You're right and IBART which uses the old GCC 8.2 seems unhappy (`14405:  Incorrect selection of kernel headers: expected 4.19.x, got 4.18.x`), I thought I checked that explicitly yesterday, but I had ~5 builds running in parallel mixing old and new toolchains (+ AOSP builds), so there's a high risk that I goofed things up.

So, shall we:
1. Ignore the IBART error and rely on my [sanity builds done today](https://github.com/OP-TEE/build/pull/340#issuecomment-493430556) and then merge it.
 1.1 Then I'll immediately replace the Gcc8.2 with Gcc8.3 on IBART.
 or

2. Replace Gcc8.2 with Gcc8.3 on IBART
 2.1 Then Rebuild this PR which hopefully should be green and then merge.
 2.1. Then merge.

Approach 1: Will give IBART errors for new PRs/patch updates while working with "1.1" (10-15 minutes).

Approach 2: Will give IBART errors on all PRs/patch updates until this particular PR has been merged.

I vote for approach 1.

jforissier:
@jbech-linaro I'm fine with approach 1.

jbech-linaro:
> @jbech-linaro I'm fine with approach 1.

Ok, I'll fix the commit message and add tags.

jbech-linaro:
Re-tried with gcc 8.2 and this patch set. Got successful builds on both Qemu v8 and HiKey. Not sure though why Ibart isn't happy with it.

jforissier:
@jbech-linaro have you cleaned buildroot first? (`make buildroot-clean` or `rm -rf ../build/out-br`).

jbech-linaro:
Yes, my last try with Hikey was with rm -rf on the whole repo forest.

Den fre 17 maj 2019 16:58Jérôme Forissier <notifications@github.com> skrev:

jbech-linaro:
Numerous builds on two different machines for HiKey, this is my conclusion:

| Toolchain | Kernel headers | Result (Debian / Ubuntu) |
| --- | --- | --- |
| 8.2 | 4.18 | OK / OK|
| 8.2 | 4.19 | NOK / NOK|
| 8.3 | 4.18 | NOK / (didn't run) |
| 8.3 | 4.19 | OK / (didn't run) |

This is pretty annoying, since this means that you're forced to use GCC8.3 for our OP-TEE developer setup's after introducing this patch (contrary you cannot step GCC if you aren't doing this). IIRC in the past it's been somewhat flexible when it comes to use a bit older versions of GCC also. Or maybe I'm mixing things up with our old way of building CPIO archives directly.

Anyhow, since it's weekend, which means basically no builds on IBART, I've updated toolchain there to 8.3 and hopefully that should resolve this.

Also note that I squashed the two commits, since keeping them apart will create a build regression regardless the order of the patches. I.e., it would break a git bisect session for example. I've also reworked the commit message, to something that hopefully is more clear than before.

jbech-linaro:

jforissier:
Note, AFAICT, the "kernel headers" version refers to the headers that were used to build the toolchain, not to the kernel headers installed on the build machine.
So, what you say:
> this means that you're forced to use GCC8.3 for our OP-TEE developer setup's after introducing this patch
...is not surprising.
If we could find a GCC 8.3 build on a machine with kernel headers 4.18, then things would be different (old and new compiler could be used indifferently).

I don't think all this matters much. Users will download the new toolchain and that's it!

PR #341, opened 2019-02-01
Rpi3 nfs fixes

jbech-linaro:
Some general fixes and one of them:

Fixes: https://github.com/OP-TEE/optee_os/issues/2773

### DIFF
--- a/br-ext/package/optee_client/S30optee
+++ b/br-ext/package/optee_client/S30optee
@@ -19,6 +19,10 @@ case "$1" in
 			chown root:ion /dev/ion && chmod 0660 /dev/ion
 			[ $? = 0 ] && echo "OK" || echo "FAIL"
 		fi
+		printf "Create and set permissions on /data... "
+		mkdir -p /data/tee && chown -R tee:tee /data/tee && \
+			chmod 0770 /data/tee
+		[ $? = 0 ] && echo "OK" || echo "FAIL"
 		printf "Starting tee-supplicant... "
 		su tee -c '/usr/sbin/tee-supplicant -d'
 		[ $? = 0 ] && echo "OK" || echo "FAIL"
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 /serial1.log
 .gdb_history
 gdb
+pi3.cfg
--- a/br-ext/package/optee_client/optee_client.mk
+++ b/br-ext/package/optee_client/optee_client.mk
@@ -28,9 +28,4 @@ define OPTEE_CLIENT_USERS
 	test -1 test -1 - - /bin/sh teeclnt,ion Test user, may run TEE client applications
 endef
 
-define OPTEE_CLIENT_PERMISSIONS
-	/data d 755 root root - - - - -
-	/data/tee d 770 tee tee - - - - -
-endef
-
 $(eval $(cmake-package))
--- a/docs/rpi3.md
+++ b/docs/rpi3.md
@@ -110,8 +110,9 @@ filesystem accordingly.
 
 In the description below we will use the following terminology:
 ```
-HOST_IP=192.168.1.100   <--- This is your desktop computer
-RPI_IP=192.168.1.200    <--- This is the Raspberry Pi
+192.168.1.100   <--- This is your desktop computer
+192.168.1.200   <--- This is the Raspberry Pi
+/srv/nfs/rpi    <--- Location for the NFS share
 ```
 
 ## 5.1 Configure NFS
@@ -152,7 +153,7 @@ so update accordingly.
 
 ```bash
 $ cd /srv/nfs/rpi
-$ sudo gunzip -cd /home/jbech/devel/optee_projects/rpi3/out-br/images/rootfs.cpio.gz | sudo cpio -idmv
+$ sudo gunzip -cd <rpi3-project>/out-br/images/rootfs.cpio.gz | sudo cpio -idmv
 $ sudo rm -rf /srv/nfs/rpi/boot/*
 ```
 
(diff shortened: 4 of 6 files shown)


jforissier:
> rpi3: give user 'tee' permission to write to /data
I don't think user 'tee' should be allowed to write to `/data` because `/data` could contain things that are unrelated to TEE (hence why there is `/data/tee` for this purpose).

jbech-linaro:
I think you're right, so, what about this (it works fine, just tested)
```patch
-               mkdir -p /data/tee/logs && chown -R root:tee /data && \
-                       chmod 0770 /data
+               mkdir -p /data/tee/logs && chown -R tee:tee /data/tee && \
+                       chmod 0770 /data/tee
```
The only thing I'm concerned about is the other thing you mentioned in the other ticket:
   "@jforissier *One could decide to use a different, existing parent directory to host the tee directory, for instance.*"

jbech-linaro:
There is one really weird thing about this and I suspect it has something with timing and/or NFS to do. After untar the entire root fs. I get (`jbech` is `tee` here, both maps to 1000)

NFS share:
```
# ls -al data/
total 12
drwxr-xr-x  3 root  root  4096 Feb  1 13:50 .
drwxr-xr-x 19 root  root  4096 Feb  1 13:50 ..
drwxrwx---  2 jbech jbech 4096 Jan 31 10:18 tee
```

On the PI
```
# ls -al /data
total 12
drwxr-xr-x    3 root     root          4096 Feb  1  2019 .
drwxr-xr-x   19 root     root          4096 Feb  1  2019 ..
drwxrwx---    2 tee      tee           4096 Jan 31  2019 tee
```

Still, without my S30... fixes we get the `ERR [91] TSUP:main:684: error tee_supp_fs_init`.

**After** applying my S30 fix in this PR, it looks identical, the only difference is that the date changed, but permission are the same.

NFS share:
```
# ls -al data/
total 12
drwxr-xr-x  3 root  root  4096 Feb  1 13:50 .
drwxr-xr-x 19 root  root  4096 Feb  1 13:50 ..
drwxrwx---  3 jbech jbech 4096 Feb  1 13:55 tee
```

On the PI
```
# ls -al /data
total 12
drwxr-xr-x    3 root     root          4096 Feb  1  2019 .
drwxr-xr-x   19 root     root          4096 Feb  1  2019 ..
drwxrwx---    3 tee      tee           4096 Feb  1  2019 tee
```
But now it's OK!

jbech-linaro:
Btw, throwing in one more patch that also updates the documentation (even though I didn't want to deal with that now .. but it's needed to make it somewhat accurate until we've have the new doc in place).

br-ext/package/optee_client/S30optee
```diff
@@ -19,6 +19,10 @@ case "$1" in
 			chown root:ion /dev/ion && chmod 0660 /dev/ion
 			[ $? = 0 ] && echo "OK" || echo "FAIL"
 		fi
+		printf "Create and set permissions on /data... "
+		mkdir -p /data/tee/logs && chown -R tee:tee /data/tee && \
+			chmod 0770 /data/tee
```
jforissier: TBH this looks hackish and duplicates https://github.com/OP-TEE/build/blob/master/br-ext/package/optee_client/optee_client.mk#L33. Moreover, only NFS seems to benefit from it (other platforms thn Rpi3 with NFS certainly work fine already). Perhaps this just hides some other problem? Or we could make a NFS-specific fix instead?
What is currently in `optee_client.mk` seems to be the proper way.

Is it necessary to create `logs`? What happens if it does not exists? I suppose logs are silently discarded. Perhaps `/data/tee/logs` should be added to `optee_client.mk` instead.

jforissier:
> Maybe, but optee_client.mk is also kind of a hack for Buildroot only.

Not a hack, it is properly documented... https://buildroot.org/downloads/manual/manual.html#customize-device-permission. The S30 script is for the Buildroot environment too. Therefore, I see no reason to duplicate things in the .mk and in the S30 script.

Worse case, we could set all permissions manually in S30 and discard the _PERMISSION section in `optee_client.mk`.

> I have no idea how to come up with another fix, my fixes are the S30 changes :) 

Well, that could be a good reason for doing things in S30, indeed :)

jbech-linaro:
> > Maybe, but optee_client.mk is also kind of a hack for Buildroot only.
> 
> Not a hack, it is properly documented... https://buildroot.org/downloads/manual/manual.html#customize-device-permission. The S30 script is for the Buildroot environment too. Therefore, I see no reason to duplicate things in the .mk and in the S30 script.

The optee_client.mk which is documented on the link you provided **is** a Buildroot only thing, but S30 start-up files in /etc/init.d/ is something that is not only for and used by Buildroot. That is something you see used in all sorts of Linux distros (all with different root fs). So I argue that putting stuff in init-scripts in /etc/init.d/ is more generic than having things in Buildroot specific files.

> 
> Worse case, we could set all permissions manually in S30 and discard the _PERMISSION section in `optee_client.mk`.

Sure, S30 only or S30 + `optee_client.mk`, doesn't matter to me. I'm fine with either of them. But for a S30 only thing I'd like to do that in a separate patch set, since it feels like that have bigger impact than my current changes have.

jforissier:
> That is something you see used in all sorts of Linux distros (all with different root fs)

"all with different root fs", that is my point. The startup scripts are likely to be tailored to the distro. We should not claim that *this* S30 file may be re-used outside of Buildroot. It should be adapted.
Anyway, it does not really matter ;-)

> But for a S30 only thing I'd like to do that in a separate patch set

I'm not convinced. I find one commit would be very clear:
```
Move creation of /data etc. from optee_client.mk to S30optee

For some reason Rpi3 with NFS root <blah blah describe the issue> and we
found a workaround which is to create the /data hierarchy and set its
permissions in S30optee rather than in the root fs definition in
optee_client.mk.
```

jbech-linaro:
> I'm not convinced. I find one commit would be very clear:

I've pushed another fix for this, tested it. Works as expected. If IBART and Travis are happy I guess we're good to go. Will squash together the commit messages when force pushing.

jforissier:
Thanks, LGTM now :)

jbech-linaro:

PR #343, opened 2019-02-05
optee service: update tee-supplicant install path

etienne-lms:
tee-supplicant is installed in /usr/bin instead of /usr/sbin.

This change depends on https://github.com/OP-TEE/optee_client/pull/142.

### DIFF
--- a/br-ext/package/optee_client/S30optee
+++ b/br-ext/package/optee_client/S30optee
@@ -6,7 +6,7 @@
 #
 case "$1" in
     start)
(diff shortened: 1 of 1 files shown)


jbech-linaro:
I think this has been debated in the past (I think I had it in `bin` initially). I'm fine with either one.

etienne-lms:
Rejected: see https://github.com/OP-TEE/optee_client/pull/142#issuecomment-460698917.

PR #344, opened 2019-02-05
[RFC] optee services: update script

etienne-lms:
Update script based on the feedback from Buildroot maintainer [1]:

>  > \+# /etc/init.d/optee
> 
>  Drop this comment, it is useless, and in fact wrong: the file will not
>  have this name in a Buildroot filesystem.
>
> (...)
> > \+	if [ -e /usr/sbin/tee-supplicant -a -e /dev/teepriv0 ]; then
>
> Drop this test, just start tee-supplicatn.
> (...)
> > \+		echo "Starting tee-supplicant..."
> > \+		/usr/sbin/tee-supplicant &
>
> Please use start-stop-daemon. See
> https://patchwork.ozlabs.org/patch/994013/ for the "right" way of
> writing an init script.

This maybe not be best way, if there's one, but it seems more
standard to use start-stop-daemon. Moreover the proposed script
ensures its return value matches the status of the caller request.

Note that I preserved the **status** case which Buildroot does not want.

[1] http://lists.busybox.net/pipermail/buildroot/2018-December/238354.html

### DIFF
--- a/br-ext/package/optee_client/S30optee
+++ b/br-ext/package/optee_client/S30optee
@@ -1,42 +1,82 @@
 #!/bin/sh
-#
-# /etc/init.d/optee
-#
-# Start/stop tee-supplicant (OP-TEE normal world daemon)
-#
-case "$1" in
-    start)
-	if [ -e /usr/sbin/tee-supplicant -a -e /dev/teepriv0 ]; then
-		# tee-supplicant and the client applications need not run as
-		# root provided that the TEE devices and the data store have
-		# proper permissions
-		printf "Setting permissions on /dev/tee*... "
-		chown root:tee /dev/teepriv0 && chmod 0660 /dev/teepriv0 && \
-			chown root:teeclnt /dev/tee0 && chmod 0660 /dev/tee0
-		[ $? = 0 ] && echo "OK" || echo "FAIL"
-		if [ -e /dev/ion ]; then
-			printf "Setting permissions on /dev/ion... "
-			chown root:ion /dev/ion && chmod 0660 /dev/ion
-			[ $? = 0 ] && echo "OK" || echo "FAIL"
-		fi
-		printf "Create and set permissions on /data... "
-		mkdir -p /data/tee && chown -R tee:tee /data/tee && \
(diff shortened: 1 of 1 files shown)


jenswi-linaro:
If Buildroot doesn't want **status**, why not remove it?

etienne-lms:
From some reading on the web (mainly https://refspecs.linuxfoundation.org/LSB_5.0.0/LSB-Core-generic/LSB-Core-generic/iniscrptact.html), it seemed `status` is part of the expected actions from an init script.

Thomas explicitly asked to not embed it in Buildroot but I though that our OP-TEE users may want this facility. We can simply remove it if we feel its is useless.

br-ext/package/optee_client/S30optee
```diff
@@ -1,42 +1,96 @@
...
+	mkdir -p /data/tee && chown -R tee:tee /data/tee && chmod 0770 /data/tee
+	status=$?
+	if [ "$status" -eq 0 ]; then
+		echo "OK"
+	else
+		echo "FAIL"
+		return "$status"
+	fi
+	printf 'Starting %s: ' "$DAEMON"
+	start-stop-daemon -S -q -p "$PIDFILE" -x "$DAEMON_PATH/$DAEMON" \
```
jenswi-linaro: `--chuid tee` missing?

etienne-lms:
> Since you're dealing with upstream I leave to **status** case to you.

I tend to be conservative, hence not removing it for now. But that's true that once op-tee will in be merged in BR and we sync with it, the **status** action will be removed. Does anyone use this? Maybe it's better to remove it now and get feedback if one gets annoyed with such change. "P'têt ben qu'oui, p'têt ben qu'non..." (maybe, maybe not)

jforissier:
> Does anyone use this?

I usually do "ps | grep supp" so I don't mind removing 'status'...

br-ext/package/optee_client/S30optee
etienne-lms: right.

jenswi-linaro:

etienne-lms:
Tag applied.

PR #345, opened 2019-02-06
common: remove deprecated rootfs targets

etienne-lms:
Targets `update_rootfs-common`, `update_rootfs-clean-common` and `filelist-tee-common` are not used by any of the supported platform and their content is not accurate regarding OP-TEE packages content (**init.d.optee**, **tee-supplicant** install path).

This change removes this make targets.

This change also remove the **init.d.optee** init script which is no more embedded in the filesystem of the supported platforms.


jbech-linaro:
It'd be nice to get rid of that, thanks!

jenswi-linaro:

etienne-lms:
Tags applied.

PR #350, opened 2019-02-25
Add support for Xilinx zcu102, zcu104 and zcu106 ZynqMP boards

mimok:
This commit adds support for Xilinx ZynqMP boards to the OPTEE/build repository.
The provided makefile relies on Xilinx Petalinux 2018.2 itself based on Yocto.
Petalinux and BSP files shall be downloaded from Xilinx website. You may have to create a Xilinx account.
The makefile helps to create, patch and build a customized Petalinux project with OP-TEE support.
Boards can be emulated using Petalinux's QEMU.
xtest test suite pass on Petalinux's QEMU emulating zcu102 board executing Petalinux v2018.2 and OP-TEE 3.4.0

### DIFF
--- a/zynqmp.mk
+++ b/zynqmp.mk
@@ -1,0 +1,63 @@
+ROOT			= $(PWD)/..
+PLATFORM		?= zcu102
+PETALINUX_PATH		?= /opt/Xilinx/petalinx_2018_2
+BSP_PATH		?= ./xilinx-${PLATFORM}-v2018.2-final.bsp
+PRJ_PATH		?= $(ROOT)/xilinx-${PLATFORM}-2018.2
+PETALINUX_CFG_PATH	?= $(ROOT)/build/zynqmp
+OPTEE_VER		?= latest
+
+define set_cfg
+	@sed -i 's/$(1)=.*/$(1)=$(2)/' $(3)
+endef
+
+define set_optee_version
+	@if [ "$(1)" != "latest" ]; then \
+		echo 'OPTEE_VERSION ?= "$(1)"' > $(2); \
+		echo 'SRCREV ?= "$(1)"' >> $(2); \
+	else \
+		echo 'OPTEE_VERSION ?= "latest"' > $(2); \
+		echo 'SRCREV ?= "$${AUTOREV}"' >> $(2); \
+	fi
+endef
+
+.PHONY: check-petalinux
+
+check-petalinux:
+ifndef PETALINUX_VER
+	$(error You have to source Petalinux settings)
+endif
+ifneq ($(PETALINUX_VER),2018.2)
+	$(error This makefile only support Petalinux 2018.2)
+endif
+
+petalinux-create: check-petalinux
+	@cd $(ROOT) && petalinux-create -t project -s $(BSP_PATH)
+	@cd $(PRJ_PATH) && petalinux-create -t apps --template install -n optee-client --enable
+	@cd $(PRJ_PATH) && petalinux-create -t apps --template install -n optee-test --enable
+	@#
+	$(call set_cfg,CONFIG_SUBSYSTEM_ATF_COMPILE_EXTRA_SETTINGS,"SPD=opteed ZYNQMP_BL32_MEM_BASE=0x60000000 ZYNQMP_BL32_MEM_SIZE=0x80000",$(PRJ_PATH)/project-spec/configs/config)
+	$(call set_cfg,CONFIG_SUBSYSTEM_ZYNQMP_ATF_MEM_SIZE,0x16001,$(PRJ_PATH)/project-spec/configs/config)
+	@#
+	@cp $(PETALINUX_CFG_PATH)/kernel_optee.cfg $(PRJ_PATH)/project-spec/meta-user/recipes-kernel/linux/linux-xlnx/
+	@cp $(PETALINUX_CFG_PATH)/linux-xlnx_%.bbappend $(PRJ_PATH)/project-spec/meta-user/recipes-kernel/linux/linux-xlnx_%.bbappend
+	@cp $(PETALINUX_CFG_PATH)/system-user.dtsi $(PRJ_PATH)/project-spec/meta-user/recipes-bsp/device-tree/files/
+	@#
+	@mkdir -p $(PRJ_PATH)/project-spec/meta-user/recipes-bsp/arm-trusted-firmware/
+	@cp -r $(PETALINUX_CFG_PATH)/arm-trusted-firmware/* $(PRJ_PATH)/project-spec/meta-user/recipes-bsp/arm-trusted-firmware/
+	@#
+	@mkdir -p $(PRJ_PATH)/project-spec/meta-user/recipes-bsp/optee-os
+	@cp -r $(PETALINUX_CFG_PATH)/optee-os/* $(PRJ_PATH)/project-spec/meta-user/recipes-bsp/optee-os/
+	@cp -r $(PETALINUX_CFG_PATH)/optee-client/* $(PRJ_PATH)/project-spec/meta-user/recipes-apps/optee-client/
+	@cp -r $(PETALINUX_CFG_PATH)/optee-test/* $(PRJ_PATH)/project-spec/meta-user/recipes-apps/optee-test/
+	
+petalinux-config:
+	$(call set_optee_version,$(OPTEE_VER),$(PRJ_PATH)/project-spec/meta-user/recipes-apps/optee-test/optee-test.bbappend)
+	$(call set_optee_version,$(OPTEE_VER),$(PRJ_PATH)/project-spec/meta-user/recipes-apps/optee-client/optee-client.bbappend)
+	$(call set_optee_version,$(OPTEE_VER),$(PRJ_PATH)/project-spec/meta-user/recipes-bsp/optee-os/optee-os.bbappend)
+	@cd $(PRJ_PATH) && petalinux-config --oldconfig
+
+petalinux-build:
+	@cd $(PRJ_PATH) && petalinux-build
+	
+qemu:
+	@cd $(PRJ_PATH) && petalinux-boot --qemu --qemu-args "-device loader,file=${PRJ_PATH}/images/linux/bl32.elf" --kernel
--- a/zynqmp/arm-trusted-firmware/arm-trusted-firmware_2018.2.bb
+++ b/zynqmp/arm-trusted-firmware/arm-trusted-firmware_2018.2.bb
@@ -1,0 +1,71 @@
+ATF_VERSION = "2.0"
+SRCREV ??= "dbc8d9496ead9ecdd7c2a276b542a4fbbbf64027"
+BRANCH ??= "master"
+DESCRIPTION = "Trusted Firmware-A"
+
+LICENSE = "BSD"
+LIC_FILES_CHKSUM = "file://license.rst;md5=e927e02bca647e14efd87e9e914b2443"
+
+PROVIDES = "virtual/arm-trusted-firmware"
+
+inherit deploy
+
+DEPENDS += "u-boot-mkimage-native"
+
+S = "${WORKDIR}/git"
+B = "${WORKDIR}/build"
+
+ATF_VERSION_EXTENSION ?= "-arm"
+PV = "${ATF_VERSION}${ATF_VERSION_EXTENSION}+git${SRCPV}"
+
+BRANCH ??= ""
+REPO ??= "git://github.com/ARM-software/arm-trusted-firmware.git;protocol=https"
+BRANCHARG = "${@['nobranch=1', 'branch=${BRANCH}'][d.getVar('BRANCH', True) != '']}"
+SRC_URI = "${REPO};${BRANCHARG}"
+
+ATF_BASE_NAME ?= "${PN}-${PKGE}-${PKGV}-${PKGR}-${DATETIME}"
+ATF_BASE_NAME[vardepsexclude] = "DATETIME"
+
+COMPATIBLE_MACHINE = "zynqmp"
+PLATFORM_zynqmp = "zynqmp"
+
+# requires CROSS_COMPILE set by hand as there is no configure script
+export CROSS_COMPILE="${TARGET_PREFIX}"
+
+# Let the Makefile handle setting up the CFLAGS and LDFLAGS as it is a standalone application
+CFLAGS[unexport] = "1"
+LDFLAGS[unexport] = "1"
+AS[unexport] = "1"
+LD[unexport] = "1"
+
+ATF_CONSOLE ?= ""
+ATF_CONSOLE_zynqmp = "cadence"
+
+DEBUG ?= ""
+EXTRA_OEMAKE_zynqmp_append = "${@' ZYNQMP_CONSOLE=${ATF_CONSOLE}' if d.getVar('ATF_CONSOLE', True) != '' else ''}"
+EXTRA_OEMAKE_append = " ${@bb.utils.contains('DEBUG', '1', ' DEBUG=${DEBUG}', '', d)}"
+
+OUTPUT_DIR = "${@bb.utils.contains('DEBUG', '1', '${B}/${PLATFORM}/debug', '${B}/${PLATFORM}/release', d)}"
+
+ATF_MEM_BASE ?= ""
+ATF_MEM_SIZE ?= ""
+
+EXTRA_OEMAKE_zynqmp_append = "${@' ZYNQMP_ATF_MEM_BASE=${ATF_MEM_BASE}' if d.getVar('ATF_MEM_BASE', True) != '' else ''}"
+EXTRA_OEMAKE_zynqmp_append = "${@' ZYNQMP_ATF_MEM_SIZE=${ATF_MEM_SIZE}' if d.getVar('ATF_MEM_SIZE', True) != '' else ''}"
+
+do_configure() {
+	oe_runmake clean -C ${S} BUILD_BASE=${B} PLAT=${PLATFORM}
+}
+
+do_compile() {
+	oe_runmake -C ${S} BUILD_BASE=${B} PLAT=${PLATFORM} RESET_TO_BL31=1 bl31
+}
+
+do_deploy() {
+	install -d ${DEPLOYDIR}
+	install -m 0644 ${OUTPUT_DIR}/bl31/bl31.elf ${DEPLOYDIR}/${ATF_BASE_NAME}.elf
+	ln -sf ${ATF_BASE_NAME}.elf ${DEPLOYDIR}/${PN}.elf
+	install -m 0644 ${OUTPUT_DIR}/bl31.bin ${DEPLOYDIR}/${ATF_BASE_NAME}.bin
+	ln -sf ${ATF_BASE_NAME}.bin ${DEPLOYDIR}/${PN}.bin
+}
+addtask deploy before do_build after do_compile
--- a/zynqmp/optee-client/optee-client.bb
+++ b/zynqmp/optee-client/optee-client.bb
@@ -1,0 +1,36 @@
+OPTEE_VERSION ??= "latest"
+SRCREV ??= "${AUTOREV}"
+BRANCH ??= "master"
+
+DESCRIPTION = "OP-TEE Client"
+
+LICENSE = "BSD"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=69663ab153298557a59c67a60a743e5b"
+
+PROVIDES = "virtual/optee-client"
+DEPENDS += "virtual/optee-os"
+
+S = "${WORKDIR}/git"
+PV = "${OPTEE_VERSION}+git${SRCPV}"
+
+REPO ??= "git://github.com/OP-TEE/optee_client.git;protocol=https"
+SRC_URI = "${REPO};branch=${BRANCH}"
+
+# requires CROSS_COMPILE set by hand as there is no configure script
+export CROSS_COMPILE="${TARGET_PREFIX}"
+
+EXPORT_DIR = "${TMPDIR}/deploy/images/${MACHINE}/optee/export_client"
+
+EXTRA_OEMAKE_append = " CROSS_COMPILE=${CROSS_COMPILE}"
+EXTRA_OEMAKE_append = " DESTDIR=${EXPORT_DIR}"
+EXTRA_OEMAKE_append = " SBINDIR=/sbin"
+EXTRA_OEMAKE_append = " LIBDIR=/lib"
+EXTRA_OEMAKE_append = " INCLUDEDIR=/include"
+
+do_install() {
+	oe_runmake install
+	install -d ${D}${libdir}
+	install -m 0644 ${EXPORT_DIR}/lib/libteec.so.1.0.0 ${D}${libdir}
+	install -d ${D}${sbindir}
+	install -m 0744 ${EXPORT_DIR}/sbin/tee-supplicant ${D}${sbindir}
+}
--- a/zynqmp/optee-os/optee-os.bb
+++ b/zynqmp/optee-os/optee-os.bb
@@ -1,0 +1,68 @@
+OPTEE_VERSION ??= "latest"
+SRCREV ??= "${AUTOREV}"
+BRANCH ??= "master"
+
+DESCRIPTION = "OP-TEE OS"
+
+# Define as closed license to prevent MD5 checksum verification since
+# LICENSE file changed around 3.5.0 making this recipe less flexible.
+# LICENSE = "BSD"
+# ante-3.5.0: LIC_FILES_CHKSUM = "file://LICENSE;md5=69663ab153298557a59c67a60a743e5b"
+# post-3.5.0: LIC_FILES_CHKSUM = "file://LICENSE;md5=c1f21c4f72f372ef38a5a4aee55ec173"
+LICENSE = "CLOSED"
+
+inherit deploy
+
+PROVIDES = "virtual/optee-os"
+DEPENDS += "u-boot-mkimage-native"
+
+S = "${WORKDIR}/git"
+PV = "${OPTEE_VERSION}+git${SRCPV}"
+
+REPO ??= "git://github.com/OP-TEE/optee_os.git;protocol=https"
+SRC_URI = "${REPO};branch=${BRANCH}"
+
+OPTEE_BASE_NAME ?= "${PN}-${PKGE}-${PKGV}-${PKGR}-${DATETIME}"
+OPTEE_BASE_NAME[vardepsexclude] = "DATETIME"
+
+# requires CROSS_COMPILE set by hand as there is no configure script
+export CROSS_COMPILE="${TARGET_PREFIX}"
+
+# Let the Makefile handle setting up the CFLAGS and LDFLAGS as it is a standalone application
+CFLAGS[unexport] = "1"
+LDFLAGS[unexport] = "1"
+AS[unexport] = "1"
+LD[unexport] = "1"
+
+DEBUG ??= "0"
+TA_DEV_KIT_DIR = "${TMPDIR}/deploy/images/${MACHINE}/optee/export-ta_arm64"
+OUTPUT_DIR = "${S}/out/arm-plat-zynqmp"
+TEE_LOG_LEVEL = "${@bb.utils.contains('DEBUG', '1', '3', '2', d)}"
+TEE_CORE_DEBUG = "${@bb.utils.contains('DEBUG', '1', 'y', 'n', d)}"
+
+EXTRA_OEMAKE_append = " comp-cflagscore=--sysroot=${STAGING_DIR_HOST}"
+EXTRA_OEMAKE_append = " CROSS_COMPILE=${CROSS_COMPILE}"
+EXTRA_OEMAKE_append = " CROSS_COMPILE_core=${CROSS_COMPILE}"
+EXTRA_OEMAKE_append = " CROSS_COMPILE_ta_arm64=${CROSS_COMPILE}"
+EXTRA_OEMAKE_append = " PLATFORM=zynqmp"
+EXTRA_OEMAKE_append = " CFG_ARM64_core=y"
+EXTRA_OEMAKE_append = " CFG_ARM32_core=n"
+EXTRA_OEMAKE_append = " CFG_USER_TA_TARGETS=ta_arm64"
+EXTRA_OEMAKE_append = " CFG_TEE_CORE_LOG_LEVEL=${TEE_LOG_LEVEL}"
+EXTRA_OEMAKE_append = " CFG_TEE_CORE_DEBUG=${TEE_CORE_DEBUG}"
+EXTRA_OEMAKE_append = " DEBUG=${DEBUG}"
+
+do_install() {
+	install -d ${TA_DEV_KIT_DIR}
+	cp -aR ${OUTPUT_DIR}/export-ta_arm64/* ${TA_DEV_KIT_DIR}
+}
+
+do_deploy() {
+	install -d ${DEPLOYDIR}
+	install -d ${TMPDIR}/../../images/linux/
+	install -m 0644 ${OUTPUT_DIR}/core/tee.elf ${DEPLOYDIR}/${OPTEE_BASE_NAME}.elf
+	install -m 0644 ${OUTPUT_DIR}/core/tee.elf ${TMPDIR}/../../images/linux/bl32.elf
+	install -m 0644 ${OUTPUT_DIR}/core/tee.bin ${DEPLOYDIR}/${OPTEE_BASE_NAME}.bin
+	install -m 0644 ${OUTPUT_DIR}/core/tee.bin ${TMPDIR}/../../images/linux/bl32.bin
+}
+addtask deploy before do_build after do_install
--- a/zynqmp/optee-test/optee-test.bb
+++ b/zynqmp/optee-test/optee-test.bb
@@ -1,0 +1,40 @@
+OPTEE_VERSION ??= "latest"
+SRCREV ??= "${AUTOREV}"
+BRANCH ??= "master"
+
+DESCRIPTION = "OP-TEE Test"
+
+LICENSE = "BSD"
+LIC_FILES_CHKSUM = "file://LICENSE.md;md5=daa2bcccc666345ab8940aab1315a4fa"
+
+PROVIDES = "virtual/optee-test"
+DEPENDS += "virtual/optee-os virtual/optee-client"
+
+S = "${WORKDIR}/git"
+PV = "${OPTEE_VERSION}+git${SRCPV}"
+
+REPO ??= "git://github.com/OP-TEE/optee_test.git;protocol=git"
+SRC_URI = "${REPO};branch=${BRANCH}"
+
+# requires CROSS_COMPILE set by hand as there is no configure script
+export CROSS_COMPILE="${TARGET_PREFIX}"
+
+TA_DEV_KIT_DIR = "${TMPDIR}/deploy/images/${MACHINE}/optee/export-ta_arm64"
+OPTEE_CLIENT_EXPORT = "${TMPDIR}/deploy/images/${MACHINE}/optee/export_client"
+
+EXTRA_OEMAKE_append = " CROSS_COMPILE=${CROSS_COMPILE}"
+EXTRA_OEMAKE_append = " CROSS_COMPILE_TA=${CROSS_COMPILE}"
+EXTRA_OEMAKE_append = " TA_DEV_KIT_DIR=${TA_DEV_KIT_DIR}"
+EXTRA_OEMAKE_append = " OPTEE_CLIENT_EXPORT=${OPTEE_CLIENT_EXPORT}"
+EXTRA_OEMAKE_append = " COMPILE_NS_USER=64"
+
+do_install() {
+	install -d ${D}/lib/optee_armtz
+	find ${S}/out/ta -type f -iname '*.ta' -exec install -m 0644 {} ${D}/lib/optee_armtz/ \;
+	install -d ${D}${bindir}
+	install -m 0744 ${S}/out/xtest/xtest ${D}${bindir}
+}
+
+FILES_${PN} = "/lib/optee_armtz"
+FILES_${PN} += "/lib"
+FILES_${PN} += "${bindir}"
--- a/zynqmp/system-user.dtsi
+++ b/zynqmp/system-user.dtsi
@@ -1,0 +1,9 @@
+/include/ "system-conf.dtsi"
+/ {
+	firmware {
+		optee {
+			compatible = "linaro,optee-tz";
+			method = "smc";
+		};
+	};
+};
--- a/zynqmp/kernel_optee.cfg
+++ b/zynqmp/kernel_optee.cfg
@@ -1,0 +1,6 @@
+CONFIG_TEE=y
+
+#
+# TEE drivers
+#
+CONFIG_OPTEE=y
--- a/zynqmp/linux-xlnx_%.bbappend
+++ b/zynqmp/linux-xlnx_%.bbappend
@@ -1,0 +1,5 @@
+SRC_URI += "file://bsp.cfg \
+            "
+SRC_URI_append += "file://kernel_optee.cfg"
+
+FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"


zynqmp.mk
```diff
@@ -0,0 +1,34 @@
+ROOT = $(PWD)/..
+
+PLATFORM ?= zcu102
```
jbech-linaro: Either indent all so they line up or don't indent at all.

zynqmp/arm-trusted-firmware/arm-trusted-firmware.inc
```diff
@@ -0,0 +1,73 @@
+DESCRIPTION = "ARM Trusted Firmware"
```
jbech-linaro: The project have changed name, It's "Trusted Firmware A" nowadays. So where possible I would write that instead and likewise for abbreviations, I would also change those where possible, i.e., from `ATF` to `TF-A` (or `TFA`).

zynqmp.mk
```diff
@@ -0,0 +1,34 @@
...
+	
+petalinux-config:
+	cd $(PRJ_PATH) && petalinux-config --oldconfig
+
+petalinux-build:
+	cd $(PRJ_PATH) && petalinux-build
+	
+qemu:
+	cd $(PRJ_PATH) && petalinux-boot --qemu --qemu-args "-device loader,file=${PRJ_PATH}/images/linux/bl32.elf" --kernel
+
```
jbech-linaro: Remove empty lines  here at the end.

zynqmp/arm-trusted-firmware/arm-trusted-firmware.inc
```diff
@@ -0,0 +1,73 @@
...
+BRANCHARG = "${@['nobranch=1', 'branch=${BRANCH}'][d.getVar('BRANCH', True) != '']}"
+SRC_URI = "${REPO};${BRANCHARG}"
+
+ATF_BASE_NAME ?= "${PN}-${PKGE}-${PKGV}-${PKGR}-${DATETIME}"
+ATF_BASE_NAME[vardepsexclude] = "DATETIME"
+
+COMPATIBLE_MACHINE = "zynqmp"
+PLATFORM_zynqmp = "zynqmp"
+
+
```
jbech-linaro: Remove double empty lines (single is enough).

zynqmp/config.patch
```diff
@@ -0,0 +1,14 @@
+--- config	2018-06-15 13:08:27.000000000 +0200
```
jbech-linaro: I'm a bit uneasy doing patching of files on the fly. What is the reason for this not being submitted directly to the upstream tree?

zynqmp/optee-client/optee-client.bb
```diff
@@ -0,0 +1,58 @@
+OPTEE_VERSION = "3.4.0"
+SRCREV ??= "08428734c67fb559e420d87fa52fd74a955ea1bd"
+BRANCH ??= "master"
+
+DESCRIPTION = "OPTEE Client"
```
jbech-linaro: s/OPTEE/OP-TEE/

zynqmp/optee-os/files/0001-allow-setting-sysroot-for-libgcc-lookup.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/mk/gcc.mk b/mk/gcc.mk
```
jbech-linaro: Same here, I'd like to avoid doing live patching if possible.

zynqmp/optee-os/optee-os.bb
```diff
@@ -0,0 +1,88 @@
+OPTEE_VERSION = "3.4.0"
+SRCREV ??= "406c609bbf08407c6493082bc1cf246465c484c6"
+BRANCH ??= "master"
+
+DESCRIPTION = "OPTEE OS"
```
jbech-linaro: s/OPTEE/OP-TEE/

zynqmp/optee-test/optee-test.bb
```diff
@@ -0,0 +1,66 @@
+OPTEE_VERSION = "3.4.0"
+SRCREV ??= "669058459e4a544be12f37dab103ee4c2b32e31d"
+BRANCH ??= "master"
+
+DESCRIPTION = "OPTEE Test"
```
jbech-linaro: s/OPTEE/OP-TEE/

zynqmp/optee-os/optee-os.bb
```diff
@@ -0,0 +1,88 @@
...
+           file://0001-allow-setting-sysroot-for-libgcc-lookup.patch \
+	"
+
+OPTEE_BASE_NAME ?= "${PN}-${PKGE}-${PKGV}-${PKGR}-${DATETIME}"
+OPTEE_BASE_NAME[vardepsexclude] = "DATETIME"
+
+COMPATIBLE_MACHINE = "zynqmp"
+PLATFORM_zynqmp = "zynqmp"
+
+
```
jbech-linaro: Remove double empty lines.

zynqmp/optee-os/optee-os.bb
```diff
@@ -0,0 +1,88 @@
...
+# requires CROSS_COMPILE set by hand as there is no configure script
+export CROSS_COMPILE="${TARGET_PREFIX}"
+
+# Let the Makefile handle setting up the CFLAGS and LDFLAGS as it is a standalone application
+CFLAGS[unexport] = "1"
+LDFLAGS[unexport] = "1"
+AS[unexport] = "1"
+LD[unexport] = "1"
+
+
```
jbech-linaro: Remove double empty lines.

zynqmp/optee-os/optee-os.bb
```diff
@@ -0,0 +1,88 @@
...
+EXTRA_OEMAKE_append = " LIBGCC_LOCATE_CFLAGS=--sysroot=${STAGING_DIR_HOST}"
+EXTRA_OEMAKE_append = " O=${OUTPUT_DIR}/out/arm"
+EXTRA_OEMAKE_append = " CROSS_COMPILE=${CROSS_COMPILE}"
+EXTRA_OEMAKE_append = " CROSS_COMPILE_core=${CROSS_COMPILE}"
+EXTRA_OEMAKE_append = " CROSS_COMPILE_ta_arm64=${CROSS_COMPILE}"
+EXTRA_OEMAKE_append = " ta-targets=ta_arm64 CFG_TEE_CORE_LOG_LEVEL=3 CFG_TEE_BENCHMARK=n PLATFORM=zynqmp CFG_ARM64_core=y CFG_ARM32_core=n CFG_PM_DEBUG=0 CFG_DT=y CFG_DT_ADDR=0x1407f000"
+
+
+
```
jbech-linaro: Remove double/triple empty lines.

zynqmp/optee-test/optee-test.bb
```diff
@@ -0,0 +1,66 @@
...
+S = "${WORKDIR}/git"
+B = "${WORKDIR}/build"
+
+PV = "${OPTEE_VERSION}+git${SRCPV}"
+
+BRANCH ??= ""
+REPO ??= "git://github.com/OP-TEE/optee_test.git;protocol=https"
+BRANCHARG = "${@['nobranch=1', 'branch=${BRANCH}'][d.getVar('BRANCH', True) != '']}"
+SRC_URI = "${REPO};${BRANCHARG} \
+	"
```
jbech-linaro: Why is the `"` on a separate line here?

zynqmp/optee-test/optee-test.bb
```diff
@@ -0,0 +1,66 @@
...
+
+OPTEE_BASE_NAME ?= "${PN}-${PKGE}-${PKGV}-${PKGR}-${DATETIME}"
+OPTEE_BASE_NAME[vardepsexclude] = "DATETIME"
+
+COMPATIBLE_MACHINE = "zynqmp"
+PLATFORM_zynqmp = "zynqmp"
+
+
```
jbech-linaro: Remove double empty lines.

mimok:
@jbech-linaro thank you for your comments
I will try to answer each question:

>     * This is totally different to the other devices who are using the same scheme (repo + manifest + build to set it all up). Maybe we should think in generic terms for future devices/setup's that are also based on OE/bb.

Actually, my first attempt was to follow the standard OP-TEE workflow. However, it ended up with kernel inexplicably hanging during boot process. Most of the Xilinx documentation on Linux from the Atlassian wiki appear to be outdated and it was therefore difficult to solve the problem. Then i tried to use a pure Yocto workflow but once again the Xilinx meta-layers were poorly documented (and i'm a proficient in BitBake and Yocto). Finally, the easiest solution was to configure a Petalinux project to add OP-TEE support. 

This solution heavily relies on Petalinux and therefore can't be generalized. However, i think that a skilled OE/bb developer might develop a pure Yocto workflow.
 
>     * bb-files etc, I wonder isn't there any metalayer upstream where those could reside instead and just use them from there? However, my lack of knowledge about "bb"/OE might make this comment invalid.

Actually there is some support from OP-TEE in OE but it seems to be outdated (v2.4.0). We can't use meta layers from Xilinx out of the box as they use outdated version of some piece of software (BL32 was not really supported by Xilinx on older version of ATF/TFA). As said above, a skilled OE/bb developer might do that with a fair amount of work. Anyway we cant remove the use of xilinx tools as they are needed to build some piece of software (fsbl, pmu firmware, boot.bin file, ...).

>     * Live patching, I would prefer send proper patches to the corresponding gits/project instead of doing runtime/live patching when configuring/building.

Ok, i could remove patching of OP-TEE OS by proposing a pull request. The other patch targeting ``config`` file is not really a patch. Petalinux does not provide a mean to override the parameters defined in this file using command line. Only the GUI (menuconfig) can be used. Each BSP file come with a customized config file (but the structure is the same), therefore the simplest way to modify files from different BSP is to use the patch trick. otherwise i would have to provide a config file for each board. 

>     * All this would work magically if you try to follow what you have written in [OP-TEE/optee_docs#7](https://github.com/OP-TEE/optee_docs/pull/7)? We have a 96boards ZynqMP (https://www.96boards.org/product/ultra96/), I suppose we could try this out using that?

It works out of the box for zcu102, zcu104 and zcu106 QEMU emulation. I'm pretty sure that it would work on real targets but i don't own any of these boards. To use this code with an ultra96 board we would have to :

- update optee_os repository to add a new platform flavour with only 2GB of memory. 
- update recipes to take into account this new flavor.

Maybe it would be necessary to modify others things but i'm not sure.
Actually i have a Ultra96 board too and the next step is to update my code to use it with this board.

I will take into account your comments ASAP.

zynqmp/arm-trusted-firmware/arm-trusted-firmware.inc
mimok: Don't want to change variable names as Petalinux defines some of them. project name updated.

zynqmp/config.patch
mimok: Not really a patch. I would have to ship a config file for each supported board without this trick. Indeed, config have the same structure but different contents.

zynqmp/optee-os/files/0001-allow-setting-sysroot-for-libgcc-lookup.patch
mimok: patch removed, workaround provided by @etienne-lms

zynqmp/optee-test/optee-test.bb
mimok: Seems to be required by bb

mimok:
All comments taken into account

zynqmp/config.patch
mimok: Patch replaced by use of sed command

jbech-linaro:
@etienne-lms , are you happy with this PR as it stands right now. If yes, then I think we should go a head and merge that and the corresponding documentation PR.

zynqmp/optee-test/optee-test.bb
```diff
@@ -0,0 +1,62 @@
...
+TA_DEV_KIT_DIR = "${TMPDIR}/deploy/images/${MACHINE_NAME}/optee/export-user_ta"
+OPTEE_CLIENT_EXPORT = "${TMPDIR}/deploy/images/${MACHINE_NAME}/optee/export"
+
+EXTRA_OEMAKE_append = " CROSS_COMPILE=${CROSS_COMPILE}"
+EXTRA_OEMAKE_append = " CROSS_COMPILE_TA=${CROSS_COMPILE}"
+EXTRA_OEMAKE_append = " TA_DEV_KIT_DIR=${TA_DEV_KIT_DIR}"
+EXTRA_OEMAKE_append = " OPTEE_CLIENT_EXPORT=${OPTEE_CLIENT_EXPORT}"
+TARGET_CFLAGS_append = " -I${OPTEE_CLIENT_EXPORT}/include"
+
```
etienne-lms: Recommendation: set `EXTRA_OEMAKE_append = " COMPILE_NS_USER=64 "`.

zynqmp/optee-os/optee-os.bb
```diff
@@ -0,0 +1,88 @@
...
+EXTRA_OEMAKE_append = " comp-cflagscore=--sysroot=${STAGING_DIR_HOST}"
+EXTRA_OEMAKE_append = " O=${OUTPUT_DIR}"
+EXTRA_OEMAKE_append = " CROSS_COMPILE=${CROSS_COMPILE}"
+EXTRA_OEMAKE_append = " CROSS_COMPILE_core=${CROSS_COMPILE}"
+EXTRA_OEMAKE_append = " CROSS_COMPILE_ta_arm64=${CROSS_COMPILE}"
+EXTRA_OEMAKE_append = " PLATFORM=zynqmp"
+EXTRA_OEMAKE_append = " CFG_ARM64_core=y CFG_ARM32_core=n"
+EXTRA_OEMAKE_append = " CFG_USER_TA_TARGETS=ta_arm64"
+EXTRA_OEMAKE_append = " CFG_TEE_BENCHMARK=n"
+EXTRA_OEMAKE_append = " CFG_TEE_CORE_LOG_LEVEL=3 CFG_PM_DEBUG=0"
```
etienne-lms: Suggestion: maybe you would like `CFG_TEE_CORE_DEBUG` be set here.

etienne-lms:
Only 1 correction that should be fixed IMO: see https://github.com/OP-TEE/build/pull/350#discussion_r261960389 about empty lines in make target script sequences. 
Otherwise than that, LGTM, `Acked-by: Etienne Carriere <etienne.carriere@linaro.org>`.

zynqmp/optee-os/optee-os.bb
mimok: `CFG_TEE_CORE_DEBUG=n` added but `CFG_PM_DEBUG` removed because it is deprecated.

zynqmp.mk
```diff
@@ -0,0 +1,40 @@
...
+BSP_PATH		?= ./xilinx-${PLATFORM}-v2018.2-final.bsp
+PRJ_PATH		?= $(ROOT)/xilinx-${PLATFORM}-2018.2
+PETALINUX_CFG_PATH	?= $(ROOT)/build/zynqmp
+
+define set_cfg
+    sed -i 's/$(1)=.*/$(1)=$(2)/' $(3)
+endef
+
+petalinux-create:
+	echo "You have to source settings.sh from Petalinux directory!" #source $(PETALINUX_PATH)/settings.sh
```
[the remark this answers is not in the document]
mimok: two checks are added to verify the petalinux settings are actually sourced and petalinux version is equal to 2018.2

zynqmp.mk
```diff
@@ -0,0 +1,40 @@
...
+	cd $(ROOT) && petalinux-create -t project -s $(BSP_PATH)
+	cd $(PRJ_PATH) && petalinux-create -t apps --template install -n optee-client --enable
+	cd $(PRJ_PATH) && petalinux-create -t apps --template install -n optee-test --enable
+
```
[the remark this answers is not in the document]
mimok: `#` added but without `<tab>` to prevent this character to be displayed in console

zynqmp/optee-test/optee-test.bb
mimok: ok

mimok:
i will force push the squashed commit tomorrow evening once i have reviewed my modifications by rebuilding the whole project

mimok:
All comments taken into account and commits squashed

zynqmp.mk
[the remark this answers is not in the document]
etienne-lms: can use @#

zynqmp/optee-client/optee-client.bb
```diff
@@ -0,0 +1,57 @@
...
+}
+
+do_install() {
+	oe_runmake -C ${S} install
+	install -d ${D}${libdir}
+	install -m 0644 ${OUTPUT_DIR}/libteec/libteec.so.1.0.0 ${D}${libdir}
+	install -d ${D}${bindir}
+	install -m 0744 ${OUTPUT_DIR}/tee-supplicant/tee-supplicant ${D}${bindir}
+	#install -d ${TMPDIR}/deploy/images/zcu102-zynqmp/optee/include
+	#install -m 0644 ${S}/public/*.h ${TMPDIR}/deploy/images/${MACHINE_NAME}/optee/include
```
etienne-lms: Why are those 2 commented?

zynqmp.mk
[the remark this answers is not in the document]
mimok: From what I understand using `<tab>@#` will trigger a shell interpreter while `#` alone will not do that. I propose to leave this as it is. Are you agree ?

zynqmp/optee-client/optee-client.bb
mimok: These lines can be safely removed. I will update my PR in this way.

mimok:
I followed the xilinx bitbake template which fetches a given version of a git repository. I assume this is done to prevent backward compatibility issues.

However, it's possible to fetch the last version of master using a special value for `SRCREV`. I can modify the PR in this way but I never used this feature and therefore I have to test the whole thing again. I'm not sure I have time to do that before Monday night.

zynqmp.mk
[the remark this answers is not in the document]
etienne-lms: No, please fix. The indentation is weird.
As already used by our dear maintainer `<tab>@#` nicely splits line blocks with nice indentation and no ugly trace.

jforissier:
Building a fixed version doesn't seem very useful to me.

jbech-linaro:
Maybe the eleventh hour to say this, so I apologize in advance. But what would it take to do this in the traditional way that we are doing for all other boards supported in build.git/manifest.git? I.e., just use least amount of components together with Buildroot? Not 100% sure, but I believe that is how the old Zync-MP implementation was put together.

mimok:
@etienne-lms and @jforisser: I'm agree with you. This is not really efficient as one would have to update the ZynqMP target each time a new OP-TEE release is available. I'm going to fix that by providing a way to either build a given version of OP-TEE or build the last one.

@jbech-linaro: Doing that is clearly possible but it would ends up with some kind of PoC not really useable together with the xilinx toolchain at production stage. IMO a better approach would be to provide a pure bitbake target but this is not my priority. However I understand your concern and I would understand if that PR was not merged into the official repository.

mimok:
Comments taken into account.
Tested on latest version of OP-TEE and release 3.4.0.
If you are ok then i will squash the last commit and force push the result.

zynqmp/optee-client/optee-client.bb
```diff
@@ -0,0 +1,57 @@
+OPTEE_VERSION ??= "latest"
+SRCREV ??= "${AUTOREV}"
+BRANCH ??= "master"
+
+DESCRIPTION = "OP-TEE Client"
+
+#defined as closed license to prevent MD5 checksum verification
+LICENSE = "CLOSED"
```
etienne-lms: this is false.
try something like this:
```
LICENSE = "BSD"
LIC_FILES_CHKSUM = "file://${S}/LICENSE;md5=69663ab153298557a59c67a60a743e5b"
```

zynqmp/optee-os/optee-os.bb
```diff
@@ -0,0 +1,87 @@
+OPTEE_VERSION ??= "latest"
+SRCREV ??= "${AUTOREV}"
+BRANCH ??= "master"
+
+DESCRIPTION = "OP-TEE OS"
+
+#defined as closed license to prevent MD5 checksum verification
```
etienne-lms: to fix...

zynqmp/optee-os/optee-os.bb
```diff
@@ -0,0 +1,87 @@
...
+OUTPUT_DIR = "${B}/${PLATFORM}/out/arm"
+TA_DEV_KIT_DIR = "${TMPDIR}/deploy/images/${MACHINE}/optee/export-user_ta"
+
+EXTRA_OEMAKE_append = " comp-cflagscore=--sysroot=${STAGING_DIR_HOST}"
+EXTRA_OEMAKE_append = " O=${OUTPUT_DIR}"
+EXTRA_OEMAKE_append = " CROSS_COMPILE=${CROSS_COMPILE}"
+EXTRA_OEMAKE_append = " CROSS_COMPILE_core=${CROSS_COMPILE}"
+EXTRA_OEMAKE_append = " CROSS_COMPILE_ta_arm64=${CROSS_COMPILE}"
+EXTRA_OEMAKE_append = " PLATFORM=zynqmp"
+EXTRA_OEMAKE_append = " CFG_ARM64_core=y CFG_ARM32_core=n"
```
etienne-lms: suggest to split over 2 lines for consistency.

zynqmp/optee-test/optee-test.bb
```diff
@@ -0,0 +1,61 @@
...
+do_install() {
+	install -d ${D}/lib/optee_armtz
+	find ${S}/out/ta -type f -iname '*.ta' -exec install -m 0644 {} ${D}/lib/optee_armtz/ \;
+	install -d ${D}${bindir}
+	install -m 0744 ${S}/out/xtest/xtest ${D}${bindir}
+}
+
+FILES_${PN} = "/lib/optee_armtz"
+FILES_${PN} += "/lib"
+FILES_${PN} += "/usr/bin"
```
etienne-lms: maybe s/`/usr/bin`/`${bindir}/

zynqmp/optee-client/optee-client.bb
```diff
@@ -0,0 +1,57 @@
...
+EXPORT_DIR = "${TMPDIR}/deploy/images/${MACHINE}/optee/export_client"
+
+EXTRA_OEMAKE_append = " CROSS_COMPILE=${CROSS_COMPILE}"
+EXTRA_OEMAKE_append = " DESTDIR=${EXPORT_DIR}"
+EXTRA_OEMAKE_append = " SBINDIR=/sbin"
+EXTRA_OEMAKE_append = " LIBDIR=/lib"
+EXTRA_OEMAKE_append = " INCLUDEDIR=/include"
+
+do_configure() {
+	oe_runmake clean -C ${S} clean
```
etienne-lms: minor: looks a strange configure way.

zynqmp/optee-client/optee-client.bb
```diff
@@ -0,0 +1,57 @@
...
+}
+
+do_compile() {
+	oe_runmake -C ${S} build
```
etienne-lms: minor: This looks like the default oe make way. Do you think you can remove this local `do_compile` definition (maybe the configure step also)?

zynqmp/optee-client/optee-client.bb
mimok: Agree with you but license file of one of the three repositories has changed since version 3.4.0. therefore to switch between the latest version of OP-TEE and an older one, one would have to update the md5 hash in BB recipe. This is quite annoying. As far as I know, the only way to disable license checksum verification is to set license variable to "closed". If this is a problem, I believe that the only solution is to remove the possibility to build a given version of OP-TEE.

zynqmp/optee-client/optee-client.bb
mimok: What do you mean?

etienne-lms: configure is for configuring. cleaning for cleaning.

etienne-lms: this may be nitpicking. feel free to not address my comment.

zynqmp/optee-client/optee-client.bb
etienne-lms: optee-client and optee-test `LICENSE` files are quite stable, only optee-os recently changed. Maybe use this `CLOSED` hack only for optee-os, with an explicit comment:
```
# Define as closed license to prevent MD5 checksum verification since
# LICENSE file changed around 3.5.0 make this recipe less flexible.
# LICENSE = "BSD"
# ante 3.5.0: LIC_FILES_CHKSUM = "file://${S}/LICENSE;md5=69663ab153298557a59c67a60a743e5b"
# post 3.5.0: LIC_FILES_CHKSUM = "file://${S}/LICENSE;md5=c1f21c4f72f372ef38a5a4aee55ec173"
LICENSE = "CLOSED"
```

zynqmp/optee-client/optee-client.bb
mimok: My bitbake recipes are based on TF-A recipe from xilinx-meta layer. Don't know why they did that, maybe its due to Petalinux. As i don't know what might happen if i change this, i prefer not to modify it.

zynqmp/optee-client/optee-client.bb
mimok: Actually i have no idea. My guess is no and i can't find anything about default do_compile or do_configure target in Yocto documentation. It is only said that this target are automatically called when a bitbake recipe is executed. If someone have more informartion can change it otherwise i will leave it as it is.

zynqmp/optee-os/optee-os.bb
mimok: ok

zynqmp/optee-test/optee-test.bb
mimok: ok

zynqmp/optee-client/optee-client.bb
jforissier: @mimok https://en.wikipedia.org/wiki/Cargo_cult_programming ;-)

mimok:
comments taken into account (see replies to comments).

ghost11235:
I will be going through and using this build configuration for an ultra96 board. It looks like the ultra96 flavor has been added to the os. Following from below:
> * All this would work magically if you try to follow what you have written in [OP-TEE/optee_docs#7](https://github.com/OP-TEE/optee_docs/pull/7)? We have a 96boards ZynqMP (https://www.96boards.org/product/ultra96/), I suppose we could try this out using that?

I should be able to set PLATFORM = ultra96 and make everything.
I will be testing this tomorrow. 
@mimok Has this been tested on an ultra96 board?

zynqmp/optee-client/optee-client.bb
mimok: Sure I'm not a BB guru. However I find out that the default behaviour of `do_configure` is to perform a `make clean`. I'm going to check if I can remove that as this seems to be the yocto default behaviour.

mimok:
@ghost11235: yes it works, I managed to boot an ultra96 board with optee running. However some modifications are needed. I cannot publish this work yet but main ideas are to find out a way to package the files using `petalinux-package`, disable some packages otherwise kernel will hang and finally if you want to have a working serial output (using UART/JTAG pod) you will have to use the latest hdf file for ultra96. You can find this file in xilinx's hdf repository on GitHub.

mimok:
@etienne-lms : i removed useless code from BB recipes

zynqmp/arm-trusted-firmware/arm-trusted-firmware_2018.2.bb
```diff
@@ -0,0 +1,75 @@
...
+OUTPUT_DIR = "${@bb.utils.contains('DEBUG', '1', '${B}/${PLATFORM}/debug', '${B}/${PLATFORM}/release', d)}"
+
+ATF_MEM_BASE ?= ""
+ATF_MEM_SIZE ?= ""
+
+EXTRA_OEMAKE_zynqmp_append = "${@' ZYNQMP_ATF_MEM_BASE=${ATF_MEM_BASE}' if d.getVar('ATF_MEM_BASE', True) != '' else ''}"
+EXTRA_OEMAKE_zynqmp_append = "${@' ZYNQMP_ATF_MEM_SIZE=${ATF_MEM_SIZE}' if d.getVar('ATF_MEM_SIZE', True) != '' else ''}"
+
+do_configure() {
+	oe_runmake clean -C ${S} BUILD_BASE=${B} PLAT=${PLATFORM}
```
etienne-lms: minor: `BUILD_BASE=${B} PLAT=${PLATFORM}` could be added to `EXTRA_OEMAKE` and `do_configure()` removed.

zynqmp/arm-trusted-firmware/arm-trusted-firmware_2018.2.bb
```diff
@@ -0,0 +1,75 @@
...
+}
+
+do_compile() {
+	oe_runmake -C ${S} BUILD_BASE=${B} PLAT=${PLATFORM} RESET_TO_BL31=1 bl31
+}
+
+do_install() {
+	:
```
etienne-lms: for info: this means nothing to install.

zynqmp/optee-os/optee-os.bb
```diff
@@ -0,0 +1,67 @@
+OPTEE_VERSION ??= "latest"
+SRCREV ??= "${AUTOREV}"
+BRANCH ??= "master"
+
+DESCRIPTION = "OP-TEE OS"
+
+# Define as closed license to prevent MD5 checksum verification since
+# LICENSE file changed around 3.5.0 making this recipe less flexible.
+# LICENSE = "BSD"
+# ante 3.5.0: LIC_FILES_CHKSUM = "file://LICENSE;md5=69663ab153298557a59c67a60a743e5b"
```
etienne-lms: Minor: check if `ante` was formal English. Looks ok [1] but with a hyphen:
```patch
 # LICENSE = "BSD"
-# ante 3.5.0: LIC_FILES_CHKSUM = "file://LICENSE;md5=69663ab153298557a59c67a60a743e5b"
+# ante-3.5.0: LIC_FILES_CHKSUM = "file://LICENSE;md5=69663ab153298557a59c67a60a743e5b"
 # post 3.5.0: LIC_FILES_CHKSUM = "file://LICENSE;md5=c1f21c4f72f372ef38a5a4aee55ec173"

File: zynqmp/system-user.dtsi [hunk 29]
```diff
@@ -0,0 +1,9 @@
+/include/ "system-conf.dtsi"
+/ {
+	firmware {
+		        optee {
```
etienne-lms: suggest to simplify indentation with only tabulations.

zynqmp/optee-os/optee-os.bb
```diff
@@ -0,0 +1,67 @@
...
+EXTRA_OEMAKE_append = " comp-cflagscore=--sysroot=${STAGING_DIR_HOST}"
+EXTRA_OEMAKE_append = " CROSS_COMPILE=${CROSS_COMPILE}"
+EXTRA_OEMAKE_append = " CROSS_COMPILE_core=${CROSS_COMPILE}"
+EXTRA_OEMAKE_append = " CROSS_COMPILE_ta_arm64=${CROSS_COMPILE}"
+EXTRA_OEMAKE_append = " PLATFORM=zynqmp"
+EXTRA_OEMAKE_append = " CFG_ARM64_core=y"
+EXTRA_OEMAKE_append = " CFG_ARM32_core=n"
+EXTRA_OEMAKE_append = " CFG_USER_TA_TARGETS=ta_arm64"
+EXTRA_OEMAKE_append = " CFG_TEE_BENCHMARK=n"
+EXTRA_OEMAKE_append = " CFG_TEE_CORE_LOG_LEVEL=1"
```
etienne-lms: suggest to use level 2. About same verbosity as error level, but useful info at boot time.

zynqmp/optee-os/optee-os.bb
```diff
@@ -0,0 +1,67 @@
...
+
+EXTRA_OEMAKE_append = " comp-cflagscore=--sysroot=${STAGING_DIR_HOST}"
+EXTRA_OEMAKE_append = " CROSS_COMPILE=${CROSS_COMPILE}"
+EXTRA_OEMAKE_append = " CROSS_COMPILE_core=${CROSS_COMPILE}"
+EXTRA_OEMAKE_append = " CROSS_COMPILE_ta_arm64=${CROSS_COMPILE}"
+EXTRA_OEMAKE_append = " PLATFORM=zynqmp"
+EXTRA_OEMAKE_append = " CFG_ARM64_core=y"
+EXTRA_OEMAKE_append = " CFG_ARM32_core=n"
+EXTRA_OEMAKE_append = " CFG_USER_TA_TARGETS=ta_arm64"
+EXTRA_OEMAKE_append = " CFG_TEE_BENCHMARK=n"
```
etienne-lms: this is not needed. Benchmark is not enabled by default in OP-TEE.

zynqmp/optee-os/optee-os.bb
```diff
@@ -0,0 +1,67 @@
...
+EXTRA_OEMAKE_append = " CFG_TEE_CORE_LOG_LEVEL=1"
+EXTRA_OEMAKE_append = " CFG_TEE_CORE_DEBUG=n"
+EXTRA_OEMAKE_append = " DEBUG=0"
```
etienne-lms: Maybe relate these debug info to `DEBUG` state, as in tf-a recipe.

zynqmp/arm-trusted-firmware/arm-trusted-firmware_2018.2.bb
mimok: Not so simple I think. By default oe_runmake is executed in ${B} therefore if I do that BB will not find the source code which is located in ${S}. 
The alternative is to compile TF-A directly in ${S} by not overriding ${B} which seems to be initially equal to ${S}.

zynqmp/optee-os/optee-os.bb
mimok: In TF-A `DEBUG` is useful because it can be set trough petalinux GUI which is not the case here. This why I removed line of code related to debug state. However I can add a variable to allow debug state overriding trough a bbappend file.

zynqmp/arm-trusted-firmware/arm-trusted-firmware_2018.2.bb
mimok: ok, removed as this is default behavior

zynqmp/arm-trusted-firmware/arm-trusted-firmware_2018.2.bb
mimok: i prefer not changing that as this recipe is almost the same as the one available in xilinx-meta layer. Actually, i will remove it once xilinx will use TF-A 2.0.

mimok:
all comments taken into account except the one about `do_configure` in TF-A recipe. Hope this commit will be the last one as i will be off for a week. If it is ok, i will squash and force push tomorrow

zynqmp/arm-trusted-firmware/arm-trusted-firmware_2018.2.bb
etienne-lms: fine

zynqmp/optee-os/optee-os.bb
```diff
@@ -0,0 +1,69 @@
...
+EXTRA_OEMAKE_append = " CFG_ARM32_core=n"
+EXTRA_OEMAKE_append = " CFG_USER_TA_TARGETS=ta_arm64"
+EXTRA_OEMAKE_append = " CFG_TEE_CORE_LOG_LEVEL=${TEE_LOG_LEVEL}"
+EXTRA_OEMAKE_append = " CFG_TEE_CORE_DEBUG=${TEE_CORE_DEBUG}"
+EXTRA_OEMAKE_append = " DEBUG=${DEBUG}"
+
+do_install() {
+    #install TA devkit
+    install -d ${TA_DEV_KIT_DIR}
+    cp -aR  ${OUTPUT_DIR}/export-ta_arm64/*  ${TA_DEV_KIT_DIR}
```
etienne-lms: use tabulations for indentation, for consistency with `do_deploy()`.
minor: 2 spaces:
```
    cp -aR  ${OUTPUT_DIR}/export-ta_arm64/*  ${TA_DEV_KIT_DIR}
          ^                                ^

File: zynqmp/optee-os/optee-os.bb [hunk 34]
```diff
@@ -0,0 +1,69 @@
...
+EXTRA_OEMAKE_append = " PLATFORM=zynqmp"
+EXTRA_OEMAKE_append = " CFG_ARM64_core=y"
+EXTRA_OEMAKE_append = " CFG_ARM32_core=n"
+EXTRA_OEMAKE_append = " CFG_USER_TA_TARGETS=ta_arm64"
+EXTRA_OEMAKE_append = " CFG_TEE_CORE_LOG_LEVEL=${TEE_LOG_LEVEL}"
+EXTRA_OEMAKE_append = " CFG_TEE_CORE_DEBUG=${TEE_CORE_DEBUG}"
+EXTRA_OEMAKE_append = " DEBUG=${DEBUG}"
+
+do_install() {
+    #install TA devkit
```
etienne-lms: remove the comment, commands are explicit.

zynqmp/optee-os/optee-os.bb
```diff
@@ -0,0 +1,69 @@
...
+    install -d ${TA_DEV_KIT_DIR}
+    cp -aR  ${OUTPUT_DIR}/export-ta_arm64/*  ${TA_DEV_KIT_DIR}
+}
+
+do_deploy() {
+	install -d ${DEPLOYDIR}
+	install -m 0644 ${OUTPUT_DIR}/core/tee.elf ${DEPLOYDIR}/${OPTEE_BASE_NAME}.elf
+	install -d ${TMPDIR}/../../images/linux/
```
etienne-lms: minor: can you swap lines 63 and 64, easier to read:
```
do_deploy() {
	install -d ${DEPLOYDIR}
	install -d ${TMPDIR}/../../images/linux/
	install -m 0644 ${OUTPUT_DIR}/core/tee.elf ${DEPLOYDIR}/${OPTEE_BASE_NAME}.elf
	install -m 0644 ${OUTPUT_DIR}/core/tee.elf ${TMPDIR}/../../images/linux/bl32.elf
	install -m 0644 ${OUTPUT_DIR}/core/tee.bin ${DEPLOYDIR}/${OPTEE_BASE_NAME}.bin
	install -m 0644 ${OUTPUT_DIR}/core/tee.bin ${TMPDIR}/../../images/linux/bl32.bin
}

File: zynqmp/optee-client/optee-client.bb [hunk 36]
```diff
@@ -0,0 +1,36 @@
...
+EXTRA_OEMAKE_append = " DESTDIR=${EXPORT_DIR}"
+EXTRA_OEMAKE_append = " SBINDIR=/sbin"
+EXTRA_OEMAKE_append = " LIBDIR=/lib"
+EXTRA_OEMAKE_append = " INCLUDEDIR=/include"
+
+do_install() {
+	oe_runmake install
+	install -d ${D}${libdir}
+	install -m 0644 ${EXPORT_DIR}/lib/libteec.so.1.0.0 ${D}${libdir}
+	install -d ${D}${bindir}
```
etienne-lms: suggest to add an explicit comment to not be surprised tee-supplicant is generated in sbin/ but installed in {bindir}.
`# Intentionally install supplicant from sbin to distribution ${bindir}`

zynqmp/optee-os/optee-os.bb
```diff
@@ -0,0 +1,69 @@
...
+SRCREV ??= "${AUTOREV}"
+BRANCH ??= "master"
+
+DESCRIPTION = "OP-TEE OS"
+
+# Define as closed license to prevent MD5 checksum verification since
+# LICENSE file changed around 3.5.0 making this recipe less flexible.
+# LICENSE = "BSD"
+# ante-3.5.0: LIC_FILES_CHKSUM = "file://LICENSE;md5=69663ab153298557a59c67a60a743e5b"
+# post 3.5.0: LIC_FILES_CHKSUM = "file://LICENSE;md5=c1f21c4f72f372ef38a5a4aee55ec173"
```
etienne-lms: while your at it, maybe also add an hyphen to `post-`.

zynqmp/optee-client/optee-client.bb
mimok: installed into `/usr/sbin` instead

mimok:
all comment taken into account. Squashed into a single commit.

jbech-linaro:
@mimok, we've created the 3.5.0-rc1 tag, which means that we usually only merge bug fixes until the final release tag has been created (Friday this week). I'd be happy to merge this now as such, but since this is more than a bug fix I would need a "Tested-by" tag from you in this pull request https://github.com/OP-TEE/optee_os/pull/2937. Is that something you can do before Friday this week? If yes, then I'll go ahead a merge this?

mimok:
@jbech-linaro : tested and commit amended

PR #351, opened 2019-03-06
Hikey960 kernel v5.0

jforissier:
One patch to support Linux kernel v5.0 on HiKey960, plus one patch to improve build time.


jforissier:
@jbech-linaro it is possible that HiKey needs the same fix as HiKey960, I'm not sure (commit "hikey960: add CONFIG_SCSI_UFS_HISI=y").

Would you mind checking? To do that, `cd linux`, checkout branch `wip/optee-v5.0` from `linaro-swg/linux.git` and do a clean build. If the board hangs before showing the login prompt, then we likely need `CONFIG_SCSI_UFS_HISI=y`.

jbech-linaro:
> Would you mind checking?

I can give it a try, I'll keep you posted.

jbech-linaro:
```bash
...
[    0.000000] Linux version 5.0.0-238943-g0fd2debf7f00 (jbech@debby) (gcc version 8.2.1 2018
...
# xtest
24049 subtests of which 0 failed
94 test cases of which 0 failed
0 test cases were skipped
TEE test application done!
```
:+1: 

And actually, IBART is also enabled for this git, so my checking here wasn't really necessary when thinking about it :) But now we have double checked that it seems OK for the old HiKey.

jbech-linaro:

PR #352, opened 2019-03-08
buildroot: add optee_os package to copy shared libraries into the root FS

jforissier:
When OP-TEE is built with CFG_ULIBS_SHARED=y, TA shared libraries are
created. They have to be installed into the target root FS in order
to be found at runtime.

This patch adds a custom package to do just that.


jforissier:
Ping?

jforissier:
Comment added to `CMakeLists.txt`. Thanks.

PR #353, opened 2019-03-20
[RFC] build optee-{test|examples} from CMake only

etienne-lms:
Remove the extra build/install commands for packages **optee_test** and **optee_examples** since these packages now build and install trusted applications from the CMake build process without requiring an extra commands.

This RFC depends on https://github.com/linaro-swg/optee_examples/pull/56 and https://github.com/OP-TEE/optee_test/pull/337.


etienne-lms:
rebased.

PR #358, opened 2019-04-10
Add support for Raspberry Pi 3 B plus

johnphilby:
Make use of the PLATFORM variable to set the specific
target being used "rpi3-b" or "rpi3-bplus".

Signed-off-by: Philby John <philby.j@hcl.com>
Signed-off-by: Vv Ramya <ramyavv@hcl.com>

### DIFF
--- a/rpi3.mk
+++ b/rpi3.mk
@@ -37,7 +37,8 @@ OPTEE_BIN_EXTRA1	?= $(OPTEE_PATH)/out/arm/core/tee-pager_v2.bin
 OPTEE_BIN_EXTRA2	?= $(OPTEE_PATH)/out/arm/core/tee-pageable_v2.bin
 
 LINUX_IMAGE		?= $(LINUX_PATH)/arch/arm64/boot/Image
-LINUX_DTB		?= $(LINUX_PATH)/arch/arm64/boot/dts/broadcom/bcm2710-rpi-3-b.dtb
+LINUX_DTB_RPI3_B	?= $(LINUX_PATH)/arch/arm64/boot/dts/broadcom/bcm2710-rpi-3-b.dtb
+LINUX_DTB_RPI3_BPLUS	?= $(LINUX_PATH)/arch/arm64/boot/dts/broadcom/bcm2710-rpi-3-b-plus.dtb
 MODULE_OUTPUT		?= $(ROOT)/module_output
 
 ################################################################################
@@ -163,7 +164,8 @@ buildroot: update_rootfs
 update_rootfs: arm-tf linux u-boot
 	@mkdir -p --mode=755 $(BUILDROOT_TARGET_ROOT)/boot
 	@mkdir -p --mode=755 $(BUILDROOT_TARGET_ROOT)/usr/bin
-	@install -v -p --mode=755 $(LINUX_DTB) $(BUILDROOT_TARGET_ROOT)/boot/bcm2710-rpi-3-b.dtb
+	@install -v -p --mode=755 $(LINUX_DTB_RPI3_B) $(BUILDROOT_TARGET_ROOT)/boot/bcm2710-rpi-3-b.dtb
+	@install -v -p --mode=755 $(LINUX_DTB_RPI3_BPLUS) $(BUILDROOT_TARGET_ROOT)/boot/bcm2710-rpi-3-b-plus.dtb
 	@install -v -p --mode=755 $(RPI3_BOOT_CONFIG) $(BUILDROOT_TARGET_ROOT)/boot/config.txt
 	@install -v -p --mode=755 $(LINUX_IMAGE) $(BUILDROOT_TARGET_ROOT)/boot/kernel8.img
 	@install -v -p --mode=755 $(ARM_TF_BOOT) $(BUILDROOT_TARGET_ROOT)/boot/armstub8.bin


rpi3.mk
```diff
@@ -11,7 +11,8 @@ override COMPILE_S_KERNEL  := 64
 # Need to set this before including common.mk
 BUILDROOT_GETTY_PORT ?= ttyS0
 BR2_ROOTFS_POST_BUILD_SCRIPT ?= "board/raspberrypi3-64/post-build.sh"
-
+#For Raspberry Pi 3 B plus set to rpi3-bplus
```
jbech-linaro: Add space after `#`.

rpi3.mk
```diff
@@ -11,7 +11,8 @@ override COMPILE_S_KERNEL  := 64
...
+PLATFORM ?= rpi3-b
```
jbech-linaro: I think I saw a comment from you somewhere that you couldn't use the composite form? Or did I mix that up? What I'm looking for here is something like
```make
PLATFORM=rpi-3b
PLATFORM=rpi-3bplus
```

rpi3.mk
```diff
@@ -31,13 +32,21 @@ RPI3_BOOT_CONFIG	?= $(RPI3_FIRMWARE_PATH)/config.txt
 RPI3_UBOOT_ENV		?= $(ROOT)/out/uboot.env
 RPI3_UBOOT_ENV_TXT	?= $(RPI3_FIRMWARE_PATH)/uboot.env.txt
 RPI3_STOCK_FW_PATH	?= $(ROOT)/firmware
-RPI3_STOCK_FW_PATH_BOOT	?= $(RPI3_STOCK_FW_PATH)/boot
+ifeq ($(PLATFORM),rpi3-b)
+RPI3_STOCK_FW_PATH_BOOT ?= $(RPI3_STOCK_FW_PATH)/boot
```
jbech-linaro: When first looking at this I ask myself why does this have to be on different paths (line 36 and 38, besides the PLATFORM)? I guess you didn't want to deal with the "firmware" git? In that case I believe it's better to actually [create another xml/manifest file](https://optee.readthedocs.io/building/gits/manifest.html) for Rpi3-b+.

rpi3.mk
```diff
@@ -31,13 +32,21 @@ RPI3_BOOT_CONFIG	?= $(RPI3_FIRMWARE_PATH)/config.txt
...
 OPTEE_BIN		?= $(OPTEE_PATH)/out/arm/core/tee-header_v2.bin
 OPTEE_BIN_EXTRA1	?= $(OPTEE_PATH)/out/arm/core/tee-pager_v2.bin
 OPTEE_BIN_EXTRA2	?= $(OPTEE_PATH)/out/arm/core/tee-pageable_v2.bin
 
 LINUX_IMAGE		?= $(LINUX_PATH)/arch/arm64/boot/Image
-LINUX_DTB		?= $(LINUX_PATH)/arch/arm64/boot/dts/broadcom/bcm2710-rpi-3-b.dtb
+ifeq ($(PLATFORM),rpi3-b)
+LINUX_DTB               ?= $(LINUX_PATH)/arch/arm64/boot/dts/broadcom/bcm2710-rpi-3-b.dtb
+else
+LINUX_DTB               ?= $(LINUX_PATH)/arch/arm64/boot/dts/broadcom/bcm2710-rpi-3-b-plus.dtb
```
jbech-linaro: This is good, but I noticed that the same dtb is showing up on several places, which might be a bit confusing. We should try clean that up at some point in the future.
```bash
$ find . -name "bcm2710-rpi-3-b.dtb"
./build/rpi3/firmware/rpi3-bplus/bcm2710-rpi-3-b.dtb
./linux/arch/arm64/boot/dts/broadcom/bcm2710-rpi-3-b.dtb
./firmware/boot/bcm2710-rpi-3-b.dtb
./out-br/target/boot/bcm2710-rpi-3-b.dtb

$ find . -name "bcm2710-rpi-3-b-plus.dtb"
./build/rpi3/firmware/rpi3-bplus/bcm2710-rpi-3-b-plus.dtb
./linux/arch/arm64/boot/dts/broadcom/bcm2710-rpi-3-b-plus.dtb
```

rpi3.mk
```diff
@@ -147,10 +156,18 @@ linux-cleaner: linux-cleaner-common
 ################################################################################
 # OP-TEE
 ################################################################################
-OPTEE_OS_COMMON_FLAGS += PLATFORM=rpi3
+ifeq ($(PLATFORM),rpi3-b)
```
jbech-linaro: Line 159 to 162: Please update to the platform names as I proposed further up.

rpi3.mk
```diff
@@ -179,6 +197,24 @@ update_rootfs: arm-tf linux u-boot
 	@install -v -p --mode=755 $(RPI3_STOCK_FW_PATH)/boot/start_db.elf $(BUILDROOT_TARGET_ROOT)/boot/start_db.elf
 	@install -v -p --mode=755 $(RPI3_STOCK_FW_PATH)/boot/start.elf $(BUILDROOT_TARGET_ROOT)/boot/start.elf
 	@install -v -p --mode=755 $(RPI3_STOCK_FW_PATH)/boot/start_x.elf $(BUILDROOT_TARGET_ROOT)/boot/start_x.elf
+else
```
jbech-linaro: This `else` is almost unnecessary with exception of a few files.  If we put all files under a common path where only `$(PLATFORM)` makes them different, then there should be almost no need for `if/else`.

rpi3/firmware/rpi3-bplus/COPYING.linux
```diff
@@ -0,0 +1,356 @@
+
```
jbech-linaro: I don't want to get GPL-v2 files in this git, which even more justifies creating another manifest to just clone the firmware (as done already today), instead of copy files like this.

rpi3/firmware/rpi3-bplus/LICENCE.broadcom
```diff
@@ -0,0 +1,30 @@
+Copyright (c) 2006, Broadcom Corporation.
```
jbech-linaro: Same here, this belongs to the firmware.git instead.

rpi3/firmware/rpi3-bplus/config.txt
```diff
@@ -0,0 +1,3 @@
+enable_uart=1
```
jbech-linaro: This belongs here, only thing is to change it to my proposed platform name instead `rpi-3bplus`.

rpi3.mk
jforissier: Platform directory in `optee_os` is `core/arch/arm/plat-rpi3` so PLATFORM has to be `rpi3` and `PLATFORM_FLAVOR` may be whatever we want (for instance `b` or `bplus`).
The consequence is, the "composite" form of `PLATFORM` is written `rpi3-something`.

https://github.com/OP-TEE/optee_os/blob/master/Makefile#L32-L33

Another option would be to rename `core/arch/arm/plat-rpi3` to `core/arch/arm/plat-rpi`, but since all variants are still 3 more or less, would it be justified?

jbech-linaro: > Another option would be to rename `core/arch/arm/plat-rpi3` to `core/arch/arm/plat-rpi`, but since all variants are still 3 more or less, would it be justified?

I was thinking about the same. My initial thought was to leave the `rpi3-...` and start using `rpi-...` instead. But as long as it's only rpi3 devices we have, it might be more work than we gain by doing so. In that case, use what I think I proposed in the other github issue about this:
```make
PLATFORM=rpi3-b
PLATFORM=rpi3-bplus
```

igoropaniuk:
@johnphilby Please remove `kernel8.img`, `armstub8.bin`, `uboot.env` and *.dtbs, as they also are being built (Linux kernel, TF-A and U-boot accordingly)

And you also don't need another copy of rpi3 firmware binaries, as the one for RPi3-B also works for RPi3-B-Plus. You can just do what Joakim suggested, creating a separate repo manifest for rpi3b-plus (and using the same link from https://github.com/OP-TEE/manifest/blob/master/rpi3.xml#L25)

johnphilby:
@jbech-linaro and @igoropaniuk thank you for your review comments. I will incorporate them and resubmit again.

johnphilby:
@jbech-linaro @igoropaniuk I think I have addressed your comments. Created a new manifest for supporting RPi3 B Plus. Please review the commits from both the "manifest" and "build" repo.

https://github.com/OP-TEE/manifest/pull/134

jbech-linaro:
This all looks good, having that said, the only difference compared to the original `rpi3.mk` file are two lines pointing to the DTB file. Therefore I'd like to get your opinion about eventually sharing the same Makefile for both flavors. I.e., we could select either one (see "Suggestion one" below) depending on the platform or ... we simply just include the DTB for both platforms (see "Suggestion two" below). It's not the end of the day if there is an unused DTB laying around if it otherwise simplifies things a lot. Having that said, the latter idea can only be used of both DTB's actually are being compiled when building the kernel, which after thinking about it I believe is not the case.

The downside with both of my proposals are that the user must not forget to set the PLATFORM flag when building for RPi3 B-plus which is something that we doesn't really do for other OP-TEE full developer builds.

In any case, I'm happy with it is as it is here also, we can refactor the makefiles at a later point also to better share common parts (`ti/ti-common.mk` is probably the best to compare with).

**Suggestion one**:
```patch
diff --git a/rpi3.mk b/rpi3.mk
index e172571..d7918c5 100644
--- a/rpi3.mk
+++ b/rpi3.mk
@@ -37,9 +37,17 @@ OPTEE_BIN_EXTRA1	?= $(OPTEE_PATH)/out/arm/core/tee-pager_v2.bin
 OPTEE_BIN_EXTRA2	?= $(OPTEE_PATH)/out/arm/core/tee-pageable_v2.bin
 
 LINUX_IMAGE		?= $(LINUX_PATH)/arch/arm64/boot/Image
+
+ifeq ($(PLATFORM),rpi3-bplus)
+LINUX_DTB		?= $(LINUX_PATH)/arch/arm64/boot/dts/broadcom/bcm2710-rpi-3-b-plus.dtb
+$(info "Building RPi3-bplus")
+else
 LINUX_DTB		?= $(LINUX_PATH)/arch/arm64/boot/dts/broadcom/bcm2710-rpi-3-b.dtb
+$(info "Building RPi3b (default)")
+endif
 MODULE_OUTPUT		?= $(ROOT)/module_output
 
+
 # Targets
@@ -163,7 +171,11 @@ buildroot: update_rootfs
 update_rootfs: arm-tf linux u-boot
 	@mkdir -p --mode=755 $(BUILDROOT_TARGET_ROOT)/boot
 	@mkdir -p --mode=755 $(BUILDROOT_TARGET_ROOT)/usr/bin
+ifeq ($(PLATFORM),rpi3-bplus)
+	@install -v -p --mode=755 $(LINUX_DTB) $(BUILDROOT_TARGET_ROOT)/boot/bcm2710-rpi-3-b-plus.dtb
+else
 	@install -v -p --mode=755 $(LINUX_DTB) $(BUILDROOT_TARGET_ROOT)/boot/bcm2710-rpi-3-b.dtb
+endif
 	@install -v -p --mode=755 $(RPI3_BOOT_CONFIG) $(BUILDROOT_TARGET_ROOT)/boot/config.txt
 	@install -v -p --mode=755 $(LINUX_IMAGE) $(BUILDROOT_TARGET_ROOT)/boot/kernel8.img
 	@install -v -p --mode=755 $(ARM_TF_BOOT) $(BUILDROOT_TARGET_ROOT)/boot/armstub8.bin
```

**Suggestion two**:
```patch
diff --git a/rpi3.mk b/rpi3.mk
index e172571..4272cb2 100644
--- a/rpi3.mk
+++ b/rpi3.mk
@@ -37,9 +37,12 @@ OPTEE_BIN_EXTRA1	?= $(OPTEE_PATH)/out/arm/core/tee-pager_v2.bin
 OPTEE_BIN_EXTRA2	?= $(OPTEE_PATH)/out/arm/core/tee-pageable_v2.bin
 
 LINUX_IMAGE		?= $(LINUX_PATH)/arch/arm64/boot/Image
-LINUX_DTB		?= $(LINUX_PATH)/arch/arm64/boot/dts/broadcom/bcm2710-rpi-3-b.dtb
+
+LINUX_DTB_RPI3B		?= $(LINUX_PATH)/arch/arm64/boot/dts/broadcom/bcm2710-rpi-3-b.dtb
+LINUX_DTB_RPI3BPLUS	?= $(LINUX_PATH)/arch/arm64/boot/dts/broadcom/bcm2710-rpi-3-b-plus.dtb
 MODULE_OUTPUT		?= $(ROOT)/module_output
 
+
 # Targets
@@ -163,7 +166,8 @@ buildroot: update_rootfs
 update_rootfs: arm-tf linux u-boot
 	@mkdir -p --mode=755 $(BUILDROOT_TARGET_ROOT)/boot
 	@mkdir -p --mode=755 $(BUILDROOT_TARGET_ROOT)/usr/bin
-	@install -v -p --mode=755 $(LINUX_DTB) $(BUILDROOT_TARGET_ROOT)/boot/bcm2710-rpi-3-b.dtb
+	@install -v -p --mode=755 $(LINUX_DTB_RPI3B) $(BUILDROOT_TARGET_ROOT)/boot/bcm2710-rpi-3-b.dtb
+	@install -v -p --mode=755 $(LINUX_DTB_RPI3BPLUS) $(BUILDROOT_TARGET_ROOT)/boot/bcm2710-rpi-3-b-plus.dtb
 	@install -v -p --mode=755 $(RPI3_BOOT_CONFIG) $(BUILDROOT_TARGET_ROOT)/boot/config.txt
 	@install -v -p --mode=755 $(LINUX_IMAGE) $(BUILDROOT_TARGET_ROOT)/boot/kernel8.img
 	@install -v -p --mode=755 $(ARM_TF_BOOT) $(BUILDROOT_TARGET_ROOT)/boot/armstub8.bin
```

igoropaniuk:
@jbech-linaro my two cents:
I thinkthat  just copying 2 dtbs (For RPi3 and RPi3B plus) makes the deal and is the most simple solution here that works. I'll vote for this

jbech-linaro:
> @jbech-linaro my two cents:
> I thinkthat just copying 2 dtbs (For RPi3 and RPi3B plus) makes the deal and is the most simple solution here that works. I'll vote for this

Yeah, I guess that this is the better solution as of now, but that requires that both dtb's are there. I believe the install command will complain otherwise.

igoropaniuk:
@jbech-linaro AFAIK, they are already in the same rpi3 firmware tarball (there is no need to download something else), so anyway you can just copy both after unpacking the archive.

jbech-linaro:
I see, well,  then it's a no brainer I guess.

johnphilby:
> @jbech-linaro my two cents:
> I thinkthat just copying 2 dtbs (For RPi3 and RPi3B plus) makes the deal and is the most simple solution here that works. I'll vote for this

@jbech-linaro In lieu of this comment and our common consensus that it is cleaner to just copy both the dtbs, I have updated the pull request.

To sum up: 
a) we will avoid creating another rpi3-bplus makefile.
b) we will also avoid creating an extra manifest file too.

Please take a look.

https://github.com/OP-TEE/build/pull/358/commits/f4ad1e3bfdead4a25ff69223123541c8afed58b7

https://github.com/OP-TEE/manifest/commit/397ab2729ac08575e362d45f04f5d5964d669bb0

jbech-linaro:
> I am able to restart the job. I am letting you know because I don't know if it is intended for random people to have that permission.

@IKyriazis I'm well aware of that and as long as it isn't misused I'm going to leave it like that. But if there is someone starting to misuse it, then I'll have to remove that ability for non-trusted users. Anyhow, thanks for mentioning this!

jbech-linaro:
@johnphilby , sorry for the late reply, this is looking good to me.

jforissier:
@johnphilby, would you please squash the 3 commits into one and add @jbech-linaro's Reviewed-by tag?
Then I can merge this PR.

johnphilby:
> @johnphilby, would you please squash the 3 commits into one and add @jbech-linaro's Reviewed-by tag?
> Then I can merge this PR.

Sure. Doing that right now.

jbech-linaro:
FYI, as part of testing for 3.6.0, I'm doing a test build with this PR.

jbech-linaro:
> FYI, as part of testing for 3.6.0, I'm doing a test build with this PR.

With the patches applied on master.
```bash
24078 subtests of which 0 failed
95 test cases of which 0 failed
0 test cases were skipped
TEE test application done!
```
So, if tags haven't already been applied you can add my TB also.
`Tested-by: Joakim Bech <joakim.bech@linaro.org>` (rpi3b)

jforissier:
@jbech-linaro why is IBART failing? Wrong cross-compiler version?
```
14089:  Incorrect selection of kernel headers: expected 4.18.x, got 4.19.x
```

johnphilby:
> Tested-by: Joakim Bech <joakim.bech@linaro.org>

I will add this as well. I am just trying out this interactive rebase thingy. Did not get it right. Trying again. Apparently, "--autosquash" isn't working as I would like it to.

johnphilby:
Guess I got it right this time.

Please verify and merge.

https://github.com/johnphilby/build/commit/1c0e779370efb7e7f0fe6f563e69620d76969805

and

https://github.com/johnphilby/manifest/commit/637979080e1e63c16c9e08e39c727a83e105c142

jbech-linaro:
> @jbech-linaro why is IBART failing? Wrong cross-compiler version?
> 
> ```
> 14089:  Incorrect selection of kernel headers: expected 4.18.x, got 4.19.x
> ```

It's because @johnphilby branch is a bit old, it needs a rebase so it gets my related patches onto his own branch (https://github.com/OP-TEE/build/commit/6e167bf1144ce50a7c9c855c54e75e1b655d59f4 and https://github.com/OP-TEE/build/commit/e029816141bbe106e4574a69dab242f5b017d8b9).

jforissier:
> It's because @johnphilby branch is a bit old,

OK, then I guess all is well. Let's merge this. Thanks!

PR #362, opened 2019-04-23
common: remove depraceted gnome-terminal flag

jbech-linaro:
On recent versions of gnome-terminal you will get this warning:

"Option "-t" is deprecated and might be removed in a later version of
gnome-terminal."

There seems to be no substitution for it, therefore let's just get rid
of it.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -343,8 +343,6 @@ define run-help
 	@echo
 endef
 
-# Note: Using the LAUNCH_TERMINAL environment variable, it is not currently possible to set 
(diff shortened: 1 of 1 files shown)


jbech-linaro:
Fixed the commit message and removed the two "note" comments.
Squashed, rebased and tag(s) applied (should be) ready for merge (no hurry can save it until after the 3.5.0 release)! Thanks!

jbech-linaro:
Patch rebased, since we're done with the 3.5.0. I think we can go ahead and merge this now.

PR #366, opened 2019-05-02
buildroot: fix installation of shared libraries

jforissier:
Fixes a mistake in the optee_os install step which prevents the
libraries from being copied into the root FS staging area.

Fixes: 91ebff727d47 ("buildroot: add optee_os package to copy shared libraries into the root FS")


jenswi-linaro:

PR #368, opened 2019-05-13
Hikey960 Secure Data Path + one buildroot fix

jforissier:
A couple of patches to fix a build error I encountered with buildroot, and to enable Secure Data Path by default on HiKey960.


kconfigs/hikey960.conf
```diff
@@ -14,3 +14,12 @@ CONFIG_SQUASHFS_4K_DEVBLK_SIZE=y
 CONFIG_SCSI_UFSHCD=y
 CONFIG_SCSI_UFSHCD_PLATFORM=y
 CONFIG_SCSI_UFS_HISI=y
+
+CONFIG_STAGING=y
+CONFIG_ANDROID=y
+CONFIG_ION=y
+CONFIG_ION_DUMMY=y
```
etienne-lms: `CONFIG_ION_DUMMY=y`: this is a old config from a deprecated ION feature. We should get rid of. No need to enable it.

jforissier: OK, I'll remove it.

jforissier:
Update

PR #373, opened 2019-05-29
qemu_v8: fix "make check"

jforissier:
Updates qemu_v8.mk and qemu-check.exp so that "make check" and "make
check-only" are supported for QEMUv8, as they are for QEMU.


jforissier:
Ping?

jbech-linaro:

PR #374, opened 2019-06-06
[RFC] upgrade buildroot to latest release (2019.05)

etienne-lms:
Since release 2019.05, Buildroot integrates native OP-TEE packages. One could use them to build the its platform, including OP-TEE client, core, TA SDK, and examples and test CAs/TAs. Yet when built through buildroot, fine tuning some OP-TEE feature configurations (32b/64b TA modes, RMPB emulation, ...) needs hacking into Buildroot source tree. Evolutions of OP-TEE packages may also need BR source tree to be updated, which is not the goal of the generic OP-TEE build resources.

This P-R proposes to keep OP-TEE build process as-is with OP-TEE packages as Buildroot external packages. This require package renaming to not conflict with Buildroot native OP-TEE packages.

This RFC proposes some extra cleanup changes on top of the package renaming.


jbech-linaro:
As such I think this looks good, I haven't reviewed it deeply though. It feels like these changes are transparent to the user of OP-TEE, but it seems like at least xtest hasn't been built properly, i.e., IBART cannot find any xtest binary.

jbech-linaro:
@etienne-lms , I'm removing the stale label for this one, since as we discussed the other day. This might be quite urgent suddenly.

etienne-lms:
Should this P-R be opened back?

jbech-linaro:
> Should this P-R be opened back?

Yes, but I guess it can wait until we're actually intend to do work on it.

PR #379, opened 2019-07-20
qemu_v8: Add support for Secure Boot (TBB)

vchong:

### DIFF
--- a/qemu_v8.mk
+++ b/qemu_v8.mk
@@ -8,6 +8,11 @@ override COMPILE_NS_KERNEL := 64
 override COMPILE_S_USER    := 64
 override COMPILE_S_KERNEL  := 64
 
+################################################################################
+# If you change this, you MUST run `make arm-tf-clean` first before rebuilding
+################################################################################
+ARM_TF_TRUSTED_BOARD_BOOT ?= n
+
 BR2_ROOTFS_OVERLAY = $(ROOT)/build/br-ext/board/qemu/overlay
 
 include common.mk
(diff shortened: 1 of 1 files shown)


vchong:
Depends on https://github.com/OP-TEE/manifest/pull/141

qemu_v8.mk
```diff
@@ -8,6 +8,11 @@ override COMPILE_NS_KERNEL := 64
 override COMPILE_S_USER    := 64
 override COMPILE_S_KERNEL  := 64
 
+################################################################################
+# If you change this, you MUST run `make arm-tf-clean` first before rebuilding
+################################################################################
+CFG_SECURE_BOOT ?= 0
```
jenswi-linaro: We're usually assigning `CFG_` variables `y` or `n`. Numbers are only used if it's indeed representing a number instead of true or false.

vchong: Was trying to 'match' arm-tf's `TRUSTED_BOARD_BOOT=1`.
Changed this to `n` instead.
Please continue review.
Thanks!

qemu_v8.mk
```diff
@@ -60,12 +65,29 @@ ARM_TF_FLAGS ?= \
 	DEBUG=$(ARM_TF_DEBUG) \
 	LOG_LEVEL=$(ARM_TF_LOGLVL)
 
+ifeq ($(CFG_SECURE_BOOT),1)
+ARM_TF_FLAGS += \
+	MBEDTLS_DIR=$(ARM_TF_PATH)/../mbedtls \
```
etienne-lms: This change expects mbedtls source tree from the project root dir?
If so, would prefer `$(ROOT)/mbedtls`.

qemu_v8.mk
etienne-lms: I would rather rename `CFG_SECURE_BOOT` into `ARM_TF_TRUSTED_BOOT` to no confuse with OP-TEE configurations (`CFG_*`), as `ARM_TF_PATH`, `ARM_TF_EXPORTS`, `ARM_TF_DEBUG`, ...

jbech-linaro: Nooo, if we should rename anything, then it's probably all .... `s/ARM_TF.../TF_A/g`. With that done, then sure `TF-A_SECURE_BOOT` or something will do.

qemu_v8.mk
vchong: Ok. Will change.

qemu_v8.mk
vchong: So what's the consensus here? Ok to go with `CFG_SECURE_BOOT` for now? Or use the actual arm-tf flag of `TRUSTED_BOARD_BOOT={0,1}`? Or something else?

jbech-linaro: Understand that you're confused, three different "ideas" being raised. I think I'm with Etienne regarding CFG vs anything else, especially since this only touches TF-A related stuff. So, contrary to what I just said in previous reply, please do like this:
1. Use `ARM_TF_TRUSTED_BOOT={y,n}` for this PR.
2. Create a new PR where you rename ARM_TF to TF_A.

jforissier: Adding my $0.02 ;-) If `TRUSTED_BOARD_BOOT={0,1}` would work here, why not use it as-is? Having several names for one setting can only bring confusion IMO.

vchong:
All comments addressed

vchong:
Rebased and tags added. Ok to merge. Thanks!

PR #380, opened 2019-08-03
zynqmp.mk: Add support for Ultra96 ZynqMP board

mimok:
This PR adds support for Ultra96 board.
It relies on ultra96 flavor added to OP-TEE ZynqMP platform.
zynqmp makefile is modified in the following way:
- Default BSP rootfs is replaced by a minimal rootfs which supports
zcu10x and Ultra96 boards.
- New petalinux-package target available to generate BOOT.BIN file.

### DIFF
--- a/zynqmp.mk
+++ b/zynqmp.mk
@@ -1,8 +1,8 @@
 ROOT			= $(PWD)/..
 PLATFORM		?= zcu102
-PETALINUX_PATH		?= /opt/Xilinx/petalinx_2018_2
-BSP_PATH		?= ./xilinx-${PLATFORM}-v2018.2-final.bsp
-PRJ_PATH		?= $(ROOT)/xilinx-${PLATFORM}-2018.2
+PETALINUX_PATH		?= $(PETALINUX)
+BSP_PATH		?= ./xilinx-${PLATFORM}-v$(PETALINUX_VER)-final.bsp
+PRJ_PATH		?= $(ROOT)/$(PLATFORM)-$(PETALINUX_VER)
 PETALINUX_CFG_PATH	?= $(ROOT)/build/zynqmp
 OPTEE_VER		?= latest
 
@@ -20,8 +20,16 @@ define set_optee_version
 	fi
 endef
 
-.PHONY: check-petalinux
+ifeq ($(PLATFORM),ultra96-reva)
+	ZYNQMP_CONSOLE=cadence1
+else
+	ZYNQMP_CONSOLE=cadence0
+endif
 
+.PHONY: all
+all: petalinux-create petalinux-config petalinux-build petalinux-package
+
+.PHONY: check-petalinux
 check-petalinux:
 ifndef PETALINUX_VER
 	$(error You have to source Petalinux settings)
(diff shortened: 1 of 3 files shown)


zynqmp.mk
```diff
@@ -1,8 +1,8 @@
 ROOT			= $(PWD)/..
 PLATFORM		?= zcu102
-PETALINUX_PATH		?= /opt/Xilinx/petalinx_2018_2
+PETALINUX_PATH		?= $(PETALINUX)
```
jbech-linaro: I cannot see that `$(PETALINUX)` is defined elsewhere? You want to enforce this to be provided from command line? Isn't it possible to have some common ground to start from as a default?

jbech-linaro: Is this in the `settings.sh` file? That comes with the petalinux package?

zynqmp.mk
```diff
@@ -30,21 +38,27 @@ ifneq ($(PETALINUX_VER),2018.2)
...
-	@cd $(ROOT) && petalinux-create -t project -s $(BSP_PATH)
-	@cd $(PRJ_PATH) && petalinux-create -t apps --template install -n optee-client --enable
-	@cd $(PRJ_PATH) && petalinux-create -t apps --template install -n optee-test --enable
-	@#
+petalinux-create: check-petalinux	
+	@cd $(ROOT) && petalinux-create -n $(PLATFORM)-$(PETALINUX_VER) -t project -s $(BSP_PATH)
 	$(call set_cfg,CONFIG_SUBSYSTEM_ATF_COMPILE_EXTRA_SETTINGS,"SPD=opteed ZYNQMP_BL32_MEM_BASE=0x60000000 ZYNQMP_BL32_MEM_SIZE=0x80000",$(PRJ_PATH)/project-spec/configs/config)
 	$(call set_cfg,CONFIG_SUBSYSTEM_ZYNQMP_ATF_MEM_SIZE,0x16001,$(PRJ_PATH)/project-spec/configs/config)
 	@#
+	@# Replace BSP default rootfs by a minimal one to speed up building process and ease compatibility between different boards
```
jbech-linaro: I know there are plenty of lines here exceeding 80 characters, but please try to fit things into 80 character if possible (at least for comments it should be easy to achive).

zynqmp.mk
```diff
@@ -1,8 +1,8 @@
 ROOT			= $(PWD)/..
 PLATFORM		?= zcu102
-PETALINUX_PATH		?= /opt/Xilinx/petalinx_2018_2
+PETALINUX_PATH		?= $(PETALINUX)
 BSP_PATH		?= ./xilinx-${PLATFORM}-v2018.2-final.bsp
-PRJ_PATH		?= $(ROOT)/xilinx-${PLATFORM}-2018.2
+PRJ_PATH		?= $(ROOT)/$(PLATFORM)-$(PETALINUX_VER)
```
jbech-linaro: I had the same question regarding `$(PETALINUX_VER)`, then then I noticed the make target `check-petalinux` which seems to deal with this, so once again I supposed this is a variable stored in `settings.sh`.

zynqmp.mk
```diff
@@ -1,8 +1,8 @@
 ROOT			= $(PWD)/..
 PLATFORM		?= zcu102
-PETALINUX_PATH		?= /opt/Xilinx/petalinx_2018_2
+PETALINUX_PATH		?= $(PETALINUX)
 BSP_PATH		?= ./xilinx-${PLATFORM}-v2018.2-final.bsp
```
jbech-linaro: Will this still have a "static" value (`v2018.2-final`)? Shouldn't that also leverage `$(PETALINUX_VER)`?

zynqmp.mk
```diff
@@ -54,10 +68,15 @@ petalinux-config:
 	$(call set_optee_version,$(OPTEE_VER),$(PRJ_PATH)/project-spec/meta-user/recipes-apps/optee-test/optee-test.bbappend)
 	$(call set_optee_version,$(OPTEE_VER),$(PRJ_PATH)/project-spec/meta-user/recipes-apps/optee-client/optee-client.bbappend)
 	$(call set_optee_version,$(OPTEE_VER),$(PRJ_PATH)/project-spec/meta-user/recipes-bsp/optee-os/optee-os.bbappend)
-	@cd $(PRJ_PATH) && petalinux-config --oldconfig
+	@petalinux-config -p $(PRJ_PATH) --oldconfig
```
jbech-linaro: Curious to know, are these executables (... config,build etc) setup also when running `settings.sh`?

zynqmp.mk
mimok: Yes `$(PETALINUX)` is defined by when you source `settings.sh`

zynqmp.mk
mimok: Yes, i added `check-petalinux` target as dependency for every other targets to be sure that user has actually sourced `settings.sh` before trying to use makefile.

zynqmp.mk
mimok: updated

zynqmp.mk
mimok: yes, `settings.sh` is used to setup everything related to petalinux (variable, paths to binaries, etc.)

zynqmp.mk
mimok: Is it ok for you? Result is a little less readable, in my opinion.

mimok:
All comment taken into account, if you are ok i will squash commits.

mimok:
i had a problem with division of at least one line (line where `call` is used to set some ATF parameters). Sould be ok now, test is in progress.

mimok:
problem solved

zynqmp.mk
```diff
@@ -30,34 +38,61 @@ ifneq ($(PETALINUX_VER),2018.2)
...
 endif
 
-petalinux-create: check-petalinux
-	@cd $(ROOT) && petalinux-create -t project -s $(BSP_PATH)
-	@cd $(PRJ_PATH) && petalinux-create -t apps --template install -n optee-client --enable
-	@cd $(PRJ_PATH) && petalinux-create -t apps --template install -n optee-test --enable
-	@#
+petalinux-create: check-petalinux	
+	@cd $(ROOT) && petalinux-create -n $(PLATFORM)-$(PETALINUX_VER) \
+	-t project -s $(BSP_PATH)
```
etienne-lms: suggest to add tab or space indentation for such multi-line command lines. Applies here and to many lines below:
```Patch
 	@cd $(ROOT) && petalinux-create -n $(PLATFORM)-$(PETALINUX_VER) \
-	-t project -s $(BSP_PATH)
+	    -t project -s $(BSP_PATH)
```

mimok:
In my opinion this PR is ready to be merged into master and the corresponding documentation has been already merged into `optee_docs`.  I'm waiting for the OP-TEE team validation. Let me know when I can finalize this PR by squashing commits.

jbech-linaro:
> In my opinion this PR is ready to be merged into master

Ah, sorry I missed that you had pushed "Second pass of correction". So I guess we're good to go with this one. Feel free to add my tag and squash it all together.

jforissier:
@mimok when you squash, can you also rebase onto master? Should fix the IBART error.

mimok:
@jbech-linaro ok I will do that tonight. 
@jforissier Sure

mimok:
@jbech-linaro and @jforissier Done

PR #383, opened 2019-09-03
common.mk: buildroot: append BR2_ variables to config automatically

jforissier:
To set additional BR2_ variables, the user currently needs to edit
common.mk and modify the buildroot recipe. It is iconvenient, especially
since a line with a cumbersome syntax has to be added. One would expect
to be able to specifiy the desired value on the make command line or
directly in the Makefile. For instance, to enable debug mode:
make BR2_ENABLE_DEBUG=y.

This patch updates the buildroot recipe to enumerate all the Makefile
variables that start with BR2_, and each such variable is automatically
added to ../out-br/extra.conf.

As a result the existing BR2_ variables can be set in a simpler way.

Tested on QEMU, results in the same ../out-br/.config except for the
line order.

Suggested-by: Heino Madsen <heino.madsen@huawei.com>

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -194,55 +194,44 @@ else
 BUILDROOT_TOOLCHAIN=toolchain-aarch$(COMPILE_NS_USER)-legacy
 endif
 endif
-BUILDROOT_GETTY_PORT ?= \
-	$(if $(CFG_NW_CONSOLE_UART),ttyAMA$(CFG_NW_CONSOLE_UART),ttyAMA0)
+
+BR2_PACKAGE_LIBOPENSSL ?= y
+BR2_PACKAGE_MMC_UTILS ?= y
+BR2_PACKAGE_OPENSSL ?= y
+BR2_PACKAGE_OPTEE_BENCHMARK ?= $(CFG_TEE_BENCHMARK)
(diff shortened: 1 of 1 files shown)


common.mk
```diff
@@ -194,55 +194,45 @@ else
...
+BR2_PACKAGE_STRACE ?= y
+BR2_TARGET_GENERIC_GETTY_PORT ?= $(if $(CFG_NW_CONSOLE_UART),ttyAMA$(CFG_NW_CONSOLE_UART),ttyAMA0)
+
+# All BR2_* variables from the makefile or the environment are appended to
+# ../out-br/extra.conf unless already present
+# All values are quoted "..." except y and n
+double-quote = "#" # This really sets the variable to " and avoids upsetting vim's syntax highlighting
+streq = $(and $(findstring $(1),$(2)),$(findstring $(2),$(1)))
+y-or-n = $(or $(call streq,y,$(1)),$(call streq,n,$(1)))
+append-var_ = grep -q '^$(1)=' $(2) || echo '$(1)=$(3)'$($(1))'$(3)' >>$(2);
```
jenswi-linaro: Why is the grep needed?

jforissier: Yeah sorry about that I have removed it already. Initially I had not removed all the `@echo ...` lines and I thought they should have higher priority. Useless now.

jenswi-linaro:
Convenient.

PR #385, opened 2019-09-10
common.mk: use ccache to build OP-TEE user space libraries

jforissier:
Currently, when ccache is available, it is used to build OP-TEE
but due to some wrong settings the user space libraries use the
compiler directly without ccache. Fix that.


jenswi-linaro:

PR #386, opened 2019-10-02
hikey960: enable Wi-Fi

jforissier:
Adds the required flags and configuration files to enable Wi-Fi on the
HiKey960 board. The network name and password have to be set in
br-ext/board/hikey960/overlay/etc/wpa_supplicant.conf before building.
Then Wi-Fi can be enabled on the board by running 'ifup wlan0' as root.


jbech-linaro:

PR #388, opened 2019-10-07
hikey, hikey960: remove unused variable STRACE_PATH

jforissier:
STRACE_PATH is a leftover from before buildroot and is not used anymore
so remove it.


etienne-lms:
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>`

PR #392, opened 2019-10-09
poplar: provide TF-A path for l-loader

igoropaniuk:
Provide TF-A path for l-loader

`Signed-off-by: Igor Opaniuk <igor.opaniuk@gmail.com>`

### DIFF
--- a/poplar.mk
+++ b/poplar.mk
@@ -120,7 +120,8 @@ optee-os-clean: optee-os-clean-common
 l-loader: arm-tf
 	cp $(BL1_BIN) $(LLOADER_PATH)/atf
 	cp $(FIP_BIN) $(LLOADER_PATH)/atf
-	$(MAKE) -C $(LLOADER_PATH) CROSS_COMPILE="$(AARCH32_CROSS_COMPILE)"
+	$(MAKE) -C $(LLOADER_PATH) CROSS_COMPILE="$(AARCH32_CROSS_COMPILE)" \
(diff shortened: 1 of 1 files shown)


poplar.mk
```diff
@@ -120,8 +120,8 @@ optee-os-clean: optee-os-clean-common
 l-loader: arm-tf
 	cp $(BL1_BIN) $(LLOADER_PATH)/atf
 	cp $(FIP_BIN) $(LLOADER_PATH)/atf
-	$(MAKE) -C $(LLOADER_PATH) CROSS_COMPILE="$(AARCH32_CROSS_COMPILE)"
-
```
jforissier: Why remove this blank line? Keep it please.

poplar.mk
```diff
@@ -120,8 +120,8 @@ optee-os-clean: optee-os-clean-common
...
+	$(MAKE) -C $(LLOADER_PATH) CROSS_COMPILE="$(AARCH32_CROSS_COMPILE)" \
+		ARM_TRUSTED_FIRMWARE="$(TF_A_PATH)"
```
jforissier: I'd rather we keep the current name: `ARM_TRUSTED_FIRMWARE`, avoids a dependency on the same change being erged in upstream l-loader.

igoropaniuk: @jforissier 
so let's make it simpler then

poplar.mk
igoropaniuk: this is usually a kind of rhetorical question, as I have no idea why I did that :)

PR #398, opened 2019-11-22
qemu, qemu_v8: do not build Linux modules

jforissier:
The QEMU and QEMUv8 builds do not need any kernel module to boot and
run xtest etc. so we can save build time by only building the kernel
image by default.

### DIFF
--- a/qemu.mk
+++ b/qemu.mk
@@ -118,7 +118,7 @@ LINUX_DEFCONFIG_COMMON_FILES := \
 
 linux-defconfig: $(LINUX_PATH)/.config
 
-LINUX_COMMON_FLAGS += ARCH=arm
+LINUX_COMMON_FLAGS += ARCH=arm zImage
 
 linux: linux-common
 	mkdir -p $(BINARIES_PATH)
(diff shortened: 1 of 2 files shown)


jbech-linaro:

jenswi-linaro:

jbech-linaro:
Should the IBART error make us worried?
```
  1354:  smh_open: ERROR fd -1 for file 'zImage'
  1355:  => 
```

jenswi-linaro:
> Should the IBART error make us worried?

Yes, I think we need `zImage` too.

jenswi-linaro:
Looks good.

jbech-linaro:
> commit message: so we can save build time by only building the kernel image by default.

Reflection after seeing the end result. Build time seems to have increased a tiny bit. It hoovered around 22:10 / 22:20, but the last two builds for this PR was 22:16 and 22:49. Anyhow, this is PR fixed an issue, so we're all good.

jforissier:
@jbech-linaro indeed it does not make much difference with QEMU, because only a dozen features are configured as modules (`=m` in `linux/.config`). But with QEMUv8 the count is 351 and the build time on my laptop decreases from ~18 minutes to ~9 minutes (for a clean build without ccache).

jbech-linaro:
> But with QEMUv8 the count is 351 and the build time on my laptop decreases from ~18 minutes to ~9 minutes (for a clean build without ccache).

OK, then there is a clear benefit, thanks! :+1:

PR #405, opened 2019-12-22
qemu: Add VirtIORNG to prevent hangs on getentropy() calls

crtxcr:
This uses /dev/urandom to pass random data to the guest. This prevents hangs in some applications that make calls to getentropy() etc.

The rate limiting `max-bytes=1024,period=1000` has been taken from https://wiki.qemu.org/Features/VirtIORNG

Fixes https://github.com/OP-TEE/optee_os/issues/3396

As for security discussion regarding using /dev/urandom as the source, see
https://bugzilla.redhat.com/show_bug.cgi?id=1337101
https://bugzilla.redhat.com/show_bug.cgi?id=1374227
https://lists.sr.ht/~philmd/qemu/patches/3941

So all in all it should be fine. (Even if it wasn't, the main interest of the qemu target, as far as I understand is, is development work with TA and not the generation of secure keys for production usage. Although you can never rule it out... but above sources state it's fine to use /dev/urandom).


jbech-linaro:
> So all in all it should be fine. (Even if it wasn't, the main interest of the qemu target, as far as I understand is, is development work with TA and not the generation of secure keys for production usage.

Indeed, QEMU is a development platform. Thanks for the patch!

jforissier:
`Signed-off-by:` tags are missing, and commit messages should preferably wrap at 75 characters.
Other than that, LGTM.

@nageshshamnur would you please check if this fixes your issue, and give a `Tested-by:` if it does? (no obligation). Thanks.

crtxcr:
Added tags and fixed line wrapping

nageshshamnur:
Hi Jerome,
  Will surely check at the earliest and will update the status. Thanks for the patch @crtxcr

PR #406, opened 2019-12-25
devbox: build scp

FulcronZ:
* add SCP build recipe for DeveloperBox
* add required arm-none-eabi toolchain using [GNU Embedded Toolchain for Arm 9-2019-q4][arm-toolchain]

Its manifest is updated in OP-TEE/manifest#151. 
Note that the build is working now because of too large FIP package. The FIP package contains only TF-A and OP-TEE. SCP is not included in FIP, so it is not the cause breaking the build. 

[arm-toolchain]: https://developer.arm.com/tools-and-software/open-source-software/developer-tools/gnu-toolchain/gnu-rm/downloads

### DIFF
--- a/synquacer.mk
+++ b/synquacer.mk
@@ -15,19 +15,24 @@ include common.mk
 include toolchain.mk
 
 ifeq ($(DEBUG),1)
+SCP_BUILD ?= debug
 TFA_BUILD ?= debug
-else
-TFA_BUILD ?= release
-endif
-ifeq ($(DEBUG),1)
 EDK2_BUILD ?= DEBUG
 else
+SCP_BUILD ?= release
+TFA_BUILD ?= release
 EDK2_BUILD ?= RELEASE
 endif
 
 ################################################################################
 # Paths to git projects and various binaries
 ################################################################################
+BINARIES_PATH ?= $(ROOT)/out/bin
+SCP_PATH ?= $(ROOT)/SCP-firmware
+SCP_OUT ?= $(SCP_PATH)/build/product/synquacer
+SCP_ROM ?= $(SCP_OUT)/scp_romfw/$(SCP_BUILD)/bin/scp_romfw.bin
+SCP_RAM ?= $(SCP_OUT)/scp_ramfw/$(SCP_BUILD)/bin/scp_ramfw.bin
+SCP_ROMRAM ?= $(SCP_OUT)/scp_romramfw.bin
 TFA_PATH ?= $(ROOT)/trusted-firmware-a
 TFA_OUT ?= $(TFA_PATH)/build/synquacer/$(TFA_BUILD)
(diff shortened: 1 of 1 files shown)


toolchain.mk
```diff
@@ -28,12 +33,16 @@ define dltc
 endef
 
 .PHONY: toolchains
-toolchains: aarch32 aarch64
+toolchains: aarch32 aarch32-none aarch64
```
jforissier: Only `synquacer.mk` needs the `none` toolchain, so it may be wiser to add the dependency in `synquacer.mk` rather than `toolchain.mk`.

FulcronZ: good idea. will do.

synquacer.mk
```diff
@@ -46,10 +51,31 @@ EDK2_ARCH ?= AARCH64
...
+################################################################################
+# SCP
+################################################################################
+SCP_FLAGS ?= \
+	CC=$(AARCH32_NONE_CROSS_COMPILE)gcc \
+	PRODUCT=synquacer \
+	MODE=$(SCP_BUILD)
+
+.PHONY: scp
+scp: aarch32-none
```
jenswi-linaro: I think it's possible to add a `toolchains: aarch32-none` somewhere in this file instead.

FulcronZ: HNY Jens. 
OK. I will move the recipe to aarch32-none. 
I think aarch32-none should be in toolchain.mk in long run if other products start using it. 
But we can think of moving the aarch32-none recipe when that time comes.

jforissier: I agree with Jens. This whole block should be:
```make
toolchains: aarch32-none

.PHONY: scp
scp:
	$(MAKE) -C $(SCP_PATH) $(SCP_FLAGS) all
	# other commands...
```

FulcronZ: The devbox still depends on aarch64 to build other stuffs. 
Shouldn't the line contain both aarch32-none and aarch64? 
Otherwise , synquacer.mk will override the recipe and forget to download the aarch64 toolchain. 

```Makefile
toolchains: aarch32-none aarch64

.PHONY: scp
scp:
	$(MAKE) -C $(SCP_PATH) $(SCP_FLAGS) all
	# other commands...
```

jforissier: No, makefiles don't work like that. Please note that:
```make
a: b c
```
...is equivalent to:
```make
a: b
a: c
```
In other words, the second target line will not override the first, it will just add a dependency (`c`). Things would be different if the targets had recipes, indeed.

FulcronZ: Thanks for correction. It works as you said. 
Ok. I will tidy things up and push again.

jforissier: You chose to keep the dependency `scp: aarch32-none` which is technically correct, but not done for any other targets. That's fine.

PR #407, opened 2020-01-06
hikey960: fix EDK2 build with GCC 9.x

jforissier:
Actually fix the build error mentioned in the below commit. The
variable BUILD_CC is set at the wrong place (I think I had it set as
an environment variable when I tested this, which is why it would work
anyways).

Fixes: 1e8d0b5a2b38 ("hikey960: fix EDK2 build with GCC 9.x")


jenswi-linaro:

PR #408, opened 2020-01-07
hikey: fix EDK2 build with GCC 9.x

jforissier:
Fix EDK2 build error when host compiler is GCC 9.x. This is the same fix
as the one recently applied to hikey960.mk.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -148,6 +148,19 @@ endef
 
 DEBUG ?= 0
 
+# Macro to check if a compiler supports a given option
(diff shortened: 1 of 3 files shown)


jforissier:
This is annoying. Older GCC versions (< 8) such as the one used by IBART, reject the `-Wno-error=stringop-truncation` option, so the initial fix in incorrect.

Therefore, I have introduced a `cc-option` macro like we have in `optee_os` or the Linux kernel.

jenswi-linaro:

PR #409, opened 2020-01-07
hikey: do not build kernel modules

jforissier:
We currently do not copy any kernel module into our root FS, they are
not needed. Therefore, drop the "modules" target when building the
kernel to save compile time.


jenswi-linaro:

PR #413, opened 2020-02-20
br-ext: break build as soon as one TA fails to build

jforissier:
If an error occurs when building a TA, the error does not stop the
build. It may therefore go unnoticed if files have been generated
previously.

This patch updates the buildroot hooks that deals with TAs and breaks
the loop as soon as a TA faiils to build. A non-zero status is left to
the caller (a Buildroot makefile) and the build stops as expected.

### DIFF
--- a/br-ext/package/optee_examples/optee_examples.mk
+++ b/br-ext/package/optee_examples/optee_examples.mk
@@ -8,13 +8,13 @@ OPTEE_EXAMPLES_SDK = $(BR2_PACKAGE_OPTEE_EXAMPLES_SDK)
 OPTEE_EXAMPLES_CONF_OPTS = -DOPTEE_EXAMPLES_SDK=$(OPTEE_EXAMPLES_SDK)
 
 define OPTEE_EXAMPLES_BUILD_TAS
-	@for f in $(@D)/*/ta/Makefile; \
+	@(set -e; for f in $(@D)/*/ta/Makefile; \
 	do \
 	  echo Building $$f && \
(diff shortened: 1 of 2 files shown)


br-ext/package/optee_examples/optee_examples.mk
```diff
@@ -14,7 +14,8 @@ define OPTEE_EXAMPLES_BUILD_TAS
 			$(MAKE) CROSS_COMPILE="$(shell echo $(BR2_PACKAGE_OPTEE_EXAMPLES_CROSS_COMPILE))" \
 			O=out TA_DEV_KIT_DIR=$(OPTEE_EXAMPLES_SDK) \
 			$(TARGET_CONFIGURE_OPTS) -C $${f%/*} all; \
-	done
+			if [ $$? != 0 ]; then _failed=1; break; fi; \
```
jenswi-linaro: I'm wondering if `@(set -e; for ... )` couldn't be used instead. The `()` is to have a sub-shell so the `set -e` doesn't spread, this may not be needed depending on how `OPTEE_EXAMPLES_BUILD_TAS` is evaluated.

jforissier: Good suggestion, works fine. I think the parentheses are needed, based on [the way Buildroot uses this fragment](https://github.com/buildroot/buildroot/blob/95942f5fcd/package/pkg-generic.mk#L242).

jforissier:
Update

jenswi-linaro:

PR #417, opened 2020-03-23
Fix bin arch check for rpi3

GlovePuppet:
I'm building master for RPI3 and I have found a weird bug

Buildroot checks the elf files that it installs match the expected architecture - see buildroot/pkg/pkg-generic.mk the function `check_bin_arch` 

This causes the build to fail (BR expects AARCH64 but start.elf and friends are for Broadcom VideoCore III processor):

```
ERROR: architecture for "/boot/start_cd.elf" is "Broadcom VideoCore III processor", should be "AArch64"
ERROR: architecture for "/boot/start.elf" is "Broadcom VideoCore III processor", should be "AArch64"
ERROR: architecture for "/boot/start_db.elf" is "Broadcom VideoCore III processor", should be "AArch64"
ERROR: architecture for "/boot/start_x.elf" is "Broadcom VideoCore III processor", should be "AArch64"
```

The error does not occur the first time a build is made or after cleaning - I guess there is a deeper bug in BR somewhere.

Arguably, this should only be done for RPI3 builds but I don't think it will cause problems


jbech-linaro:
Huh, never seen that, but it's been a couple of months since I tried the RPI3 build. However, I trust what you're saying. However, the commit message needs to be update. Run git log and you can an idea of how a good looking commit message should look like.

GlovePuppet:
Hopefully the amended commit message is satisfactory

jbech-linaro:
> Hopefully the amended commit message is satisfactory

Almost, I'd suggest something like
```
rpi3: avoid buildroot architecture checks for boot

Buildroot checks the ELF header for each ELF it installs and compares
the MACHINE field with the architecture it is building for. If there
is a mismatch the build fails. The Raspberry 3 build install several ELF's
to the boot partition of type "Broadcom VideoCore III processor" and
not Aarch64 as expected by Buildroot.

This patch adds directory 'boot' to the list of directories excluded
from architecture checking.

Signed-off-by: GlovePuppet <touched.by.his.noodley.appendage@gmail.com>
Acked-by: Joakim Bech <joakim.bech@linaro.org>
```
Interesting email address, is that a valid address?

GlovePuppet:
> Interesting email address, is that a valid address?

Sure is :) Not sure I spelled noodley correctly tho :|

jforissier:
>> Interesting email address, is that a valid address?
> Sure is :) Not sure I spelled noodley correctly tho :|

The name is no good however :disappointed: (see https://optee.readthedocs.io/en/latest/general/contribute.html#:~:text=Use%20your%20real%20name).

GlovePuppet:
Sorry, I value my privacy so that's not going to change. 

Use the patch, re-implement it yourselves or throw it away at your discretion and with my blessing

jforissier:
@GlovePuppet no problem as far as I'm concerned. I propose we don't merge this patch now, and if some other RPi3 user meets the issue, then he/she can re-use this PR (which remains publicly accessible) for "inspiration".

GlovePuppet:
Ok, no worries, I'll leave this here too for completeness

https://github.com/OP-TEE/optee_os/issues/3197

@lcs100 It's an open goal ;)

jbech-linaro:
> Sorry, I value my privacy so that's not going to change.

Fair enough, thanks for bringing it up and finding a fix for it. Apparently I've already forgot about this once :) (seeing OP-TEE/optee_os#3197).

GlovePuppet:
More digging and it seems this is not a fix, the `check_bin_arch` test on start.elf et al is performed multiple times during the build. For now I have disabled the check, if I have time I will try and find the root cause

PR #425, opened 2020-05-25
toolchain.mk: add clang-toolchains target

jforissier:
Add "make clang-toolchains" to download and extract Clang 9.0.1 into
$(ROOT)/clang-9.0.1. Usage:

 $ make clang-toolchains
 $ PATH=../clang-9.0.1/bin:$PATH make COMPILER=clang

### DIFF
--- a/get_clang.sh
+++ b/get_clang.sh
@@ -1,0 +1,50 @@
+#!/bin/bash
+#
+# Download and extract Clang from the GitHub release page.
+# We want a x86_64 cross-compiler capable of generating aarch64 and armv7a code
+# *and* we want the compiler-rt libraries for these architectures
+# (libclang_rt.*.a).
+# Clang is configured to be able to cross-compile to all the supported
+# architectures by default (see <clang path>/bin/llc --version) which is great,
+# but compiler-rt is included only for the host architecture. Therefore we need
+# to combine several packages into one, which is the purpose of this script.
+#
+# Usage: get_clang.sh [path]
+
+DEST=${1:-./clang-9.0.1}
+
+VER=9.0.1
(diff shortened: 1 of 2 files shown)


get_clang.sh
```diff
@@ -0,0 +1,52 @@
...
+
+function cleanup() {
+  rm -f ${X86_64}.tar.xz ${AARCH64}.tar.xz ${ARMV7A}.tar.xz
+  rm -rf ${AARCH64} ${ARMV7A}
+}
+
+trap "{ exit 2; }" INT
+trap cleanup EXIT
+
+set -x
```
vchong, line 38: Copy paste typo or do you mean to `set +x` instead?

jforissier: I really meant `set -x` in order to show progress and help diagnose any issue. The output is not very tidy but I thought it is a simple and effective solution.

vchong: Isn't it already set in L21?

jforissier: Huh, yes. I was wondering why you were mentioning copy/paste :confused: Will remove this one, thanks.

jforissier:
> `Acked-by: Etienne Carriere <etienne.carriere@linaro.org>` with suggestion in commit log.
> 
> To avoid path issue in the build process, suggestion to change commit log:
> 
> ```diff
> - $ PATH=../clang-9.0.1/bin:$PATH make COMPILER=clang
> + $ PATH=$(PWD)/../clang-9.0.1/bin:$PATH make COMPILER=clang
> ```

Will fix, thanks.

etienne-lms:
> Will fix, thanks. It is $PWD or ${PWD} in the shell by the way.

indeed :)

PR #426, opened 2020-05-26
Fix build to dump BR to later releases + few cleanup changes

etienne-lms:


jenswi-linaro:
It sounds a bit violent when reading "_ext termination", how about "_ext suffix" instead?
Now that we have these packages in the Buildroot repository, can't we use them instead?

etienne-lms:
> Now that we have these packages in the Buildroot repository, can't we use them instead?

I considered that since a while but the main issues are that we test OP-TEE with quite exotic :) configuration switches set from our make command line for convenience. Using Buildroot native OP-TEE packages would mean each new `CFG_` switch must be declared in the makefile of BR optee_os package, hence we would need to hack into buildroot repo. Also BR considers only a single toolchain. This prevents us from testing setup with 64bit Core and 32bit TAs for example.

In all one, I think it is easier for OP-TEE official release to use our current build environment.
Unless there is a nice way to tackle those issues.

FYI: one can still experience OP-TEE Qemu/armv7 test using Buildroot native OP-TEE packages:
clone buildroot and follow instruction from **board/qemu/arm-vexpress-tz/readme.txt**.

FYI: I will soon send a post to BR ML to bump its OP-TEE packages from 3.7.0 to the recent 3.9.0.

etienne-lms:
> It sounds a bit violent when reading "_ext termination", how about "_ext suffix" instead?

Sorry being violent! i'll change to suffix.

common.mk
```diff
@@ -395,7 +369,9 @@ OPTEE_OS_COMMON_FLAGS ?= \
 optee-os-common:
 	$(MAKE) -C $(OPTEE_OS_PATH) $(OPTEE_OS_COMMON_FLAGS)
 
-OPTEE_OS_CLEAN_COMMON_FLAGS ?= $(OPTEE_OS_COMMON_EXTRA_FLAGS)
+OPTEE_OS_CLEAN_COMMON_FLAGS ?= \
```
jenswi-linaro, line 372: Why is this needed at all? Couldn't `OPTEE_OS_COMMON_FLAGS` be used instead?

etienne-lms: Agree. Could use `OPTEE_OS_COMMON_FLAGS`.

I don't known if anyone uses a local `OPTEE_OS_CLEAN_COMMON_FLAGS`. So I wonder if we should use
```makefile
OPTEE_OS_CLEAN_COMMON_FLAGS ?= $(OPTEE_OS_COMMON_FLAGS)
```
or use straight at line 380 below
```makefile
optee-os-clean-common: xtest-clean-common optee-examples-clean-common
	$(MAKE) -C $(OPTEE_OS_PATH) $(OPTEE_OS_COMMON_FLAGS) clean
```


Note: maybe some other simplification could be made on other `*_CLEAN_COMMON_FLAGS`.

jenswi-linaro: I'd prefer the latter and get rid of `OPTEE_OS_CLEAN_COMMON_FLAGS` completely.

> Note: maybe some other simplification could be made on other `*_CLEAN_COMMON_FLAGS`.

Not necessarily in scope here. Do as you like.

etienne-lms: thanks for the feedback

etienne-lms:
comment addressed

jenswi-linaro:

etienne-lms:
tag applied

PR #428, opened 2020-06-02
hikey: do not call hisi-idt.py with hardcoded python command

jforissier:
Let hisi-idt.py decide which python interpreter is used based on the
hashbang line. Calling "python" can cause compatibility problems when
"python" is Python 2 but the script needs Python3.


jbech-linaro:

PR #434, opened 2020-08-12
qemu-check.exp: clean assertion failure message

jforissier:
The message displayed by "make check" when an assertion failure is
detected is redundant and possibly misleading. For instance:

 $ PATH=~/work/clang-10.0.0/bin:$PATH make -j10 COMPILER=clang check
 [...]
 Starting QEMU... done, guest is booted.
 Running: xtest...
 '##########!!! TEE core assertion failed: 'assertion 'maps[map_idx].sz == sz' failed at ldelf/ta_elf.c:1351 in ta_elf_print_mappings()
 make: *** [Makefile:221: check] Error 1
 $

The "TEE core assertion failed:' part is not needed and possibly
misleading in this case (the assertion occured in ldelf which is
arguably not "TEE core"). Remove it.

### DIFF
--- a/qemu-check.exp
+++ b/qemu-check.exp
@@ -99,7 +99,7 @@ expect {
 	"+-----------------------------------------------------\r\r" {}
 	# Handle errors in TEE core output
 	-i $teecore -re {(assertion.*failed at.*)\n} {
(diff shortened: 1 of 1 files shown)


jforissier:
Thanks. I have slightly reworded the commit message to avoid a redundancy (!).

PR #435, opened 2020-08-12
fvp: update path to simulator

jenswi-linaro:
The path to the simulator when unpacked has changed in recent FVP
foundation model. Adjusting the path to match the latest version
11.11_34.

### DIFF
--- a/fvp.mk
+++ b/fvp.mk
@@ -198,7 +198,7 @@ run: all
 
 run-only:
 	@cd $(FOUNDATION_PATH); \
-	$(FOUNDATION_PATH)/models/Linux64_GCC-4.9/Foundation_Platform \
+	$(FOUNDATION_PATH)/models/Linux64_GCC-6.4/Foundation_Platform \
 	--arm-v8.0 \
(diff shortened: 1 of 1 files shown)


jenswi-linaro:
I'm not sure we care that much about older releases. There's a manual step anyway to make the model available, you could add a symlink or whatever to make the old model available.

jenswi-linaro:
Tag applied

jbech-linaro:
> I'm not sure we care that much about older releases. There's a manual step anyway to make the model available, you could add a symlink or whatever to make the old model available.

Probably pretty hard to even get hold of older "models". So I don't think it's super-important to consider that issue.

jforissier:
> > I'm not sure we care that much about older releases. There's a manual step anyway to make the model available, you could add a symlink or whatever to make the old model available.
> 
> Probably pretty hard to even get hold of older "models". So I don't think it's super-important to consider that issue.

Agreed in general, but for some people it might actually be illegal to download the newer version :wink:

PR #437, opened 2020-08-31
optee_test_ext.mk: make sure TAs are built in correct order

jforissier:
The optee_test TAs are built by a $(wildcard ...) loop. This is
unreliable because there is no guarantee that os_test_lib and
os_test_lib_dl will appear before os_test in the list. Yet os_test
depends on the libraries being built first.

Fix the issue by explicitely listing the libraries and TAs with
dependencies in the correct order, then using a wildcard for the other
TAs.


jenswi-linaro:

PR #441, opened 2020-11-10
Optional C++ and TLS tests

jforissier:
Relates to https://github.com/OP-TEE/optee_test/pull/464.


jenswi-linaro:
If we're building with this git, wouldn't it be reasonable to assume that the toolchains selected by this gits are used too?
If so we could make `WITH_TLS_TESTS` and `WITH_CPLUSPLUS_TESTS` `y` by default for all platforms supported here.

etienne-lms (commented):
Commit "Introduce BR2_PACKAGE_OPTEE_TEST_EXT_WITH_CPLUSPLUS_TESTS":
Buildroot uses `CXX_` to refer to C++ stuff: `BR2_PACKAGE_OPTEE_TEST_EXT_WITH_CXX_TESTS`?
Otherwise 
`Acked-by: Etienne Carriere <etienne.carriere@linaro.org>` for the whole series.

etienne-lms:
FYI, Buildroot has `BR2_INSTALL_LIBSTDCPP` that can help. With the help of the BR team, I recently [changed](https://patchwork.ozlabs.org/project/buildroot/patch/20201030075045.6960-6-etienne.carriere@linaro.org/) the BR upstream optee_test package on that dependency. I  forgot to check back our upstream.

jforissier:
@jenswi-linaro yes, fixed.
@etienne-lms `CXX` is better indeed, fixed. I think the `LIBSTDCPP` patch is not directly related so it should be applied separately.

jforissier:
Sorry I pushed garbage :-1: updated now.

jenswi-linaro:

PR #446, opened 2021-01-08
[RFC] br-ext: install PKCS11 TA upon CFG_PKCS11_TA=y

etienne-lms:
Add configuration switch BR2_PACKAGE_OPTEE_OS_EXT_PKCS11_TA in
optee_os_ext to install PKCS11 TA built from optee_os into target
filesystem (/lib/optee_armtz/xxx.ta). It is default enabled when
CFG_PKCS11_TA=y in OP-TEE common build.


jforissier:
The problem with this is, it won't scale very well. I'm thinking perhaps optee_os should have a target to install in-tree TAs. Such as `make install-ta DESTDIR=/path/to` would copy TAs to `/path/to/lib/optee_armtz`, and we could later make `/lib/optee_armtz` configurable if needed. Basically same usage as recommended by [GNU coding standards](https://www.gnu.org/prep/standards/html_node/DESTDIR.html).

jenswi-linaro (commented):

jenswi-linaro:
> The problem with this is, it won't scale very well. I'm thinking perhaps optee_os should have a target to install in-tree TAs. Such as `make install-ta DESTDIR=/path/to` would copy TAs to `/path/to/lib/optee_armtz`, and we could later make `/lib/optee_armtz` configurable if needed. Basically same usage as recommended by [GNU coding standards](https://www.gnu.org/prep/standards/html_node/DESTDIR.html).

I'm not so keen on installing TAs outside the out directory of the build. Perhaps we can put them in a special directory in the devkit?

jforissier:
> > The problem with this is, it won't scale very well. I'm thinking perhaps optee_os should have a target to install in-tree TAs. Such as `make install-ta DESTDIR=/path/to` would copy TAs to `/path/to/lib/optee_armtz`, and we could later make `/lib/optee_armtz` configurable if needed. Basically same usage as recommended by [GNU coding standards](https://www.gnu.org/prep/standards/html_node/DESTDIR.html).
> 
> I'm not so keen on installing TAs outside the out directory of the build. Perhaps we can put them in a special directory in the devkit?

The build would not install them by default, it would still build them in the out directory. It would be an additional step, which most packages usually have to install the binaries in their final location, or to a staging directory (when DESTDIR is set).

etienne-lms:
> Perhaps we can put them in a special directory in the devkit?

Yes, that would be another way. We already do that for TA shared libs built from optee_os.

jforissier:
> > Perhaps we can put them in a special directory in the devkit?
> 
> Yes, that would be another way. We already do that for TA shared libs built from optee_os.

True, that's a good argument :) The libs are under `lib/` which is not so good for in-tree TAs but we could perhaps create another directory. `bin/` perhaps?

etienne-lms:
`CFG_PKCS11_TA=y` could be used to export the ta in the devkit once built.

jenswi-linaro:
> > > Perhaps we can put them in a special directory in the devkit?
> > 
> > 
> > Yes, that would be another way. We already do that for TA shared libs built from optee_os.
> 
> True, that's a good argument :) The libs are under `lib/` which is not so good for in-tree TAs but we could perhaps create another directory. `bin/` perhaps?

Sounds good to me.

jforissier:
> > > > Perhaps we can put them in a special directory in the devkit?
> > > 
> > > 
> > > Yes, that would be another way. We already do that for TA shared libs built from optee_os.
> > 
> > 
> > True, that's a good argument :) The libs are under `lib/` which is not so good for in-tree TAs but we could perhaps create another directory. `bin/` perhaps?
> 
> Sounds good to me.

OK I have something working already, I'll post it soon.

etienne-lms:
ok @jforissier, i leave it to you.

jforissier:
@etienne-lms https://github.com/OP-TEE/optee_os/pull/4310. With that you should be able to simplify the br-ext pat quite a bit ;-)

etienne-lms:
fyi: i'll push a proposal.
Will install all TAs found in devkit, similarly to how shared libs .ta file as also installed.

An alternative would be to have a `BR2_OPTEE_OS_EXT_xxx` config switch to select target TAs. I think its over killing for this repo.

etienne-lms:
superseded by https://github.com/OP-TEE/build/pull/452

PR #448, opened 2021-01-12
rpi3: recommend 64MB for the boot partition

jbech-linaro:
(This doesn't matter whether it's merged right away of after the 3.12.0 release)

In some cases depending on configurations etc 32MB is a bit tight for
the boot partition. By increasing it to 64MB you are in a better
position for not hitting the limit. Likewise you can store additional
kernel versions in the boot partition which can be useful when debugging
issues showing up in one kernel build, but not the other. Because of
this the help text should state 64MB instead of 32MB.

### DIFF
--- a/rpi3.mk
+++ b/rpi3.mk
@@ -183,7 +183,7 @@ img-help:
 	@echo "   > p             # create primary"
 	@echo "   > 1             # make it the first partition"
 	@echo "   > <enter>       # use the default sector"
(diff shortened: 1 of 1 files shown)


jforissier:
The Apollo Guidance Computer had 72 kilobytes of ROM. It was enough to fly to the moon and back, more than fifty years ago. And now you're saying you need 64 **megabytes** to print a few text messages on a screen? :rofl:  

jbech-linaro:

PR #449, opened 2021-01-19
fix make error "/bin/sh: 1: Bad substitution"

liuyunli:
while use cmd " make -f toolchain.mk -j2" for download gcc on ubuntu machine, error occur "/bin/sh: 1: Bad substitution make:*** [toolchain.mk:39:aarch64] Error 2"
root reason is that sh of ubuntu point to dash, not bash. 
the code caused the error was "(cd $(1)/bin && for f in *-none-linux*; do ln -s $$f $${f//-none} ; done;) \"


toolchain.mk
```diff
@@ -1,6 +1,7 @@
 ################################################################################
 # Toolchains
 ################################################################################
+SHELL				:= /bin/bash
```
jforissier, line 4: Please use `=` instead of `:=` because there is no reason for `:=`.

jforissier (commented):
Look OK to me with one minor comment below, but note that this is not needed if you simply drop the `-f toolchain.mk`, in other words if invoke the main `Makefile` which does include `common.mk` which sets `SHELL = bash`.
Anyways it probably doesn't hurt to set the proper SHELL at the beginning of any makefile.

toolchain.mk
liuyunli: thanks @jforissier . take your advice

toolchain.mk
```diff
@@ -1,6 +1,7 @@
 ################################################################################
 # Toolchains
 ################################################################################
+SHELL				= /bin/bash
```
etienne-lms, line 4: maybe `?=` to inherit from `SHELL` is already defined?

jbech-linaro: > maybe `?=` to inherit from `SHELL` is already defined?

Not sure, lots of our stuff depends on bash. I.e., the shell isn't really an option for OP-TEE makefiles I guess.

jforissier: > > maybe `?=` to inherit from `SHELL` is already defined?
> 
> Not sure, lots of our stuff depends on bash. I.e., the shell isn't really an option for OP-TEE makefiles I guess.

Yes not a good idea, see https://www.gnu.org/software/make/manual/html_node/Choosing-the-Shell.html:
*"Unlike most variables, the variable SHELL is never set from the environment. This is because the SHELL environment variable is used to specify your personal choice of shell program for interactive use. It would be very bad for personal choices like this to affect the functioning of makefiles."*

etienne-lms: ok, understood. thanks.

liuyunli: So , do I need modify?

jbech-linaro: > So , do I need modify?

No, please keep it as it is, thanks!

etienne-lms: Sorry for the noise @liuyunli.

jforissier:
@liuyunli to merge this, I need you to:

- Squash the two commits into a single one. Change the subject to: `toolchain.mk: set SHELL = /bin/bash`
- Do not use Chinese symbols in the Git Author field, they are a bit too hard for westerners to read ;-) We need a real name in latin letters, preferably in order "First name, family name". I think a valid value for you would be `Yunli Liu <994605959@qq.com>` but feel free to adjust.
- Add your `Signed-off-by: Yunli Liu <994605959@qq.com>`
- Add also my `Acked-by: Jerome Forissier <jerome@forissier.org>`

Thanks.

liuyunli:
@jforissier  Thank you very much . Please check again.

jforissier:
@liuyunli perfect, thank you.

PR #450, opened 2021-01-21
qemu_v8: use Python 3 when building EDK2

jforissier:
Set PYTHON3_ENABLE=TRUE so that EDK2 is built with Python 3. Python 2 is
deprecated. Requires EDK2 tag edk2-stable201905 or later.

### DIFF
--- a/qemu_v8.mk
+++ b/qemu_v8.mk
@@ -122,7 +122,7 @@ qemu-clean:
 # EDK2 / Tianocore
 ################################################################################
 define edk2-env
-	export WORKSPACE=$(EDK2_PATH)
+	export WORKSPACE=$(EDK2_PATH) PYTHON3_ENABLE=TRUE
 endef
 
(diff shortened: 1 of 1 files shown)


jforissier:
Depends on https://github.com/OP-TEE/manifest/pull/177.

jbech-linaro:

jbech-linaro:
Have you considered doing the same changes for other platforms using EDK2?

jforissier:
> Have you considered doing the same changes for other platforms using EDK2?

That would be nice indeed but a little bit trickier perhaps, because:
- HiKey and HiKey960 use their own fork of EDK2
- FVP is on upstream but also imports `edk2-platforms` so I assume there is some kind of synchronization to keep between `edk2` and `edk2-platforms`, might not work on first try
- Same for Synquacer, with one more project: `edk2-non-osi`.

PR #452, opened 2021-01-25
br-ext: install OP-TEE devkit TAs

etienne-lms:
Install all TAs found in devkit into target filesystem.

### DIFF
--- a/br-ext/package/optee_os_ext/optee_os_ext.mk
+++ b/br-ext/package/optee_os_ext/optee_os_ext.mk
@@ -4,7 +4,7 @@ OPTEE_OS_EXT_SITE = $(BR2_PACKAGE_OPTEE_OS_EXT_SITE)
 OPTEE_OS_EXT_SITE_METHOD = local
 OPTEE_OS_EXT_SDK = $(BR2_PACKAGE_OPTEE_OS_EXT_SDK)
 
-define OPTEE_OS_EXT_INSTALL_OPTEE_OS_SHLIBS
+define OPTEE_OS_EXT_INSTALL_DEVKIT_SHLIBS
 	@mkdir -p $(TARGET_DIR)/lib/optee_armtz && \
 		for f in $(OPTEE_OS_EXT_SDK)/lib/*.ta; do \
 			[ -f "$$f" ] || continue; \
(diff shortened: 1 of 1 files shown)


jforissier:
Perhaps mention that it is useful for PKCS11? Anyway, looks good:

@jbech-linaro when this is merged there is no need for `CFG_IN_TREE_EARLY_TAS=pkcs11/fd02c9da-306c-48c7-a49c-bbd827ae86ee` in https://github.com/jbech-linaro/ibart/commit/38462e7c58eab49d37e7e766eb838307e1d7ae17. Only `CFG_PKCS11_TA=y` would be needed. But things can also be left as-is since it is a way to exercise the early TA code.

jbech-linaro:
> @jbech-linaro when this is merged there is no need for ...

OK, I'll let it be as it is since as you're saying it'll exercise early TA code. We can always change if we change our mind later.

etienne-lms:
> Perhaps mention that it is useful for PKCS11? Anyway, looks good:

I could but this is not specific the PKCS11 TA unless what I would have preferred only this TA gets installed.
Note I considered this build repo is dedicated to testing optee, not building some kind of product. Will all other platforms agree to embed optee_os intree TAs from this build env?

> But things can also be left as-is since it is a way to exercise the early TA code.

Agree. Nice to have an early TA tested in the CI.

etienne-lms:
Supersedes https://github.com/OP-TEE/build/pull/446.

jforissier:
> > Perhaps mention that it is useful for PKCS11? Anyway, looks good:
> 
> I could but this is not specific the PKCS11 TA 

No but it still answers the question "why this commit and what if I don't have it?".

> unless what I would have preferred only this TA gets installed.
> Note I considered this build repo is dedicated to testing optee,

It is.

> not building some kind of product. Will all other platforms agree to embed optee_os intree TAs from this build env?

I see no reason why it would be a problem. People building products will probably use the upstream Buildroot packages for OP-TEE, or some other building/packaging scripts (yocto for instance). If they use this as a reference, then they should expect to adjust things a bit.

etienne-lms:
ok. i'll add a ref to optee pkcs11 TA test env in the commit log.

etienne-lms:
commit log updated & tag applied.

PR #453, opened 2021-01-25
travis: remove the .travis.yml file

jbech-linaro:
In build.git we haven't been using Travis for almost 3 years and even if
we would enable it again, it would probably not work since it's
configured to run a very old Ubuntu release. We also stopped using this
as a full regression build testing since a) our build time had increased
and b) Travis had become more popular that prolonged the build time c)
we often hit the run-time limit for free accounts.

Because of this and to avoid confusion we should remove the Travis
configuration.


jforissier:

jbech-linaro:

PR #454, opened 2021-02-01
common.mk: qemu: do not build docs

jforissier:
Add the --disable-docs option to QEMU_CONFIGURE_PARAMS_COMMON so that no
documentation is generated during the QEMU build. Documentation is not
needed and it may fail to build in some environments. For instance on
Ubuntu 18.04.5 LTS:

 Warning, treated as error:
 docs/qemu-option-trace.rst.inc:4:Malformed option description '[enable=]PATTERN', should look like "opt", "-opt args", "--opt args", "/opt args" or "+opt args"
 make[1]: *** [.docs_built_system_qemu.1_docs_built_system_qemu-block-drivers.7_docs_built_system_qemu-cpu-models.7.sentinel.] Error 2

The error is caused by sphinx-build:

 $ sphinx-build --version
 sphinx-build 3.2.1

The version that comes with Ubuntu 20.04.1 LTS works fine:

 $ sphinx-build --version
 sphinx-build 1.8.5

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -370,7 +370,8 @@ edk2-clean-common:
 ################################################################################
 # QEMU / QEMUv8
 ################################################################################
-QEMU_CONFIGURE_PARAMS_COMMON = --cc="$(CCACHE)gcc" --extra-cflags="-Wno-error"
+QEMU_CONFIGURE_PARAMS_COMMON = --cc="$(CCACHE)gcc" --extra-cflags="-Wno-error" \
+			       --disable-docs
 QEMU_EXTRA_ARGS +=\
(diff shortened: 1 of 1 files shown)


jbech-linaro (commented):

jforissier:
Note that another option to fix the sphinx-build error is to upgrade to QEMU v5.2.0, which has fixed the issue in https://github.com/qemu/qemu/commit/5a4b0d4112 but it seems the build system has changed between v5.1.0 and v5.2.0 and the configure step takes quite a bit longer so we would need to change our Makefile to run it only when needed.

PR #456, opened 2021-03-02
qemu: enable initrd support

jenswi-linaro:
With 6edcf9dc2e1a ("efi/libstub: EFI_GENERIC_STUB_INITRD_CMDLINE_LOADER
should not default to yes") (included in v5.11) we need to set
EFI_GENERIC_STUB_INITRD_CMDLINE_LOADER=y explicitly in order to enable
initrd support.


jforissier:
Please write "included in **Linux** v5.11" in the commit log, because at first I thought it was a QEMU commit and had difficulties finding it ;-) although it should be obvious since you're changing a kernel `CONFIG_`...

jenswi-linaro:
Comment addressed and tag applied.

PR #457, opened 2021-03-03
Fix STM32MP157C-EV1 image generation

litchipi:
stm32mp1: Fix 157C-EV1 bootfs partition image name parameter

Genimage script configuration had wrong image name on bootfs partition. changed the parameter "image" to correct name for the image to be generated successfully and to allow successful boot on STM32MP157C-EV1 board.

Signed-off-by: Timothée Cercueil <litchi.pi@protonmail.com>
Signed-off-by: Timothée Cercueil <timothee.cercueil@st.com>
Acked-by: Jerome Forissier <jerome@forissier.org>

### DIFF
--- a/br-ext/board/stmicroelectronics/stm32mp1-tz/genimage-STM32MP157C-EV1.cfg
+++ b/br-ext/board/stmicroelectronics/stm32mp1-tz/genimage-STM32MP157C-EV1.cfg
@@ -34,7 +34,7 @@ image sdcard.img {
 	}
 
 	partition bootfs {
-		image = "rootfs.ext2"
+		image = "bootfs.ext2"
 		partition-type-uuid = L
 		size = 32M
 		bootable = "yes"


jforissier:
Hi @litchipi,
Thanks for the patch. You have two Signed-off-by tags, is it on purpose? Why? Do you want to provide two email addresses as your contact perhaps? This is a bit unusual, but that being said I see nothing wrong with it so I suppose you can leave them.

Other than that:

litchipi:
Hello,
As I am an intern for ST, my email may get deleted after my internship, this way you are able to contact me with my "real" one if needed. I keep signing with the other one to signify it is a contribution from ST, and because my local git is configured with the @st.com email.
That being said, I can keep only one

jforissier:
@litchipi that's a perfectly valid case for two S-o-b I think, let's keep them both.

etienne-lms:
Thanks Timothée,
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` assuming you update the commit log header line with what you updated in https://github.com/OP-TEE/build/pull/457#issue-583867151: 

`stm32mp1: Fix 157C-EV1 bootfs partition image name parameter`

etienne-lms (changes requested):
@litchipi, a line in commit log exceeds 80chars (74 max preferred i think in commit logs).
Please update:
```patch
 Genimage script configuration had wrong image name on bootfs partition.
-Changed the parameter "image" to correct name for the image to be generated successfully and to allow successful boot on STM32MP157C-EV1 board.
+Changed the parameter "image" to correct name for the image to be
+generated successfully and to allow successful boot on STM32MP157C-EV1
+board.
```

Issue #403, opened 2019-12-19
Unable to establish TCP connection with usermode networking


nageshshamnur:
Hi All,
   for my QEMU image, I am using the usermode networking (-netdev user,id=vmnic$(HOSTFWD) -device virtio-net-device,netdev=vmnic) as defined in common.mk and the QEMU come up and assign the IP address successfully. But however, if I run simple tcp_server on quest and tcp_client on the host using the server address as 127.0.0.1 with the port which is set in HOSTFWD, the packet doesn't seem to reach the TCP server and gets dropped at the guest. Any pointers would be helpful. I tried searching on optee links and google, but there was no solution to my problem.

becominginsane:
I'd like to remove the stale label and reopen the issue:
it is not currently possible to forward more ports in a straightforward manner.

I tried adding another network via netdev using the same device type, but the interface is down by default and when enabling it then it just won't connect.
I suggest changing [`HOSTFWD`](https://github.com/OP-TEE/build/blob/master/common.mk#L392) to append instead of simple assignment, like `QEMU_EXTRA_ARGS` does, so one can set `HOSTFWD` when running VM and put the forwarding they need!

jforissier:
Hi @becominginsane,

> I suggest changing [`HOSTFWD`](https://github.com/OP-TEE/build/blob/master/common.mk#L392) to append instead of simple assignment, like `QEMU_EXTRA_ARGS` does, so one can set `HOSTFWD` when running VM and put the forwarding they need!

That sounds like a good idea, would you like to create a PR for that?

PR #458, opened 2021-03-04
Add support for custom `HOSTFWD`

becominginsane:
It is alas not possible to use `+=` as it automatically adds a space which breaks the arguments up and results in an error from QEMU...

Refer to #403

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -389,7 +389,7 @@ endif
 endif
 
 ifeq ($(GDBSERVER),y)
-HOSTFWD := ,hostfwd=tcp::12345-:12345
+HOSTFWD := $(HOSTFWD),hostfwd=tcp::12345-:12345
 endif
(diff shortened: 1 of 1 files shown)


jforissier:
Thanks. Weird commit subject format (where did you pick `(feat)qemu` from? please look at the history), bad commit description, no Signed-off-by.

becominginsane:
@jforissier sorry that's my usual format taken from conventional commits...

forgot to add Signed-off-by, and the description was a placeholder honestly, I had expected it to be changed by you...

Check the commit now :)

jforissier:
@becominginsane that's much better now :) thank you. There is still one thing however: we [expect](https://optee.readthedocs.io/en/latest/general/contribute.html) contributors to sign-off with their real name, so I'm afraid we can't accept the pseudonym `becominginsane`. Would you mind changing this?

PR #461, opened 2021-03-15
qemu, qemuv8: avoid needless rebuilds

jforissier:
The qemu target currently runs the configure script unconditionally
which takes some time and causes unnecessary recompilations.

Split the configuration and the build steps and make the configuration
depend on $(QEMU_PATH)/config-host.mak to avoid the above issue.


jbech-linaro (commented):

PR #463, opened 2021-03-17
qemu_v8: fix error in edk2-clean

jforissier:
'make edk2-clean' fails with the following error:

 build.py...
  : error 000E: One Path in PACKAGES_PATH doesn't exist
         /home/jerome/work/optee_repo_qemu_v8/edk2/../edk2-platforms

The reason is $(ROOT)/edk2-platforms does not exist on this platform.
Fix this by using the proper platform-specific variable
$(EDK2_PLATFORMS_PATH) in the edk2-clean-common target like is done in
edk2-common.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -356,7 +356,7 @@ edk2-common:
 .PHONY: edk2-clean-common
 edk2-clean-common:
 	$(call edk2-env) && \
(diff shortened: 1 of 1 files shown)


jenswi-linaro:

PR #464, opened 2021-03-18
Update the EDK2 toolchain and QEMU invoking commands

xiangmy:
As discussed in https://github.com/OP-TEE/manifest/issues/179
+ Older EDK2 toolchains (GCC49) will cause errors when updating EDK2, therefore, we use GCC5 instead.
+ QEMU has adopted meson as the build system. As a result, the output of QEMU will be located in the $(QEMU_PATH)/build, so we should reflect this change.

### DIFF
--- a/qemu.mk
+++ b/qemu.mk
@@ -23,6 +23,7 @@ TF_A_PATH			?= $(ROOT)/trusted-firmware-a
 BINARIES_PATH			?= $(ROOT)/out/bin
 U-BOOT_PATH			?= $(ROOT)/u-boot
 QEMU_PATH			?= $(ROOT)/qemu
+QEMU_BUILD			?= $(QEMU_PATH)/build
 SOC_TERM_PATH			?= $(ROOT)/soc_term
 
 DEBUG = 1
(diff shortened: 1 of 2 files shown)


qemu.mk
```diff
@@ -195,7 +195,7 @@ endif
 check: $(CHECK_DEPS)
 	ln -sf $(ROOT)/out-br/images/rootfs.cpio.gz $(BINARIES_PATH)/
 	cd $(BINARIES_PATH) && \
-		export QEMU=$(ROOT)/qemu/arm-softmmu/qemu-system-arm && \
+		export QEMU=$(ROOT)/build/qemu/arm-softmmu/qemu-system-arm && \
```
jforissier, line 198: `$(QEMU_PATH)` not `$(ROOT)`

jforissier (changes requested):
Thanks for the patch. Please create one commit for the EDK2 update and one for QEMU and see below.

qemu.mk
xiangmy: Fixed, thanks!

jforissier: And you forgot to update `$(QEMU_PATH)/config-host.mak:`.
I suggest you define `QEMU_BUILD ?= $(QEMU_PATH)/build` just after `QEMU_PATH` then use that where needed.

jforissier:
For commit "Update the EDK2 toolchain":
`Tested-by: Jerome Forissier <jerome@forissier.org> (QEMUv8)`

qemu.mk
xiangmy: Thanks for the advice! I've made the changes.

jforissier:
For commit "Update QEMU invoking commands":
Please add prefix: "qemu, qemu_v8: update QEMU invoking commands". With that:

Please also add my review/test tags.

xiangmy:
Made the changes, thanks!

jforissier (approved):

PR #466, opened 2021-04-08
common: remove deprecated dependency for optee-os-clean-common

etienne-lms:
Remove deprecated dependencies on target optee-os-clean-common. The targets where removed by commit 819066fb88a6.

Fixes build failure with trace like:
```bash
  bash> make clean -j$(nproc)
  (...)
  make: *** No rule to make target 'xtest-clean-common', needed by 'optee-os-clean-common'.  Stop.
```

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -448,8 +448,5 @@ optee-os-common:
 	$(MAKE) -C $(OPTEE_OS_PATH) $(OPTEE_OS_COMMON_FLAGS)
 
 .PHONY: optee-os-clean-common
-ifeq ($(CFG_TEE_BENCHMARK),y)
-optee-os-clean-common: benchmark-app-clean-common
-endif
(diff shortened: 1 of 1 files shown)


jforissier:
Oops! Yes. I must admit I almost never us `make clean` because I often want to clean only optee_os and the optee buildroot stuff. That is easily done with `rm -rf ../out-br/build/optee* ../optee_os/out`...

jbech-linaro:
> Oops! Yes. I must admit I almost never us `make clean`

I **never** use make clean. Since I started to use git ages ago it has been `git clean -xdf` that has been my way of getting back inside a git (and for OP-TEE just as you, rm -rf the out stuff). My "make clean" rules in my own makefiles actually is a call to git clean. Abuse or not ... that's the way how I do it :-)

etienne-lms:
tag applied

PR #472, opened 2021-04-27
rpi3: remove rpi3/firmware/head.S

jforissier:
head.S is not used since commit 2e947f2ee5ae ("rpi3: use upstream ATF
and U-boot"). Remove it, as well as RPI3_HEAD_BIN which was the path to
the binary file after compilation.


jforissier:
Ping?

jenswi-linaro:

PR #474, opened 2021-04-28
Add Qemu Linux module building and kernel gdb python helpers

petegriffin:
These few patches extend optee-build to make it easier to build and debug Linux kernel modules using GDB and Qemu. It introduces a new option QEMU_GDB_ENABLE=y that disables kaslr when booting the qemu kernel.

e.g.

make QEMU_VIRTFS_ENABLE=y QEMU_USERNET_ENABLE=y QEMU_GDB_ENABLE=y run-only

and 

gdb-multiarch vmlinux
target remote :1234

Note add-auto-load-safe-path /path/to/optee-build/linux/scripts/gdb/vmlinux-gdb.py needs to be added to ~/.gdbinit for the python helpers to be accessible.

Afterwards lx-symbols, lx-dmesg, lx-fdtdump etc can be used, which are particularly useful if the kernel has failed in early bootup.


qemu_v8.mk
```diff
@@ -152,6 +154,8 @@ linux-defconfig: $(LINUX_PATH)/.config
 LINUX_COMMON_FLAGS += ARCH=arm64 Image
 
 linux: linux-common
+	$(MAKE) -C $(LINUX_PATH) $(LINUX_COMMON_FLAGS) modules
+	$(MAKE) -C $(LINUX_PATH) $(LINUX_COMMON_FLAGS) modules INSTALL_MOD_STRIP=1 INSTALL_MOD_PATH=$(MODULE_OUTPUT) modules_install
```
jforissier, line 158: - `make modules` twice?
- This will significantly increase the build time for everyone, including people who don't need any module (the `optee_test` suite runs fine without any module). Could we find a way to address this? Perhaps add a new target?

common.mk
```diff
@@ -381,6 +381,10 @@ QEMU_EXTRA_ARGS +=\
 	  -device virtio-9p-device,fsdev=fsdev1,mount_tag=secure
 endif
 endif
+ifeq ($(QEMU_GDB_ENABLE),y)
+# disable kaslr of the kernel to get symbolic debugging with gdb
+QEMU_GDB_BOOTARGS ?= nokaslr
+endif
```
jforissier, line 387: Why introduce two config variables? It seems `QEMU_GDB_BOOTARGS` is enough. And it had better be named `QEMU_KERNEL_BOOTARGS` I think. And I don't think it belongs in `common.mk`, simply appending it to the kernel command line in `qemu_v8.mk` should be OK.

jforissier (commented):
- For commit " qemu: Add CONFIG_GDB_SCRIPTS=y to defconfig":
s/instrospection/introspection/
- For commit " qemu: avoid short-form boolean deprecated warning":
- For commit "qemu_v8.mk: add QEMU_GDB_BOOTARGS to kernel bootargs"
With s/QEMU_GDB_BOOTARGS/QEMU_KERNEL_BOOTARGS/ or something that makes it clear we're talking about the kernel args:

For other commits please see below.

qemu_v8.mk
petegriffin: Thanks for your review :) In v2 I have removed the duplicated modules, and split it into a separate linux-modules target as you suggested.

common.mk
petegriffin: My reasoning for putting it in common.mk was mainly because a bunch of other qemu config stuff was there. But I just noticed QEMU_SMP ?= 2 is also set in qemu_v8.mk.

In v2 I have done as you suggest and renamed to QEMU_KERNEL_BOOTARGS, and moved it to qemu_v8.mk. I've left the default as nothing (as maybe turning off kaslr by default isn't what we want). But it can be passed through in the following way from cmd line if you want to debug the kernel with gdb (and same mechanism could be used for other useful debug kernel bootargs during development)

make QEMU_KERNEL_BOOTARGS=nokaslr run-only

qemu_v8.mk
```diff
@@ -40,6 +40,8 @@ EDK2_BIN		?= $(EDK2_PATH)/Build/ArmVirtQemuKernel-$(EDK2_ARCH)/$(EDK2_BUILD)_$(E
 QEMU_PATH		?= $(ROOT)/qemu
 QEMU_BUILD		?= $(QEMU_PATH)/build
 SOC_TERM_PATH		?= $(ROOT)/soc_term
+MODULE_OUTPUT		?= $(ROOT)/module_output
```
jforissier, line 43: I would prefer `$(ROOT)/out/kernel_modules` (note that `$(ROOT)/out` already exists).

qemu_v8.mk
```diff
@@ -199,6 +199,8 @@ run: all
 
 QEMU_SMP ?= 2
 
+QEMU_KERNEL_BOOTARGS ?=
+
```
jforissier, line 203: This does nothing, please remove

qemu_v8.mk
```diff
@@ -155,6 +157,10 @@ linux: linux-common
 	mkdir -p $(BINARIES_PATH)
 	ln -sf $(LINUX_PATH)/arch/arm64/boot/Image $(BINARIES_PATH)
 
+linux-modules: linux
+	$(MAKE) -C $(LINUX_PATH) $(LINUX_COMMON_FLAGS) modules
+	$(MAKE) -C $(LINUX_PATH) $(LINUX_COMMON_FLAGS) INSTALL_MOD_STRIP=1 INSTALL_MOD_PATH=$(MODULE_OUTPUT) modules_install
```
jforissier, line 163: It is slightly annoying to have `Image` in `$(LINUX_COMMON_FLAGS)` because it is repeated, but that's OK.

jforissier (commented):
Thanks for the update. It is true that things could be moved to `common.mk` but it can be done later when someone actually needs it and tests it.
With my comment addressed:

qemu_v8.mk
petegriffin: will do

qemu_v8.mk
petegriffin: sure thing, will update in v3

jforissier:
```
 qemu_v8: Add linux-modules target for building kernel modules

To avoid initramfs becoming huge with modules they
can be mounted from host into the guest filesystem.

(host)
make QEMU_VIRTFS_ENABLE=y QEMU_USERNET_ENABLE=y run-only
```
Perhaps you could mention `make linux-modules` here?

petegriffin:
> ```
>  qemu_v8: Add linux-modules target for building kernel modules
> 
> To avoid initramfs becoming huge with modules they
> can be mounted from host into the guest filesystem.
> 
> (host)
> make QEMU_VIRTFS_ENABLE=y QEMU_USERNET_ENABLE=y run-only
> ```
> 
> Perhaps you could mention `make linux-modules` here?

Good idea, will update now

PR #482, opened 2021-05-15
toolchain: check host architecture before downloading

mefistotelis:
Official opt-tee builds are only supported on 64-bit x86
machines. This change will make the tool error out if
someone tries to build it on different architecture without
providing name of own toolchain.

Signed-off-by: Mefistotelis <mefistotelis@gmail.com>

Related ticket:
https://github.com/OP-TEE/optee_docs/issues/112

off-topic: the auto-comment visible on PR creation contains outdated link to guidelines.

### DIFF
--- a/toolchain.mk
+++ b/toolchain.mk
@@ -4,15 +4,28 @@
 SHELL				= /bin/bash
 ROOT				?= $(CURDIR)/..
 TOOLCHAIN_ROOT 			?= $(ROOT)/toolchains
+UNAME_M				:= $(shell uname -m)
 
 AARCH32_PATH 			?= $(TOOLCHAIN_ROOT)/aarch32
 AARCH32_CROSS_COMPILE 		?= $(AARCH32_PATH)/bin/arm-linux-gnueabihf-
-AARCH32_GCC_VERSION 		?= gcc-arm-9.2-2019.12-x86_64-arm-none-linux-gnueabihf
-SRC_AARCH32_GCC 		?= https://developer.arm.com/-/media/Files/downloads/gnu-a/9.2-2019.12/binrel/$(AARCH32_GCC_VERSION).tar.xz
 
 AARCH64_PATH 			?= $(TOOLCHAIN_ROOT)/aarch64
 AARCH64_CROSS_COMPILE 		?= $(AARCH64_PATH)/bin/aarch64-linux-gnu-
-AARCH64_GCC_VERSION 		?= gcc-arm-9.2-2019.12-x86_64-aarch64-none-linux-gnu
+
+ifeq ($(UNAME_M),x86_64)
+AARCH32_GCC_VERSION		?= gcc-arm-9.2-2019.12-x86_64-arm-none-linux-gnueabihf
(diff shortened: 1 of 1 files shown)


jenswi-linaro:
Please give your real name in the Signed-off-by.
https://optee.readthedocs.io/en/latest/general/contribute.html#developer-certificate-of-origin

toolchain.mk
```diff
@@ -4,15 +4,28 @@
...
+AARCH64_GCC_VERSION		?= gcc-arm-9.2-2019.12-x86_64-aarch64-none-linux-gnu
+endif
+
+ifeq ($(AARCH32_GCC_VERSION),)
+$(error Host architecture $(UNAME_M) is not officially supported, and custom 32-bit toolchain was not defined)
+endif
+
+ifeq ($(AARCH64_GCC_VERSION),)
+$(error Host architecture $(UNAME_M) is not officially supported, and custom 64-bit toolchain was not defined)
+endif
```
etienne-lms, line 26: it would be nice this is not mandated.
aarch64 toolchain is not required when we build 32bit platforms.

for consistency, when building 64bit platforms, aarhc32 cross-toolchain may not be mandated.

etienne-lms:
In commit log, could you fix `opt-tee` to `OP-TEE`?

toolchain.mk
mefistotelis: Not enforcing both toolchains would be a deviation from current behavior. Good idea, but I don't think it should be a part of this change.
I actually don't like the shell script execution from within makefile; everything should be made using make targets. But again, this is a distinct subject, unrelated to this patch.

mefistotelis:
> Please give your real name in the Signed-off-by.

Agreed.

> In commit log, could you fix `opt-tee` to `OP-TEE`?

Agreed.

toolchain.mk
etienne-lms: you're right. sorry. indeed `make toolchains` currently gets both.
discard my comment.

mefistotelis:
Patch updated.

jforissier:

As a side note, and for those who may be wondering it is possible to build full Aarch64 versions of `optee_os`, `optee_client` and `optee_test` natively on an Arm server by setting CROSS_COMPILE{,64} to an empty value.

jenswi-linaro:
How about buildroot. Have you managed that too?
I'm trying to build the entire repo setup on a AArch64 machine but I've run into a few errors.
I thought I would try a shortcut by emulating a 64-bit cross compiler toolchain, but it's of course not the right thing to do and it also didn't quite work. I'll try setting up CROSS_COMPILE{,64} as you suggested instead the next time I have some spare time.

jforissier:
> How about buildroot. Have you managed that too?

Nope. didn't try :-/ Also a few adjustments are definitely needed in `optee_os` to also build the 32-bit stuff (`gcc -mabi=ilp32` I would think), I just disabled them with `supported-ta-targets=ta_arm64`.

jforissier:
Obsoleted by #488. @mefistotelis feel free to comment and/or test in your own environment and report back any issue. Thanks.

PR #492, opened 2021-06-16
Add options to add in UBOOT

ruchi393:

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -28,6 +28,7 @@ ROOT ?= $(shell pwd)/..
 
 BUILD_PATH			?= $(ROOT)/build
 LINUX_PATH			?= $(ROOT)/linux
+UBOOT_PATH			?= $(ROOT)/u-boot
 OPTEE_OS_PATH			?= $(ROOT)/optee_os
 OPTEE_CLIENT_PATH		?= $(ROOT)/optee_client
 OPTEE_TEST_PATH			?= $(ROOT)/optee_test
--- a/kconfigs/u-boot_qemu_v8.conf
+++ b/kconfigs/u-boot_qemu_v8.conf
@@ -1,0 +1,3 @@
+CONFIG_SYS_TEXT_BASE=0x60000000
+CONFIG_BOOTCOMMAND="smhload uImage ${kernel_addr_r} && smhload rootfs.cpio.uboot ${ramdisk_addr_r} &&  setenv bootargs console=ttyAMA0,115200 earlyprintk=serial,ttyAMA0,115200 root=/dev/ram && bootm ${kernel_addr_r} ${ramdisk_addr_r} ${fdt_addr}"
+CONFIG_SEMIHOSTING=y
--- a/qemu_v8.mk
+++ b/qemu_v8.mk
@@ -23,6 +23,9 @@ include common.mk
 
 DEBUG ?= 1
 
+# Option to use U-Boot in the boot flow instead of EDK2
+UBOOT ?= n
+
 ################################################################################
 # Paths to git projects and various binaries
 ################################################################################
@@ -41,14 +44,53 @@ QEMU_PATH		?= $(ROOT)/qemu
 QEMU_BUILD		?= $(QEMU_PATH)/build
 SOC_TERM_PATH		?= $(ROOT)/soc_term
 MODULE_OUTPUT		?= $(ROOT)/out/kernel_modules
+UBOOT_PATH		?= $(ROOT)/u-boot
+UBOOT_BIN		?= $(UBOOT_PATH)/u-boot.bin
+MKIMAGE_PATH		?= $(UBOOT_PATH)/tools
 
+ROOTFS_GZ		?= $(BINARIES_PATH)/rootfs.cpio.gz
+ROOTFS_UGZ		?= $(BINARIES_PATH)/rootfs.cpio.uboot
+
+KERNEL_IMAGE		?= $(LINUX_PATH)/arch/arm64/boot/Image
+KERNEL_IMAGEGZ		?= $(LINUX_PATH)/arch/arm64/boot/Image.gz
+KERNEL_UIMAGE		?= $(BINARIES_PATH)/uImage
+
+# Load and entry addresses
+KERNEL_ENTRY		?= 0x40400000
+KERNEL_LOADADDR		?= 0x40400000
+ROOTFS_ENTRY		?= 0x44000000
+ROOTFS_LOADADDR		?= 0x44000000
+
+ifeq ($(UBOOT),y)
+BL33_BIN		?= $(UBOOT_BIN)
+BL33_DEPS		?= u-boot
+else
+BL33_BIN		?= $(EDK2_BIN)
+BL33_DEPS		?= edk2
+endif
 
 ################################################################################
 # Targets
 ################################################################################
-all: arm-tf buildroot edk2 linux optee-os qemu soc-term
-clean: arm-tf-clean buildroot-clean edk2-clean linux-clean optee-os-clean \
+TARGET_DEPS := arm-tf buildroot linux optee-os qemu soc-term
+TARGET_CLEAN := arm-tf-clean buildroot-clean linux-clean optee-os-clean \
 	qemu-clean soc-term-clean check-clean
+
+TARGET_DEPS 		+= $(BL33_DEPS)
+
+ifeq ($(UBOOT),y)
+TARGET_DEPS		+= $(KERNEL_UIMAGE) $(ROOTFS_UGZ)
+TARGET_CLEAN		+= u-boot-clean
+else
+TARGET_CLEAN		+= edk2-clean
+endif
+
+all: $(TARGET_DEPS)
+
+clean: $(TARGET_CLEAN)
+
+$(BINARIES_PATH):
+	mkdir -p $@
 
 include toolchain.mk
 
@@ -71,7 +113,7 @@ TF_A_FLAGS ?= \
 	BL32=$(OPTEE_OS_HEADER_V2_BIN) \
 	BL32_EXTRA1=$(OPTEE_OS_PAGER_V2_BIN) \
 	BL32_EXTRA2=$(OPTEE_OS_PAGEABLE_V2_BIN) \
-	BL33=$(EDK2_BIN) \
+	BL33=$(BL33_BIN) \
 	PLAT=qemu \
 	ARM_TSP_RAM_LOCATION=tdram \
 	BL32_RAM_LOCATION=tdram \
@@ -86,7 +128,7 @@ TF_A_FLAGS += \
 	GENERATE_COT=1
 endif
 
-arm-tf: optee-os edk2
+arm-tf: optee-os $(BL33_DEPS)
 	$(TF_A_EXPORTS) $(MAKE) -C $(TF_A_PATH) $(TF_A_FLAGS) all fip
 	mkdir -p $(BINARIES_PATH)
 	ln -sf $(TF_A_OUT)/bl1.bin $(BINARIES_PATH)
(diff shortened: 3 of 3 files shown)


ruchi393:
This PR is dependent on 
https://github.com/OP-TEE/manifest/pull/185

qemu_v8.mk
```diff
@@ -41,15 +47,62 @@ QEMU_PATH		?= $(ROOT)/qemu
...
+BL33_BIN		?= $(EDK2_BIN)
+BL33_DEPS		?= edk2
+endif
 
 ################################################################################
 # Targets
 ################################################################################
-all: arm-tf buildroot edk2 linux optee-os qemu soc-term
-clean: arm-tf-clean buildroot-clean edk2-clean linux-clean optee-os-clean \
+TARGET_DEPS ?= arm-tf buildroot linux optee-os qemu soc-term
```
jenswi-linaro, line 86: Does it really make sense to have this overridable?

Overall I think that most of the `?=` locations in these make files should have `:=` or `=` instead. There are only few variables which it makes sense to override. Fixing this is out of scope for this PR, but there's no need to make matters worse.

qemu_v8.mk
```diff
@@ -141,6 +195,37 @@ edk2: edk2-common
 
 edk2-clean: edk2-clean-common
 
+################################################################################
```
jenswi-linaro, line 198: This is perhaps something that could be moved to `common.mk` instead. There's quite a few platforms now compiling U-Boot using a somewhat similar pattern, probably due to copy paste + some small tweak in each case.

qemu_v8.mk
```diff
@@ -189,6 +274,35 @@ soc-term:
 soc-term-clean:
 	$(MAKE) -C $(SOC_TERM_PATH) clean
 
+################################################################################
+# mkimage - create images to be loaded by U-boot
+################################################################################
+# Without the objcopy, the uImage will be 10x bigger.
+uimage: linux $(BINARIES_PATH)
```
jenswi-linaro, line 281: Why not make the target `$(KERNEL_UIMAGE)`? It seems that `uimage` will always be out of date.

qemu_v8.mk
```diff
@@ -189,6 +274,35 @@ soc-term:
...
+	${AARCH64_CROSS_COMPILE}objcopy -O binary \
+					-R .note \
+					-R .comment \
+					-S $(LINUX_PATH)/vmlinux \
+					$(BINARIES_PATH)/linux.bin && \
```
jenswi-linaro, line 286: No need for `&& \` at the end.

qemu_v8.mk
```diff
@@ -141,6 +195,37 @@ edk2: edk2-common
...
+
+.PHONY: uboot-menuconfig
+uboot-menuconfig: uboot-defconfig
+	$(MAKE) -C $(UBOOT_PATH) \
+		CROSS_COMPILE="$(CCACHE)$(AARCH64_CROSS_COMPILE)" \
+		menuconfig
+
+.PHONY: uboot-clean
+uboot-clean:
+	cd $(UBOOT_PATH) && git clean -xdf
```
jenswi-linaro, line 223: U-Boot already have a `clean` target.

qemu_v8.mk
```diff
@@ -189,6 +274,35 @@ soc-term:
...
+	$(MKIMAGE_PATH)/mkimage -A arm64 \
+				-O linux \
+				-T kernel \
+				-C none \
+				-a $(KERNEL_LOADADDR) \
+				-e $(KERNEL_ENTRY) \
+				-n "Linux kernel" \
+				-d $(BINARIES_PATH)/linux.bin $(KERNEL_UIMAGE)
+
+urootfs: buildroot $(BINARIES_PATH)
```
jenswi-linaro, line 296: Shouldn't the target be ` $(ROOTFS_UGZ)`?

qemu_v8.mk
```diff
@@ -189,6 +274,35 @@ soc-term:
 soc-term-clean:
 	$(MAKE) -C $(SOC_TERM_PATH) clean
 
+################################################################################
+# mkimage - create images to be loaded by U-boot
+################################################################################
+# Without the objcopy, the uImage will be 10x bigger.
```
jenswi-linaro, line 280: Aha, good to know!

qemu_v8.mk
ruchi393: I will change TARGET_DEPS ?= to TARGET_DEPS :=

qemu_v8.mk
ruchi393: Ok, will do

qemu_v8.mk
ruchi393: I will remove that

jforissier:
Cosmetic issues...
- It would be nice to state how to enable U-Boot in the commit description of "qemu_v8.mk: Add option to use uboot instead of edk2". Adding "(make UBOOT=y)" at the end of the sentence would be good enough I think.
- U-Boot is spelt U-Boot, not uboot or UBOOT.

jforissier:
Is GICv3 needed for U-Boot support? I don't think so, therefore commit "qemu_v8.mk: Add option to compile with GICv3" probably shouldn't be in this PR right?

ruchi393:
> Is GICv3 needed for U-Boot support? I don't think so, therefore commit "qemu_v8.mk: Add option to compile with GICv3" probably shouldn't be in this PR right?

Yes, you are right. I will split the PR up.

ruchi393:
Fixed based on comments on U-Boot patch. 
Removed GICv3 patch from the PR.

common.mk
```diff
@@ -361,6 +362,26 @@ edk2-clean-common:
...
+uboot-defconfig: $(UBOOT_PATH)/.config
+
+.PHONY: uboot-common
+uboot-common: uboot-defconfig
+	$(MAKE) -C $(UBOOT_PATH) $(UBOOT_COMMON_FLAGS)
+
+.PHONY: uboot-clean-common
+uboot-clean-common:
+	$(MAKE) -C $(UBOOT_PATH) $(UBOOT_COMMON_FLAGS) distclean
+
```
jforissier, line 384: 1. I think all platforms should use these targets as soon as they are introduced. How bout one refactoring commit with the changes to `common.mk` and all platform files ("Move u-boot* targets to common.mk"), then one separate commit for the new QEMUv8 stuff?
2. Why `uboot*` when `qemu.mk` already uses the spelling `u-boot*`?

qemu_v8.mk
```diff
@@ -23,6 +23,9 @@ include common.mk
 
 DEBUG ?= 1
 
+# Option to use uboot in the boot flow instead of EDK2
```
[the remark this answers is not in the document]
ruchi393: Sorry, I missed to replace this one.

kconfigs/u-boot_qemu_v8.conf
```diff
@@ -0,0 +1,4 @@
+CONFIG_SYS_TEXT_BASE=0x60000000
+CONFIG_BOOTCOMMAND="smhload uImage ${kernel_addr_r} && smhload rootfs.cpio.uboot ${ramdisk_addr_r} &&  setenv bootargs console=ttyAMA0,115200 earlyprintk=serial,ttyAMA0,115200 root=/dev/ram && bootm ${kernel_addr_r} ${ramdisk_addr_r} ${fdt_addr}"
+CONFIG_SEMIHOSTING=y
+
```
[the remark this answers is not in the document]
ruchi393: Will remove

qemu_v8.mk
```diff
@@ -189,6 +243,35 @@ soc-term:
 soc-term-clean:
 	$(MAKE) -C $(SOC_TERM_PATH) clean
 
+################################################################################
+# mkimage - create images to be loaded by U-Boot
+################################################################################
+# Without the objcopy, the uImage will be 10x bigger.
+$(KERNEL_UIMAGE): linux $(BINARIES_PATH)
```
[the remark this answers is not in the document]
ruchi393: Ok

qemu_v8.mk
```diff
@@ -189,6 +243,35 @@ soc-term:
...
+	$(MKIMAGE_PATH)/mkimage -A arm64 \
+				-O linux \
+				-T kernel \
+				-C none \
+				-a $(KERNEL_LOADADDR) \
+				-e $(KERNEL_ENTRY) \
+				-n "Linux kernel" \
+				-d $(BINARIES_PATH)/linux.bin $(KERNEL_UIMAGE)
+
+$(ROOTFS_UGZ): buildroot $(BINARIES_PATH)
```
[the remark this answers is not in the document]
ruchi393: Ok

common.mk
ruchi393: > 1. I think all platforms should use these targets as soon as they are introduced. How about one refactoring commit with the changes to `common.mk` and all platform files ("Move u-boot* targets to common.mk"), then one separate commit for the new QEMUv8 stuff?

Can I just add u-boot target in QEMUv8 file for now as I did in earlier PR. I need this added to get the xen workflow added. Then I can work on a separate PR to move things to common.mk and fix all platforms. Would that work ?

> 2. Why `uboot*` when `qemu.mk` already uses the spelling `u-boot*`?

 No particular reason. I actually didn't see the qemu.mk. I can change to u-boot for consistency.

jforissier: > Then I can work on a separate PR to move things to common.mk and fix all platforms. Would that work ?

Yes, totally acceptable.

> No particular reason. I actually didn't see the qemu.mk. I can change to u-boot for consistency.

Yes, please. No big deal, TBH I would have chosen `uboot` too, but consistency is more important.

jbech-linaro: > > No particular reason. I actually didn't see the qemu.mk. I can change to u-boot for consistency.
> 
> Yes, please. No big deal, TBH I would have chosen `uboot` too, but consistency is more important

I could be the one to blame for that. I think I've tried to use the correct spelling. But if it was today, I probably would've went with `uboot` as well. (and I'm glad we use `optee_xyz` for gits etc and not `op-tee_xyx`).

ruchi393:
Added a fix commit to address all comments. Please help review.

qemu_v8.mk
```diff
@@ -41,15 +44,54 @@ QEMU_PATH		?= $(ROOT)/qemu
...
+TARGET_DEPS 		+= $(BL33_DEPS)
+
+ifeq ($(UBOOT),y)
+TARGET_DEPS		+= $(KERNEL_UIMAGE) $(ROOTFS_UGZ)
+TARGET_CLEAN		+= u-boot-clean
+else
+TARGET_CLEAN		+= edk2-clean
+endif
+
+all: $(TARGET_DEPS)
```
jenswi-linaro, line 88: These dependencies without actions can be repeated as many times as needed so there's generally not a need to do this trick with a variable. There's not much harm in doing like this either so keep this if you prefer that.

qemu_v8.mk
```diff
@@ -189,6 +256,35 @@ soc-term:
 soc-term-clean:
 	$(MAKE) -C $(SOC_TERM_PATH) clean
 
+################################################################################
+# mkimage - create images to be loaded by U-Boot
+################################################################################
+# Without the objcopy, the uImage will be 10x bigger.
+$(KERNEL_UIMAGE): linux | $(BINARIES_PATH)
```
jforissier, line 263: `$(KERNEL_UIMAGE): u-boot linux | $(BINARIES_PATH)`

(`mkimage` is generated by target `u-boot`)

qemu_v8.mk
```diff
@@ -189,6 +256,35 @@ soc-term:
...
+	$(MKIMAGE_PATH)/mkimage -A arm64 \
+				-O linux \
+				-T kernel \
+				-C none \
+				-a $(KERNEL_LOADADDR) \
+				-e $(KERNEL_ENTRY) \
+				-n "Linux kernel" \
+				-d $(BINARIES_PATH)/linux.bin $(KERNEL_UIMAGE)
+
+$(ROOTFS_UGZ): buildroot | $(BINARIES_PATH)
```
jforissier, line 278: `$(ROOTFS_UGZ): u-boot buildroot | $(BINARIES_PATH)`

(`mkimage` is generated by target `u-boot`)

jforissier (commented):
One issue I found when building with `make -j10`. With that fixed:

`Tested-by: Jerome Forissier <jerome@forissier.org>`

qemu_v8.mk
ruchi393: Thanks @jforissier , I will add the dependency

qemu_v8.mk
ruchi393: I just found it a little more readable hence did it this way.

ruchi393:
Patches squashed and tag applied

ruchi393:
@jforissier , @jenswi-linaro , thanks for the review. All comments have been addressed. Please have a look.

jforissier (approved):
LGTM

etienne-lms (commented):

ruchi393:
Tags applied

jenswi-linaro:

ruchi393:
Tags applied

PR #495, opened 2021-06-23
qemu_v8: add scripts_gdb to LINUX_COMMON_FLAGS

jforissier:
Commit c993ec995a9e ("qemu_v8.mk: add scripts_gdb kernel build target")
has added "$(MAKE) -C $(LINUX_PATH) $(LINUX_COMMON_FLAGS) scripts_gdb" to the
linux: recipe. Since linux: depends on linux-common which already runs
"$(MAKE) -C $(LINUX_PATH) $(LINUX_COMMON_FLAGS)", and since
$(LINUX_COMMON_FLAGS) contains "Image", the Image target is made twice.
It is more efficient to remove the added line and add scripts_gdb to
LINUX_COMMON_FLAGS instead.

### DIFF
--- a/qemu_v8.mk
+++ b/qemu_v8.mk
@@ -218,10 +218,9 @@ LINUX_DEFCONFIG_COMMON_FILES := \
 
 linux-defconfig: $(LINUX_PATH)/.config
 
-LINUX_COMMON_FLAGS += ARCH=arm64 Image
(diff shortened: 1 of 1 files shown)


jforissier:
CC @petegriffin

petegriffin:
Acked-by: Peter Griffin <peter.griffin@linaro.org>

PR #496, opened 2021-06-24
qemu_v8.mk: do not warn about unreadable kernel if XEN_BOOT != y

jforissier:
The readability check of the host kernel makes sense only when
XEN_BOOT=y, so avoid annoying users that are not concerned by this.


ruchi393:
`Acked-by: Ruchika Gupta <ruchika.gupta@linaro.org>`

Issue #478, opened 2021-05-07
QEMU_PSS_AUTOMOUNT seems to be only working for uid=1000


cgellner:
It seems that enabling persistent TEE storage for qemu (QEMU_PSS_AUTOMOUNT=y) will only work properly for the primary user of the PC (having uid=1000).
When I try as a different user (uid!=1000) it doesn't work out of the box.

In order to reproduce, please try to switch to a user not having uid=1000 and run
```mkdir -p /tmp/qemu-data-tee && make QEMU_PSS_AUTOMOUNT=y run``` (as described in common.mk)
inside qemu userspace execute optee_example_secure_storage.
For me it simply gets stuck.
```
# optee_example_secure_storage 
Prepare session with the TA
```
It seems this is due to tee-supplicant not being started automatically

In startup I see below logs
```
Running sysctl: OK
Saving random seed: random: dd: uninitialized urandom read (512 bytes read)
OK
Set permissions on /dev/tee*: OK
Set permissions on /dev/ion: OK
Create/set permissions on /data/tee: chown: /data/tee: Operation not permitted
chown: /data/tee: Operation not permitted
FAIL
Starting network: OK
Starting network (udhcpc): OK

Welcome to Buildroot, type root or test to login
buildroot login: root
# pidof tee-supplicant
```
When using strace I can see EPERM error for chown using UID/GID 1000 on file-handle below /tmp/qemu-data-tee
```
...
29645 fchownat(16, ".", 1000, 1000, AT_SYMLINK_NOFOLLOW <unfinished ...>
...
29645 <... fchownat resumed> )          = -1 EPERM (Operation not permitted)
...
```
Note: 1000 is the default uid of "tee"
```
# id tee
uid=1000(tee) gid=1000(tee) groups=1000(tee)
```

When I manually start tee-supplicant daemon it seems to be working.
At least secure-storage example could be executed.

jforissier:
Hi @cgellner,

That's an interesting finding. Perhaps `security_model=mapped-xattr` would help here (see https://qemu.readthedocs.io/en/latest/system/invocation.html `-fsdev local,...` parameter)?

cgellner:
Hi @jforissier,

thank you for the feedback

I did a quick test changing security_model for QEMU_PSS... to mapped-xattr in common.mk.
Your proposal seems to work.
- tee-supplicant is started automatically
- optee_example_secure_storage is executed without any obvious issues
- xtest did pass

[...]

PR #498, opened 2021-06-25
Use security_model mapped-xattr for QEMU Persistent Secure Storage

cgellner:
Use security_model mapped-xattr for QEMU_PSS_HOST_DIR.
This allows folders/files below /data/tee to be owned by
any uid/gid of the QEMU context.

Fixes https://github.com/OP-TEE/build/issues/478

Signed-off-by: Christoph Gellner <cgellner@de.adit-jv.com>
Suggested-by: Jerome Forissier <jerome@forissier.org>

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -425,7 +425,7 @@ QEMU_EXTRA_ARGS +=\
 	-device virtio-9p-device,fsdev=fsdev0,mount_tag=host
 ifeq ($(QEMU_PSS_ENABLE),y)
 QEMU_EXTRA_ARGS +=\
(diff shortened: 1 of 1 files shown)


cgellner:
I performed some short tests on QEMU V8.
Now persistent secure storage is working correctly even if uid isn't 1000.
I can see that objects#2 created by `optee_example_secure_storage` in one run is also available in the next.
Calling `optee_example_secure_storage` in second run will delete it as expected.

`xtest -l 15` didn't show any issues

jforissier:
@cgellner awesome! You may add:

Thanks.

PR #500, opened 2021-06-29
juno: download scp version 2.8

jenswi-linaro:
TF-A v2.4 and earlier is used with the SCP binaries provided in the
vexpress-firmware git. With TF-A v2.5 is supposed to use the SCP
binaries released at [1] instead. So update to download the SCP binaries
from [1] instead.

Link [1]: https://downloads.trustedfirmware.org/tf-a/css_scp_2.8.0/juno

### DIFF
--- a/juno.mk
+++ b/juno.mk
@@ -20,13 +20,43 @@ TF_A_PATH		?= $(ROOT)/trusted-firmware-a
 U-BOOT_PATH		?= $(ROOT)/u-boot
 U-BOOT_BIN		?= $(U-BOOT_PATH)/u-boot.bin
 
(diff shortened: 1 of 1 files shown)


jforissier:

jenswi-linaro:
Depends on https://github.com/OP-TEE/manifest/pull/189

jenswi-linaro:
Tag applied

jenswi-linaro:
Since OP-TEE/manifest#189 (now merged) depends on this we should merge this soon.

jforissier:
Haha I literally merged 9 seconds after your message we just checked this at the same time :)

PR #506, opened 2021-07-14
qemu: kernel: Enable Trusted Keys support

b49020:
Also, include keyutils package in rootfs to test Trusted Keys features.


jforissier:
Could we have trusted keys tested by "make check"? That would fully justify having the new components installed by defaut (in addition to being a good thing in itself)

b49020:
> Could we have trusted keys tested by "make check"?

Sounds good to me. Will add corresponding support.

PR #512, opened 2021-08-03
qemu, qemu_v8: Use telnet instead of custom soc_term

b49020:
With this change, Qemu is the one that creates tcp ports and telnet will
listen on the created ports which is the opposite case when custom
soc_term application was used.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -453,22 +453,19 @@ endef
 
 ifneq (, $(LAUNCH_TERMINAL))
 define launch-terminal
-	@nc -z  127.0.0.1 $(1) || \
-		$(LAUNCH_TERMINAL) "$(SOC_TERM_PATH)/soc_term $(1)" &
+	$(LAUNCH_TERMINAL) "telnet localhost $(1)"
 endef
 else
 gnome-terminal := $(shell command -v gnome-terminal 2>/dev/null)
 xterm := $(shell command -v xterm 2>/dev/null)
 ifdef gnome-terminal
 define launch-terminal
-	@nc -z  127.0.0.1 $(1) || \
-	$(gnome-terminal) -x $(SOC_TERM_PATH)/soc_term $(1) &
+	$(gnome-terminal) -- telnet localhost $(1)
 endef
 else
 ifdef xterm
 define launch-terminal
-	@nc -z  127.0.0.1 $(1) || \
-	$(xterm) -title $(2) -e $(BASH) -c "$(SOC_TERM_PATH)/soc_term $(1)" &
+	$(xterm) -title $(2) -e $(BASH) -c "telnet localhost $(1)"
 endef
 else
 check-terminal := @echo "Error: could not find gnome-terminal nor xterm" ; false
@@ -476,8 +473,9 @@ endif
 endif
 endif
 
-define wait-for-ports
-	@while ! nc -z 127.0.0.1 $(1) || ! nc -z 127.0.0.1 $(2); do sleep 1; done
+define launch-terminal-wait
+	@while ! nc -z localhost $(1); do sleep 1; done && \
+	$(call launch-terminal,$(1),$(2)) &
 endef
 
 ################################################################################
--- a/qemu.mk
+++ b/qemu.mk
@@ -24,16 +24,15 @@ BINARIES_PATH			?= $(ROOT)/out/bin
 U-BOOT_PATH			?= $(ROOT)/u-boot
 QEMU_PATH			?= $(ROOT)/qemu
 QEMU_BUILD			?= $(QEMU_PATH)/build
-SOC_TERM_PATH			?= $(ROOT)/soc_term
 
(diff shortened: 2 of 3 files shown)


jforissier:
> With this change, Qemu is the one that creates tcp ports and telnet will
> listen on the created ports which is the opposite case when custom
> soc_term application was used.
 
Ha. That's annoying, because it doesn't work too well with my typical workflow. I use `tmux` with pre-configured windows and panes, everything is saved and restored easily when I reboot the laptop. I have `soc_term` running continuously in two panes. When I run `make run` or `make run-only` in another pane, QEMU will always connect the running `soc_term` instances.
With this patch, this doesn't work anymore and new terminals are spawned instead. I'm thinking perhaps it was the reason for writing `soc_term` in the first place? @jenswi-linaro

jbech-linaro:
FWIW, I have a similar workflow as Jerome.

jforissier:
> FWIW, I have a similar workflow as Jerome.

Yeah the opposite actually, I remember I learned about `tmux` from you :wink:

jbech-linaro:
> > FWIW, I have a similar workflow as Jerome.
> 
> Yeah the opposite actually, I remember I learned about `tmux` from you

I've changed slightly to https://www.byobu.org (which piggy-back on tmux IIRC). Same same, but different. :) So my journey has been `screen` to `tmux` to `byobu`.

b49020:
> I use tmux with pre-configured windows and panes, everything is saved and restored easily when I reboot the laptop. I have soc_term running continuously in two panes.

With telnet, you should run following in those pre-configured panes:

```
while true; do while ! nc -z localhost 54320; do sleep 1; done && telnet localhost 54320; done
while true; do while ! nc -z localhost 54321; do sleep 1; done && telnet localhost 54321; done
```
And we should be able to make terminal launching optional in this case as well:

```
make run LAUNCH_TERMINAL=0
make run-only LAUNCH_TERMINAL=0
```
Although it will be a little one time extra setup in your environment in case we switch to telnet but it maybe worth using `telnet` given the maintenance burden for custom `soc_term`.

jbech-linaro:
> given the maintenance burden for custom `soc_term`.

7 commits since 2014, so I'm not sure it's a maintenance burden :)

jforissier:
> > I use tmux with pre-configured windows and panes, everything is saved and restored easily when I reboot the laptop. I have soc_term running continuously in two panes.
> 
> With telnet, you should run following in those pre-configured panes:
> 
> ```
> while true; do while ! nc -z localhost 54320; do sleep 1; done && telnet localhost 54320; done
> while true; do while ! nc -z localhost 54321; do sleep 1; done && telnet localhost 54321; done
> ```

That would need to be provided as a script for convenience, e.g., `./soc_term.sh 54320`.

> And we should be able to make terminal launching optional in this case as well:
> 
> ```
> make run LAUNCH_TERMINAL=0
> make run-only LAUNCH_TERMINAL=0
>  ```

OK.

Another perhaps better option could be to rewrite soc_term in Python and add it to build.git? Should be almost trivial, no? Then we could get rid of the separate `soc_term` git and associated build step without changing the flow too much. Python is already required for many things.

b49020:
There is also usability benefit with `telnet` as well. Say a user posses access to a remote build machine via `ssh` or `telnet`. So with `telnet` supported in OP-TEE Qemu build, one can just run the build on that remote machine and access consoles via `telnet client` on local PC. Although it should be possible to have a GUI terminal window over `ssh` but it may be too slow for many users with limited network bandwidth. I am not sure if it is possible with `soc_term`.

> That would need to be provided as a script for convenience, e.g., ./soc_term.sh 54320.

Agree, that can be added as well.

jbech-linaro:
FWIW, I have no strong opinion about this. Old way works, but this proposal would also work.

jforissier:
> There is also usability benefit with `telnet` as well. Say a user posses access to a remote build machine via `ssh` or `telnet`. So with `telnet` supported in OP-TEE Qemu build, one can just run the build on that remote machine and access consoles via `telnet client` on local PC. Although it should be possible to have a GUI terminal window over `ssh` but it may be too slow for many users with limited network bandwidth. I am not sure if it is possible with `soc_term`.

No need for a GUI terminal window, a text session should be fine. ssh to the remote host then run `soc_term`. In one command: `ssh user@host /path/to/soc_term 54320`.

TBH, the only benefit I see with the `telnet` approach is to avoid building `soc_term` program, but it is otherwise less convenient IMO due to the client/server inversion which doesn't fit well the development/debugging workflow. And, introducing polling is a moderately elegant way of fixing the issue. So I am still thinking a conversion of `soc_term` to Python might be the best option. Comments welcome, and I may even give it a shot ;)

jenswi-linaro:
What's the problem with the C implementation? That it's in a separate git?

jenswi-linaro:
By the way, this proposal would break my workflow too.

PR #516, opened 2021-08-11
qemu, qemu_v8: set msize when mounting host filesystem

jforissier:
When VIRTFS_AUTOMOUNT=y and/or PSS_AUTOMOUNT=y, QEMU prints the
following warning:

 qemu-system-aarch64: warning: 9p: degraded performance: a reasonable
 high msize should be chosen on client/guest side (chosen msize is <=
 8192). See https://wiki.qemu.org/Documentation/9psetup#msize for
 details.

Let's set the buffer size to 64K mainly to silence this warning. The
performance is noticeably increased too although it doesn't really
matters for typical OP-TEE use cases.


etienne-lms:

PR #517, opened 2021-08-13
Xen tools version check

jforissier:
A patch to check that Xen and Xen tools have compatible version numbers, plus a minor cleanup.

### DIFF
--- a/qemu_v8.mk
+++ b/qemu_v8.mk
@@ -346,6 +346,19 @@ xen-create-image: xen linux buildroot | $(XEN_TMP)
 xen-clean:
 	$(MAKE) -C $(XEN_PATH) clean
 
+# Make sure Xen and Xen tools have the same major.minor version or things are likely to break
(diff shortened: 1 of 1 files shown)


jenswi-linaro:

jforissier:
@ruchi393 is there a reason why we don't use Xen built from Buildroot instead? That would avoid a need for such a consistency check.

ruchi393:
> @ruchi393 is there a reason why we don't use Xen built from Buildroot instead? That would avoid a need for such a consistency check.

Sorry, I totally missed this query.  
OP-TEE support in xen is currently experimental which needs to be enabled in config options when building xen. That is not available with xen built with buuildroot. That is the reason why we need to keep the xen build separate from Buildroot.

ruchi393:
`Acked-by: Ruchika Gupta <ruchika.gupta@linaro.org>`

PR #519, opened 2021-08-20
linux-defconfig: explicit compile flag usage

ierkiaga:
Current linux config file generation does not add the cross compilation
flags, leading potentially to compilation errors. Not sure how a proper config
file is generated with current implementation but issues are observerd
specially when buildling the linux image with a single thread.

Signed-off-by: Ibai Erkiaga <ibai.erkiaga-elorza@xilinx.com>

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -376,6 +376,7 @@ linux-common: linux-defconfig
 $(LINUX_PATH)/.config: $(LINUX_DEFCONFIG_COMMON_FILES)
 	cd $(LINUX_PATH) && \
 		ARCH=$(LINUX_DEFCONFIG_COMMON_ARCH) \
+		CROSS_COMPILE=$(CROSS_COMPILE_NS_KERNEL) \
 		scripts/kconfig/merge_config.sh $(LINUX_DEFCONFIG_COMMON_FILES) \
 			$(LINUX_DEFCONFIG_BENCH)
 


jenswi-linaro:

common.mk
```diff
@@ -376,6 +376,7 @@ linux-common: linux-defconfig
 $(LINUX_PATH)/.config: $(LINUX_DEFCONFIG_COMMON_FILES)
 	cd $(LINUX_PATH) && \
 		ARCH=$(LINUX_DEFCONFIG_COMMON_ARCH) \
+		$(LINUX_COMMON_FLAGS) \
```
jforissier, line 379: Unfortunately this doesn't work because some platforms add targets to `$(LINUX_COMMON_FLAGS)` (not just flags). For example, `qemu.mk`: try `rm -f ../linux/.config ; make linux-defconfig` with this change and you will get:
```
/bin/bash: line 2: zImage: command not found
```

Perhaps `LINUX_COMMON_FLAGS` should be fixed not to include targets, but on the other hand I don't see why adding anything here is required. `ARCH` is set, and `CROSS_COMPILE` shouldn't matter for the config step (AFAICT).

ierkiaga: The issue I'm experiencing is that the config file that is generated does not have the CC set properly due to CROSS_COMPILE variable not being set when using merge_config script.

`CONFIG_CC_VERSION_TEXT="gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-39)"`

Then at build time the tool complains about the config file as CROSS_COMPILE is used for the build.
`make -C /home/ibaie/Workspace/Git/open-source/OP-TEE/workspace/build/../linux-xlnx LOCALVERSION= CROSS_COMPILE=" /home/ibaie/Workspace/Git/open-source/OP-TEE/workspace/build/../toolchains/aarch64/bin/aarch64-linux-gnu-" ARCH=arm64`
`make[1]: Entering directory /home/ibaie/Workspace/Git/open-source/OP-TEE/workspace/linux-xlnx'
arch/arm64/Makefile:44: Detected assembler with broken .inst; disassembly will be unreliable`
`  SYNC    include/config/auto.conf.cmd`
`*`
`* Restart config...`
`Enable support for pointer authentication (ARM64_PTR_AUTH) [Y/n/?] (NEW)`

What it has been surprising for me is that when I start the whole build process (repo sync....) and build the entire project with multiple threads (i.e. make -j4) then the build goes ahead and even the .config file is set properly.

I'm pretty new with the OP-TEE build repository so I might be missing something here. Anyway from the merge_config script it seems CROSS_COMPILE variable should be defined...

jforissier: > Anyway from the merge_config script it seems CROSS_COMPILE variable should be defined...

OK, then I think adding `CROSS_COMPILE=$(CROSS_COMPILE_NS_KERNEL)` should be enough here (if we don't want to rework/fix `LINUX_COMMON_FLAGS`).

ierkiaga: Agree, I think setting ARCH and CROSS_COMPILE should be enough. Let see if anyone else has comments and will update the patch

ierkiaga:
Comments have been addressed

jforissier:
@ierkiaga thanks, can you please squash the two patches and adjust the description a bit? Then that is:

ierkiaga:
@jforissier  just squashed the two patches and adjusted the description appropriately. I see that IBART failed the build for QEMU but it doesn't look to be related to my changes, PR#520 shows the same error.

jforissier:
@ierkiaga thanks for the update. Indeed the IBART error is unrelated, it will go away when rebasing this patch onto master. Those things happen when manifest.git and build.git are temporarily out of sync. In this case, IBART pulls the latest manifest.git, which also assumes the latest build.git but this PR was created before a required commit was merged.

PR #520, opened 2021-08-31
qemu: do not force kernel CONFIG_PREEMPT=y

jforissier:
Since Linux commit dcb3b06d9c34 ("tee: optee: replace might_sleep with
cond_resched") in v5.11, setting CONFIG_PREEMPT=y in kconfigs/qemu.conf
is not needed anymore. It was a workaround for the missing call to
cond_resched() in the TEE driver. At that time, might_sleep() together
with preemption enabled allowed to get rid of some RCU warning messages
when running long operations in secure world such as large key generation
(test case: "xtest -l 1 4007_rsa"). With cond_resched() the preemption
settings do not matter anymore.

As a result of this change, QEMUv8 is unmodified (CONFIG_PREEMPT=y is
set by default in the kernel's arch/arm64/configs/defconfig), but 32-bit
QEMU now has CONFIG_PREEMPT disabled. Both platforms are tested and
4007_rsa runs as expected with no warning.

### DIFF
--- a/kconfigs/qemu.conf
+++ b/kconfigs/qemu.conf
@@ -1,8 +1,5 @@
 CONFIG_TEE=y
 CONFIG_OPTEE=y
-### Enabling PREEMPT prevents random failures of
-### "make check" in Travis CI
-CONFIG_PREEMPT=y
 ### Enable 9P VFS
 CONFIG_NET_9P=y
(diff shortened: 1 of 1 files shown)


jenswi-linaro:

etienne-lms (commented):

PR #522, opened 2021-09-06
zynqmp: initial non petalinux based build

ierkiaga:
Current ZynqMP build script is based on Xilinx Petalinux tool, which
requires the installation of a proprietary tool. The following patch
removes this dependency and gets closed to the common build Makefile
used by other platforms.

The build is done using Xilinx out-of-tree repositories for Linux,
U-Boot and ATF, while using upstream OP-TEE and Buildroot repositories.

Finally the only non standard piece is the usage of pre-compiled
firmware images from the Xilinx wiki page.

Signed-off-by: Ibai Erkiaga <ibai.erkiaga-elorza@xilinx.com>


ierkiaga:
Just created a [pull request](https://github.com/OP-TEE/manifest/pull/201) with the initial manifest to support these changes

zynqmp.mk
```diff
@@ -1,88 +1,171 @@
...
+
+include toolchain.mk
+
+###############################################################################
+# Boot Image
+###############################################################################
+FIRMWARE_TARBALL = $(subst zynqmp-,2021.1-,$(PLATFORM))-release.tar.xz
+
+ifeq ("$(wildcard ../$(FIRMWARE_TARBALL))","")
+$(warning Release image tarball no present ../$(FIRMWARE_TARBALL))
```
jenswi-linaro, line 39: Is it still possible to complete the build without this?
I assume this is the binary which needs to be downloaded from the wiki page. Please add a link in the error message or at least here in the makefile.

jforissier: Or even add a target to download the file?

ierkiaga: Hi @jenswi-linaro and @jforissier, I added the link in the documentation repo update (did not pull request yet) but I think adding the URL here would also take sense. The main issue is that the wiki just provides a link to a Xilinx portal that requires account access, therefore no way to download as part of Makefile. I understand that might not be ideal but it is best than downloading few GB tools :)

The positive side is that the build can be performed even without this file, the only difference is that the bootable image will not be created but every single artifact will be built (OPTEE, Kernel, Rootfs.....)

jenswi-linaro: A problem with not erroring out here is that the log entry will be lost in a jungle of log entries.
Compare with the FVP setup where we error out if the model isn't available, even though everything could have been built.

ierkiaga: I see, actually my initial approach was that one. However I still see the value on building the sources specially for regression purposes. I mean, I just wanted to allow building OP-TEE without requiring to download binary blobs from Xilinx portal, mostly thinking on those not using Xilinx devices but doing some regression testing.

Maybe I can move the warning message print to the "all" target in a way that is printed just at the end, in a way that is more visible why the BOOT.bin file is missing from the deployment folder.

If you still think that is just better to stop the build if the firmware is missing, I can do that as well.

jforissier: Perhaps we should consider BOOT.bin is a dependency of the next step (`make flash` or whatever), and just ignore it completely until that step.

zynqmp.mk
```diff
@@ -1,88 +1,179 @@
...
+	make -C $(BOOTGEN_PATH) clean
+
+
+################################################################################
+# ZynqMPSoC Firmware mandatory for the boot
+################################################################################
+
+firmware:
+ifeq ("$(wildcard ../$(FIRMWARE_TARBALL))","")
+	$(error Release image tarball no present ../$(FIRMWARE_TARBALL))
```
jforissier, line 162: s/no/not/

jforissier (approved):
LGTM (but note the typo below).

PR #526, opened 2021-09-29
poplar fixes

igoropaniuk:
Poplar fixes


jbech-linaro:

jforissier:
> ```
> In the current version of linux kernel eMMC is enumerated as mmcblk0
> instead of mmcblk1 as before,
> ```
Can you please tell exactly which version of the kernel changed this? I think it is v5.14 (or more exactly some v5.14-rcX but that level of detail isn't required), since our `optee` branch was just rebased from `v5.13` to `v5.14`.

With that:

igoropaniuk:
@jforissier TBH I'm not sure where exactly it was changed, as I didn't do testing of Poplar for previous OP-TEE release, so 
it could be any linux version since OP-TEE 3.12.0 release date.

I've quickly checked `dw_mmc-hi3798cv200.c` driver and Poplar device tree/hi3798 SoC dt include, and I didn't see any changes related to this that could affect probe order.

That was definitely something more generic, but I didn't manage to find it yet

jforissier:
@igoropaniuk OK so let's leave the commit description as is. Thanks!

igoropaniuk:
tags applied and pushed
thanks!

PR #530, opened 2021-10-15
buildroot: Update toolchain when GDBSERVER=y

jforissier:
Fix the following build error:

 $ make GDBSERVER=y
 ...
 Makefile.legacy:9: *** "You have legacy configuration in your .config! Please check your configuration.".  Stop.

The error was introduced by the upgrade to Buildroot 2021.08 [1].

The deprecated option can be found by running "make menuconfig" in
out-br/ and checking the "Legacy options" submenu. It shows that
"gcc 8.x support removed" is selected. We can also see that
BR2_GCC_VERSION_8_X=y appears just below the comment "Legacy options
removed in 2021.08" in out-br/.config. This options comes from file
br-ext/configs/toolchain-br.

Upgrade the toolchain to 10.x which is available in Buildroot 2021.08
and corresponds to the pre-built version we use by default ("make
toolchains" currently downloads 10.2).

Link: https://github.com/OP-TEE/manifest/commit/e8d2210


jbech-linaro:

PR #538, opened 2021-12-17
FF-A: SPMC_AT_EL

jenswi-linaro:
Add support for FF-A on QEMU v8 using the new SPMC_AT_EL flag.

This needs https://github.com/OP-TEE/optee_os/pull/4886 rebased on optee_os master
2585cf9dfaad ("Linux 5.16-rc5") or later in the Linux kernel 
714ca37dc70a ("Merge "feat(plat/zynqmp): disable the -mbranch-protection flag" into integration") or later from TF-A

All these commits will appear in the next release of each project so make the needed preparations here too.


jforissier (commented):

Can this be taken in the coming OP-TEE release? Regarding the dependencies: our optee branch for the kernel is updated to v5.16 already, and I will soon merge https://github.com/OP-TEE/optee_os/pull/4886. But TF-A (v2.6) doesn't have the required commit you mentioned so it might be better to wait?

jenswi-linaro:
The only difference in default behaviour with this is that FF-A is enabled to be probed in the kernel config. Other than that we'll have unchanged behaviour unless `SPMC_AT_EL` is assigned `1` when building. It would be nice to get this out of the way, but we can also wait until after the release if you'd prefer that.

jforissier:
OK let's go ahead then.

jenswi-linaro:
Thanks. Rebased and tag applied.

PR #542, opened 2021-12-23
toolchain: Add optimizations for aarch64 host

b49020:
Lets switch over to using aarch64 hosted cross-compiler for aarch32
rather than building toochain for aarch32 from source which is very
combersome. But in case of native build for aarch64 on aarch64 host we
are left with no choice but to use buildroot toolchain as Arm doesn't
provide binary toolchain releases for aarch64 native compiler.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -108,11 +108,15 @@ QEMU_VIRTFS_MOUNTPOINT	?= /mnt/host
 ifeq ($(COMPILE_NS_USER),64)
 ifeq ($(UNAME_M),x86_64)
 MULTIARCH			:= aarch64-linux-gnu
+else ifeq ($(UNAME_M),aarch64)
+MULTIARCH			:= aarch64-linux
 else
 MULTIARCH			:= aarch64-linux
 endif
 else
 ifeq ($(UNAME_M),x86_64)
+MULTIARCH			:= arm-linux-gnueabihf
+else ifeq ($(UNAME_M),aarch64)
 MULTIARCH			:= arm-linux-gnueabihf
 else
 MULTIARCH			:= arm-linux
@@ -256,6 +260,12 @@ ifeq ($(COMPILE_LEGACY),)
 BUILDROOT_TOOLCHAIN=toolchain-aarch$(COMPILE_NS_USER)
 else
 BUILDROOT_TOOLCHAIN=toolchain-aarch$(COMPILE_NS_USER)-legacy
+endif
+else ifeq ($(UNAME_M),aarch64)
+ifeq ($(COMPILE_NS_USER),64)
+BUILDROOT_TOOLCHAIN=toolchain-aarch64-sdk
+else
+BUILDROOT_TOOLCHAIN=toolchain-aarch32
 endif
 else
 BUILDROOT_TOOLCHAIN=toolchain-aarch$(COMPILE_NS_USER)-sdk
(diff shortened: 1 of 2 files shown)


common.mk
```diff
@@ -257,6 +261,12 @@ BUILDROOT_TOOLCHAIN=toolchain-aarch$(COMPILE_NS_USER)
 else
 BUILDROOT_TOOLCHAIN=toolchain-aarch$(COMPILE_NS_USER)-legacy
 endif
+else ifeq ($(UNAME_M),aarch64)
```
jenswi-linaro, line 264: Is this some new syntax in GNU make or has this always worked?

jenswi-linaro (commented):

common.mk
b49020: As per [docs](https://www.gnu.org/software/make/manual/html_node/Conditional-Syntax.html#Conditional-Syntax), it looks like this has always worked.

b49020:
Tag applied.

common.mk
jenswi-linaro: Nice, I've totally missed that. :-)

b49020: No worries :).

toolchain.mk
```diff
@@ -61,7 +81,27 @@ endef
...
-else # $(UNAME_M) != x86_64
+else ifeq ($(UNAME_M),aarch64)
+
+AARCH32_PATH 			?= $(TOOLCHAIN_ROOT)/aarch32
+AARCH32_CROSS_COMPILE 		?= $(AARCH32_PATH)/bin/arm-linux-gnueabihf-
+AARCH32_GCC_VERSION 		?= gcc-arm-10.2-2020.11-aarch64-arm-none-linux-gnueabihf
+SRC_AARCH32_GCC 		?= https://developer.arm.com/-/media/Files/downloads/gnu-a/10.2-2020.11/binrel/$(AARCH32_GCC_VERSION).tar.xz
+
+AARCH64_PATH 			?= $(TOOLCHAIN_ROOT)/aarch64
+AARCH64_CROSS_COMPILE 		?= $(AARCH64_PATH)/bin/aarch64-linux-
```
jforissier, line 92: Why not use the native compiler directly? (by native I mean `/usr/bin/gcc` or whatever -- the compiler that comes with the Linux distribution).

b49020: That's the first thing I tried and it worked for standalone components but failed while building buildroot as it doesn't supports build using native toolchain (refer [here](https://buildroot.org/downloads/manual/manual.html#_cross_compilation_toolchain)). And then I tried to fake cross-compiler for buildroot as native compiler but that too failed while preparing sysroots and runtime libraries for rootfs.

jforissier: Alright thanks for these details. Could you mention that buildroot doesn't support the native toolchain (with the link) as a comment here perhaps?

b49020: Updated commit description accordingly.

b49020: Added comment in the code as well.

b49020:
@jforissier Any further comments?

jforissier (approved):
s/combersome/cumbersome/ in the commit description, with that fixed:

Thanks!

b49020:
Fixed typo and tag applied.

PR #543, opened 2021-12-24
qemu: Enable Linux Trusted Keys test setup

b49020:
Brief description of changes:
- Enable Qemu kernel configs for Trusted Keys.
- Add keyutils buildroot package.
- Enable Trusted Keys early TA in OP-TEE.
- Create a new test script (trusted-keys.exp) which is invoked as part
  of "make check" for automated Trusted Keys tests.


jenswi-linaro:

b49020:
Tag applied.

qemu-check.exp
```diff
@@ -158,3 +158,6 @@ if {$::env(XEN_BOOT) == "y"} {
 	send -- "$cmd\r"
 	check_test_result $teecore
 }
+# Invoke Trusted Keys tests
+set basedir [file dirname $argv0]
+source $basedir/trusted-keys.exp
```
jforissier, line 163: Wouldn't it be more convenient to have a separate target in the makefile so that one can easily run the Trusted Keys tests on their own? (similar to `check-rust`).

b49020: Actually the major motivation was to run Trusted Keys tests as part of CI as these are pretty quick tests which wouldn't have any noticeable impact on CI runtime. However, I liked your suggestion that one should be able to run the Trusted Keys tests standalone without invoking the `xtest`. How about we just add `check-only-trusted-keys` target that would omit the `xtest` while invoking the same script?

b49020: Also, `check-only-xtest` can be added on similar lines?

jforissier: Sounds good to me.

b49020:
Added a new patch that add granular control over which `make check` tests to execute.

qemu-check.exp
b49020: I just used additional command line option as `CHECK_TESTS` instead as that turned out to be much simpler.

b49020:
@jforissier do you have any further comments?

jforissier:
@b49020 LGTM, thanks for the update.

b49020:
Tag applied.

PR #552, opened 2022-02-02
[RFC] Optee os flags

jenswi-linaro:

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -517,6 +517,15 @@ OPTEE_OS_FLAGS = \
 	CFG_TEE_BENCHMARK=$(CFG_TEE_BENCHMARK) \
 	CFG_IN_TREE_EARLY_TAS=trusted_keys/f04a0fe7-1f5d-4b9b-abf7-619b85b4ce8c
 
+# For convenience here's a few variables often tweaked with during development
+#OPTEE_CFG_WERROR=y
+#OPTEE_CFG_CC_OPT_LEVEL=0
+#OPTEE_CFG_CORE_ASLR=n
+#OPTEE_CFG_CORE_UNMAP_CORE_AT_EL0=n
+#OPTEE_CFG_WITH_PAGER=y
+#OPTEE_CFG_TEE_CORE_MALLOC_DEBUG=y
+#OPTEE_CFG_TEE_CORE_DEBUG=y
+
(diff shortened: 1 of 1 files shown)


common.mk
```diff
@@ -517,6 +517,15 @@ OPTEE_OS_FLAGS = \
...
+#OPTEE_CFG_CC_OPT_LEVEL=0
+#OPTEE_CFG_CORE_ASLR=n
+#OPTEE_CFG_CORE_UNMAP_CORE_AT_EL0=n
+#OPTEE_CFG_WITH_PAGER=y
+#OPTEE_CFG_TEE_CORE_MALLOC_DEBUG=y
+#OPTEE_CFG_TEE_CORE_DEBUG=y
+
 # This is done last to allow overriding any default CFG variable
 OPTEE_OS_FLAGS += \
 	$(foreach v, $(filter OPTEE_CFG_%,$(.VARIABLES)),$(v)=$($(v)))
```
jforissier, line 531: It is an elegant solution, but is it really useful? I mean, `make CFG_CORE_ASLR=n` currently has the same result as `make OPTEE_CFG_CORE_ASLR=n` with this patch. The only difference is that the latter would set the flag only for `optee_os`, not for all the sub-projects. It is a good thing (might catch issues with some flags not being properly exported via the TA SDK for instance...) but it is worth it? I have no strong opinion.

jenswi-linaro: The lack of precision is disturbing even if it almost always will do what's intended.

By the way, why do we the pass
```
	CFG_TEE_CORE_LOG_LEVEL=$(CFG_TEE_CORE_LOG_LEVEL) \
	DEBUG=$(DEBUG) \
￼	CFG_TEE_BENCHMARK=$(CFG_TEE_BENCHMARK) \
```
above?

jforissier: > The lack of precision is disturbing even if it almost always will do what's intended.

True.

> By the way, why do we the pass
> 
> ```
> 	CFG_TEE_CORE_LOG_LEVEL=$(CFG_TEE_CORE_LOG_LEVEL) \
> 	DEBUG=$(DEBUG) \
> 	CFG_TEE_BENCHMARK=$(CFG_TEE_BENCHMARK) \
> ```
> 
> above?

Totally not needed.

PR #560, opened 2022-04-20
qemu_v8: Introducing MEMTAG option

jenswi-linaro:
Introduces the option MEMTAG which enabled tagging of memory using the
ARM Memory Tagging Extension (MTE).

### DIFF
--- a/qemu_v8.mk
+++ b/qemu_v8.mk
@@ -53,6 +53,9 @@ endif
 # Option to configure Pointer Authentication for TA's
 PAUTH ?= n
 
+# Option to configure Memory Tagging Extension
+MEMTAG ?= n
+
 ################################################################################
 # Paths to git projects and various binaries
 ################################################################################
@@ -185,6 +188,9 @@ endif
 
 ifeq ($(PAUTH),y)
 TF_A_FLAGS += CTX_INCLUDE_PAUTH_REGS=1
+endif
+ifeq ($(MEMTAG),y)
+TF_A_FLAGS += CTX_INCLUDE_MTE_REGS=1
 endif
 
 arm-tf: optee-os $(BL33_DEPS)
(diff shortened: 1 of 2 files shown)


jbech-linaro (commented):

qemu_v8.mk
```diff
@@ -441,7 +456,7 @@ run-only:
 		-nographic \
 		-serial tcp:localhost:54320 -serial tcp:localhost:54321 \
 		-smp $(QEMU_SMP) \
-		-s -S -machine virt,secure=on,gic-version=$(QEMU_GIC_VERSION),virtualization=$(QEMU_VIRT) \
+		-s -S -machine virt,secure=on,mte=$(QEMU_MTE),gic-version=$(QEMU_GIC_VERSION),virtualization=$(QEMU_VIRT) \
```
jforissier, line 459: There is a similar line in `qemu-check.exp`.

jenswi-linaro:
I've updated `qemu-check.exp` with MTE. When I try `make check MEMTAG=y -j8` it fails with:
```
Starting QEMU... done, guest is booted.
Running: xtest...
##################################################
##################################################
#############################################!!! assertion 'memtag_get_tag(addr) == memtag_read_tag(addr)' failed at lib/libutils/ext/include/memtag.h:214 in memtag_assert_tag()
make: *** [Makefile:483: check] Error 1
```

The assert is part of one of the test cases so I'm a bit surprised we error out here.

jforissier:
> The assert is part of one of the test cases so I'm a bit surprised we error out here.

IIRC there is some output that the Expect script always considers as errors, "assertion" is probably one of them. We may want to relax that a bit.

jforissier:
For commit "qemu-check.exp: trap only OP-TEE Core assert or panic":

jforissier:
For commit "qemu_v8: Introducing MEMTAG option":

jenswi-linaro:
Squashed and tags applied

jenswi-linaro:
Good to merge?

jforissier:
> Good to merge?

Yes, thanks!

PR #562, opened 2022-05-03
qemu_v8: xen: apply patch to fix issue with RPC buffer release

jforissier:
Adds a patch to be applied by Buildroot when building Xen. This fixes a
race condition in RPC buffer management, which causes random failures in
the OP-TEE CI jobs [1].

Link: [1] https://dev.azure.com/OPTEE/optee_os/_build/results?buildId=2827&view=logs&j=705748f3-7146-5e86-79af-1c0266d20a8c&t=d3458a7d-1a81-5639-8e63-8927d807e53a&l=7812
Fixes: https://github.com/OP-TEE/optee_os/issues/5288

### DIFF
--- a/br-ext/patches/xen/0001-optee-immediately-free-RPC-buffers.patch
+++ b/br-ext/patches/xen/0001-optee-immediately-free-RPC-buffers.patch
@@ -1,0 +1,70 @@
+From e6059b14107cc061cc35fad44afdd9ad25ba1181 Mon Sep 17 00:00:00 2001
+From: Jens Wiklander <jens.wiklander@linaro.org>
+Date: Thu, 28 Apr 2022 09:07:41 +0200
+Subject: [PATCH] optee: immediately free RPC buffers that are released by
+ OP-TEE
+
+This commit fixes a case overlooked in [1].
+
+There are two kinds of shared memory buffers used by OP-TEE:
+1. Normal payload buffer
+2. Internal command structure buffers
+
+The internal command structure buffers are represented with a shadow
+copy internally in Xen since this buffer can contain physical addresses
+that may need to be translated between real physical address and guest
+physical address without leaking information to the guest.
+
+[1] fixes the problem when releasing the normal payload buffers. The
+internal command structure buffers must be released in the same way.
+Failure to follow this order opens a window where the guest has freed
+the shared memory but Xen is still tracking the buffer.
+
+During this window the guest may happen to recycle this particular
+shared memory in some other thread and try to use it. Xen will block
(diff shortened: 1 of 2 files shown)


jbech-linaro:
LGTM, but would it be worth mention why we have this as a `*.patch` and not merging it directly into xen? I suppose at some point we want to merge this officially, or?

jenswi-linaro:
FYI I've just sent the patch to xen-devel https://lists.xenproject.org/archives/html/xen-devel/2022-05/msg00245.html

jforissier:
@jenswi-linaro thanks, I have added the link in the comment and in the commit description.

jforissier:
The failure that the Xen patch is supposed to fix happened again here: https://dev.azure.com/OPTEE/optee_os/_build/results?buildId=2927&view=logs&j=705748f3-7146-5e86-79af-1c0266d20a8c&t=d3458a7d-1a81-5639-8e63-8927d807e53a&l=10934
The image used during this run seems to have the patch properly applied though.
```
$ docker run -it jforissier/optee_os_ci:qemuv8_check@sha256:fdf20d5d291ee38f9d1f94bae9f6df8683c54121532bae3bb7afd171343bf63d
root@d21a68bdcd78:/# head -1160 /root/optee_repo_qemu_v8/out-br/build/xen-4.14.3/xen/arch/arm/tee/optee.c | tail
        unmap_domain_page(shm_rpc->xen_arg);
    } else if (call->rpc_op == OPTEE_SMC_RPC_FUNC_FREE) {
        uint64_t cookie = regpair_to_uint64(get_user_reg(regs, 1),
                                            get_user_reg(regs, 2));

        free_shm_rpc(ctx, cookie);
    }


    return ret;
```

jenswi-linaro:
Are you sure that the patch is used when Xen is compiled? I've started a local build, but it looks like the patch isn't used.

jenswi-linaro:
That seems to cover xen-tools only. The hypervisor itself is not build via buildroot.

jforissier:
> That seems to cover xen-tools only. The hypervisor itself is not build via buildroot.

Duh :facepalm: you're absolutely right and I should have remembered https://github.com/OP-TEE/build/pull/517 for example.
So this means two things:
1. This (closed) PR is actually useless and can be reverted
2. We need to find another way to build the patched Xen in our environment. I'd like to avoid forking Xen for such a temporary thing. It seems `repo` allows some customization and we could apply the patch after `repo sync` maybe (I'm thinking about the `repo-hooks` element). I will investigate.

jforissier:
https://github.com/OP-TEE/build/pull/565

PR #563, opened 2022-05-05
hikey: generate only one of linux-4g and linux-8g

jforissier:
CFG_FLASH_SIZE can be used to generate only linux-4g or linux-8g instead
of both.


jbech-linaro:

PR #564, opened 2022-05-05
tf-a/qemu v8: Change LOG_LEVEL to 40 when DEBUG=1

jbech-linaro:
With LOG_LEVEL=50 and DEBUG=1, BL31 becomes too big, with LOG_LEVEL=40
we will miss a few debug prints, but it will at produce a valid binary
and more importantly, you'll get an elf-file with all debug symbols,
i.e., it's GDB makes up for the missing prints.

(FWIW, there is a compiler error with `50` as well, I'll send a fix for that to TF-A)


jbech-linaro:

PR #565, opened 2022-05-06
qemu_v8: build Xen with Buildroot

jforissier:
Commit 8b808915f856 ("qemu_v8: xen: apply patch to fix issue with RPC
buffer release") really did only one part of the job. Only the Xen tools
are built with Buildroot; the hypervisor itself (which needs the patch
mentioned in the commit) is actually built from the top-level xen/
directory. As a result the issue is stil present.
This commit changes the QEMUv8 makefiles as well as the Buildroot
configuration for Xen so that both the hypervisor and the tools are
built. As for the Xen configuration (enabling OP-TEE support), it is
introduced via a patch to the defconfig file. Patch file
"0001-Revert-xen-arm-Restrict-the-amount-of-memory-that-do.patch"
addresses a regression introduced in Xen 4.14.3 (which Buildroot
provides) compared to 4.14.1 (which is currently provided by
$(ROOT)/xen).


jforissier:
@lorc what do you think of this "regression" I mentioned in the commit description? Is this something that should be addressed in OP-TEE or in Xen?

jenswi-linaro:
Is there a risk of a chicken and egg problem with this? I mean buildroot depends on OP-TEE and it's not far fetched that some boot thing might depend on the Xen binary produced by buildroot.

jforissier:
> Is there a risk of a chicken and egg problem with this? I mean buildroot depends on OP-TEE and it's not far fetched that some boot thing might depend on the Xen binary produced by buildroot.

Do you have anything specific in mind? The following builds and runs fine:
```
mkdir optee && cd optee
repo init -u https://github.com/OP-TEE/manifest.git -m qemu_v8.xml
repo sync -j20
rm -rf xen
cd build
git fetch github pull/565/head && git checkout FETCH_HEAD
make -j2 toolchains
make -j10 XEN_BOOT=y check
```

qemu_v8.mk
```diff
@@ -374,46 +374,31 @@ $(ROOTFS_UGZ): u-boot buildroot | $(BINARIES_PATH)
...
-	$(MAKE) -C $(XEN_PATH)/xen XEN_TARGET_ARCH=arm64 defconfig
-	cd $(XEN_PATH)/xen && \
-	tools/kconfig/merge_config.sh -m .config $(ROOT)/build/kconfigs/xen.conf
-
-xen: $(XEN_PATH)/xen/.config
-	$(MAKE) -C $(XEN_PATH) dist-xen \
-	XEN_TARGET_ARCH=arm64 \
-	CONFIG_XEN_INSTALL_SUFFIX=.gz	\
-	CROSS_COMPILE="$(CCACHE)$(AARCH64_CROSS_COMPILE)"
+#.PHONY: xen
```
jenswi-linaro, line 377: I guess this can be removed.

jenswi-linaro:
> 

No, not really. I guess it's fine then.

br-ext/patches/xen/0001-Revert-xen-arm-Restrict-the-amount-of-memory-that-do.patch
```diff
@@ -0,0 +1,37 @@
+From 82a8e6a916f1a7bb3910c365197319b37085d58c Mon Sep 17 00:00:00 2001
```
jenswi-linaro, line 1: Would it make sense to take this patch in a separate commit?

jenswi-linaro: Can't it be added before the "qemu_v8: build Xen with Buildroot" commit?

jforissier: Yes, I will split the commit.

jforissier:
Update

jenswi-linaro:

PR #569, opened 2022-06-02
optee-rust: prevent new execution after successful build

jforissier:
"make optee-rust OPTEE_RUST_ENABLE=y" always runs the setup.sh script
in $(ROOT)/optee_rust. The script clones about 600 MB of data. That
can't be done each time especially since the check-rust target depends
on optee-rust.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -534,10 +534,15 @@ optee-os-clean-common:
 # OP-TEE Rust
 ################################################################################
 .PHONY: optee-rust
-optee-rust:
+optee-rust: $(OPTEE_RUST_PATH)/.done
+
+$(OPTEE_RUST_PATH)/.done:
 ifeq ($(OPTEE_RUST_ENABLE),y)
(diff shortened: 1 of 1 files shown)


jforissier:
@DemesneGH is this PR acceptable? I'm almost certain it isn't, because I believe that running `setup.sh` is needed each time `make check-rust` is run in order to set the environment, right?
If we can't do what I propose here, we urgently need a fix in `optee_rust/setup.sh`. Why on earth do you systematically `rm -rf rust/` and clone tons of code?

DemesneGH:
Hi @jforissier 

> is this PR acceptable?

Yes. The `setup.sh` is for installing Rust and cloning submodules, which is needed only before building Rust examples at the first time. So it is necessary to avoid cloning submodules when building multiple times. I'll test the PR in the next few days, thanks!

DemesneGH:
@jforissier LGTM.
`Tested-by: Yuan Zhuang <zhuangyuan04@baidu.com>`

PR #574, opened 2022-06-28
qemu-check.exp: match core assert/panic strings on a single line

jforissier:
The regular expression used to detect TEE core panics or assertion
failures must not allow multi-line text to match, otherwise the following
input may trigger it:

E/TC:? 0 TA panicked with code 0xbeef
E/LD:  Status of TA 5b9e0e40-2636-11e1-ad9e-0002a5d5c51b
E/LD:   arch: arm
E/LD:  region  0: va 0x00102000 pa 0x0e300000 size 0x002000 flags rw-s (ldelf)
E/LD:  region  1: va 0x00104000 pa 0x0e302000 size 0x00e000 flags r-xs (ldelf)
E/LD:  region  2: va 0x00112000 pa 0x0e310000 size 0x001000 flags rw-s (ldelf)
E/LD:  region  3: va 0x00113000 pa 0x0e311000 size 0x004000 flags rw-s (ldelf)
E/LD:  region  4: va 0x00117000 pa 0x0e315000 size 0x001000 flags r--s
E/LD:  region  5: va 0x00118000 pa 0x0e410000 size 0x003000 flags rw-s (stack)
E/LD:  region  6: va 0x00184000 pa 0x00001000 size 0x015000 flags r-xs [0]
E/LD:  region  7: va 0x00199000 pa 0x00016000 size 0x0e5000 flags rw-s [0]
E/LD:  assertion 'maps[map_idx].sz == sz' failed at ldelf/ta_elf.c:1451 in ta_elf_print_mappings()

Test case: xtest 1007 with "make COMPILER=clang".

Fixes: 09900ef99b70 ("qemu-check.exp: trap only OP-TEE Core assert or panic")

### DIFF
--- a/qemu-check.exp
+++ b/qemu-check.exp
@@ -75,11 +75,11 @@ proc check_test_result arg {
 		# Exit when result separator is seen
 		"+-----------------------------------------------------\r\r" {}
 		# Handle errors in TEE core output
-		-i $arg -re {(^..TC:.*assertion.*failed at.*)\n} {
+		-i $arg -re {(..TC:[^\n]*assertion[^\n]*failed at[^\n]*)} {
 			info "!!! $expect_out(1,string)\n"
 			exit 1
 		}
-		-i $arg -re {(^..TC:.*Panic at.*)\n} {
+		-i $arg -re {(..TC:[^\n]*Panic at[^\n]*)} {
 			info "!!! $expect_out(1,string)\n"
 			exit 1
 		}


qemu-check.exp
```diff
@@ -75,11 +75,11 @@ proc check_test_result arg {
 		# Exit when result separator is seen
 		"+-----------------------------------------------------\r\r" {}
 		# Handle errors in TEE core output
-		-i $arg -re {(^..TC:.*assertion.*failed at.*)\n} {
+		-i $arg -re {(^..TC:[^\n]*assertion.*failed at.*)\n} {
```
jenswi-linaro, line 78: Why do we have a `\n` at the end, wouldn't it make more sense with `$`?

jforissier: Very good question which made me realize that in Expect, `^` and `$` are not the beginning and end of a line, but of the matching buffer. Therefore, the `\n` at the end of the RE is indeed correct, but the `^` at the start is not.
So the proper RE is `\n(..TC:[^\n]*assertion[^\n]*failed at[^\n]*)\n`.
I'll fix.

jforissier:
Update

jenswi-linaro (commented):

jforissier:
Oops, sorry I forgot to add the second change I mentioned in the fixup description (do not match `\n` at the beginning/end). @jenswi-linaro let me know if you agree with the final version.

jenswi-linaro:
LGTM

etienne-lms (approved):
`Acked-by: Etienne Carriere <etienne.carriere@linaro.org>` for what it's worth.

jforissier:
@etienne-lms interesting CI [failure](https://optee.mooo.com:5000/logs/OP-TEE/build/574/981050706/fba618922a0cd7baed0441dfbc68969a206eb500) (check line 16818), shouldn't the following be added to build.git?
```diff
diff --git a/br-ext/package/optee_client_ext/optee_client_ext.mk b/br-ext/package/optee_client_ext/optee_client_ext.mk
index 6e4b785..0fbaf74 100644
--- a/br-ext/package/optee_client_ext/optee_client_ext.mk
+++ b/br-ext/package/optee_client_ext/optee_client_ext.mk
@@ -3,6 +3,7 @@ OPTEE_CLIENT_EXT_SOURCE = local
 OPTEE_CLIENT_EXT_SITE = $(BR2_PACKAGE_OPTEE_CLIENT_EXT_SITE)
 OPTEE_CLIENT_EXT_SITE_METHOD = local
 OPTEE_CLIENT_EXT_INSTALL_STAGING = YES
+OPTEE_CLIENT_EXT_DEPENDENCIES = opensc

 ifeq ($(BR2_PACKAGE_OPTEE_BENCHMARK_EXT),y)
 OPTEE_CLIENT_EXT_CONF_OPTS = -DCFG_TEE_BENCHMARK=ON
```

etienne-lms:
Right. [`BR2_PACKAGE_OPENSC ?= y`](https://github.com/OP-TEE/build/blob/3.17.0/common.mk#L317-L318) default embeds pkcs11 test and possibly tools.

PR #575, opened 2022-06-29
stm32mp1: rely on stm32mp15_defconfig

etienne-lms:
Configures U-Boot with stm32mp15_defconfig that was introduced in
U-Boot tag v2021.10 for devices booted with OP-TEE and U-Boot stored
in TF-A FIP container. From U-Boot v2021.10 onward,
stm32mp15_trusted_defconfig is dedicated to boot scheme that do not
use TF-A FIP format.


jenswi-linaro:

etienne-lms:
tag applied

PR #577, opened 2022-07-04
qemu_v8: xen: increase guest RAM from 128 to 160 MB

jforissier:
Increase the size of the guest (DomU) RAM from 128 to 160 MB to fix
an issue with xtest 6018:

  * regression_6018 Large object
  o regression_6018.1 Storage id: 00000001
  writing 0
  writing 1
  writing 2
  writing 3
  writing 4
  writing 5
  regression_6000.c:1895: fs_write(&sess, obj, block, block_size) has an unexpected value: 0xffff3024 = TEE_ERROR_TARGET_DEAD, expected 0x0 = TEEC_SUCCESS
    regression_6018.1 FAILED

For some reason, this test passes in CI [1] but not on my laptop
("make XEN_BOOT=y run", login: "root", run /bin/domu, login: "test"
then run "xtest 6018"). tee-supplicant receives error ENOSPC when
trying to write to /data/tee. Since the root FS in the guest is a
ramdisk, increasing the RAM seems to be the appropriate solution.

Link: [1] https://github.com/OP-TEE/optee_os/blob/578f89d28d2b66c35128f095989c355fbc2ef3ab/.azure-pipelines.yml#L256

### DIFF
--- a/qemu_v8/xen/guest.cfg
+++ b/qemu_v8/xen/guest.cfg
@@ -1,5 +1,5 @@
 kernel="/mnt/host/linux/arch/arm64/boot/Image"
-memory=128
+memory=160
 vcpus=1
 command="console=hvc0 earlycon=xenboot"
 ramdisk = "/mnt/host/out-br-domu/images/rootfs.cpio.gz"


jenswi-linaro:

PR #578, opened 2022-08-03
toolchain.mk: allow parallel build of cross-compiler on Aarch64 hosts

jforissier:
On an Aarch64 host, "make toolchain" does not use parallel jobs when
building the Aarch64 GCC and the following warning is shown:

 $ make -j32 toolchains
 [...]
 make[1]: warning: jobserver unavailable: using -j1.  Add `+' to parent make rule.

Therefore, add the '+' sign as recommended. This speeds up the build
quite a bit on a 32-core machine for example.


jbech-linaro:

PR #583, opened 2022-08-25
buildroot: select util-linux, libuuid and pkgconf

eaaltonen:
This change allows to build the `libckteeaclc` support library from https://github.com/OP-TEE/optee_client/pull/259 (_not merged_).

### Commit

In order to use login methods `TEEC_LOGIN_USER` and `TEEC_LOGIN_GROUP`,
a user space application must initialize the tokens with the SHA1
hashed UUID of the user or group.

Select host-pkgconf, util-linux and libuuid in order to use `libuuid`
and add supporting libraries to optee_client for this type of ACL based
login initialization.

Signed-off-by: Eero Aaltonen <eero.aaltonen@vaisala.com>

[x] check workflow, rebase on master.

Also for the other repos, does all work go on top of `master`, or does the project have some supported maintenance releases?
If there are supported maintenance releases, are those done by:
* cherry-picking backwards from `master`
* merge forward from mainteance release

- [x] squash fixup commit before merging.


jenswi-linaro:
> Also for the other repos, does all work go on top of `master`, or does the project have some supported maintenance releases?

We're currently only working on top of `master`.

eaaltonen:
> We're currently only working on top of `master`.

Roger that. Rebased on top of current `master` and re-tested. No changes.

jforissier:
Why/when use `select` rather than add packages to `_DEPENDENCIES`? I don't understand the logic here.

eaaltonen:
> Why/when use `select` rather than add packages to `_DEPENDENCIES`? I don't understand the logic here.

Buildroot [user manual section "18.2.3. Choosing depends on or select"](https://www.buildroot.org/downloads/manual/manual.html#_config_files) instructs that `select` should be used to make sure any dependent libraries are built.

The section also states in the bottom:
> Note that such dependencies will ensure that the dependency option is also enabled, but not necessarily built before your package. To do so, the dependency also needs to be expressed in the .mk file of the package.

So based on:
- the documentation
- my own testing
- other examples in the main buildroot tree

`select` (_in practice_) is required to ensure the dependency is built, `_DEPENDENCIES` is required for ordering, and both are required for the package to be available in the per-package sysroot.

br-ext/package/optee_client_ext/optee_client_ext.mk
```diff
@@ -2,6 +2,8 @@ OPTEE_CLIENT_EXT_VERSION = 1.0
 OPTEE_CLIENT_EXT_SOURCE = local
 OPTEE_CLIENT_EXT_SITE = $(BR2_PACKAGE_OPTEE_CLIENT_EXT_SITE)
 OPTEE_CLIENT_EXT_SITE_METHOD = local
+OPTEE_CLIENT_EXT_DEPENDENCIES = host-pkgconf \
+	$(if $(BR2_PACKAGE_UTIL_LINUX_LIBS),util-linux-libs,util-linux)
```
jforissier, line 6: Could this be simplified to just `util-linux` perhaps? This pattern is used in two places in upstream Buildroot, and each time the commits mention circular dependencies which I think might not apply here: https://github.com/buildroot/buildroot/commit/eb058222596 and https://github.com/buildroot/buildroot/commit/e3c86f5c9e4.

jforissier (commented):
Thanks for the info about `select` vs. `_DEPENDENCIES`. Still one more question.

jforissier (commented):
@eaaltonen thanks for the updates, makes (some) sense to me now.

eaaltonen:
@jforissier : thanks. squashed and added `Acked-by`.

PR #585, opened 2022-09-02
fvp: add support for Secure Partitions in the FIP

balint-dobszay-arm:
This commit introduces a new packaging option for Secure Partitions. Instead of embedding the images into the OP-TEE binary, TF-A offers a mechanism to encapsulate an SP image and its manifest into an SP package and add that to the FIP.
https://trustedfirmware-a.readthedocs.io/en/v2.6/components/secure-partition-manager.html#secure-partition-packages

TF-A needs two config options to enable this:

- SP_LAYOUT_FILE: This json file contains the path of the SP images and their corresponding manifests. It's generated by Trusted Services.
- ARM_BL2_SP_LIST_DTS: This dts snippet describes where each SP package should be loaded by BL2.

Depends on:

- https://github.com/OP-TEE/optee_os/pull/5519
- https://review.trustedfirmware.org/q/topic:fip-sp

### DIFF
--- a/fvp-psa-sp.mk
+++ b/fvp-psa-sp.mk
@@ -5,6 +5,7 @@ MEASURED_BOOT			?= y
 MEASURED_BOOT_FTPM		?= n
 TS_SMM_GATEWAY			?= y
 TS_UEFI_TESTS			?= y
+SP_PACKAGING_METHOD		?= embedded # Supported values: embedded, fip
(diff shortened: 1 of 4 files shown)


jenswi-linaro:

balint-dobszay-arm:
Tag applied, thanks.

PR #587, opened 2022-09-12
rust: fix make optee-rust

jforissier:
In some environments it has been noted that the Rust setup fails with a netowrk error [1]. It seems to be an issue with the built-in Git library, a workaround being to tell the Rust setup to use the Git command instead. This commit does exactly that, regardless of the environment.

Note: I chose not to restore the OP-TEE Rust build commands removed by [1] in order to reduce the size of the image ($TOP/optee_rust is 1.8GB after building) and avoid disk space issues in the "make check (QEMUv8)" job which uses this image but doesn't build the Rust tests. The "make check-rust (QEMUv8)" job will do whataver it takes.

Link: [1] https://github.com/jforissier/docker_optee_os_ci/commit/6d7f39aec9436168e5e4ce44c4b6f7791575e313

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -542,7 +542,9 @@ optee-rust: $(OPTEE_RUST_PATH)/.done
 
 $(OPTEE_RUST_PATH)/.done:
 ifeq ($(OPTEE_RUST_ENABLE),y)
-	@(export OPTEE_DIR=$(ROOT) && cd $(OPTEE_RUST_PATH) && ./setup.sh && touch .done)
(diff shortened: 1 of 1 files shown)


jforissier:
Note: tested manually in the CI Docker image `jforissier/optee_os_ci:qemuv8_check`.

DemesneGH:
`Acked-by: Yuan Zhuang <zhuangyuan04@baidu.com>`

jforissier:
@DemesneGH thanks!

PR #598, opened 2022-10-17
soc_term.py: write to stdout not stdin

jforissier:
The serv_conn() function should write the data read from the connection to stdout, not stdin. Although stdin appears to work when soc_term is invoked from a terminal in an interactive session, things get bad when trying to redirect its output:

 $ ./soc_term 54320 | tee nsec.log
 <... make run etc. ...>
 ^C
 $ file nsec.log
 nsec.log: empty


etienne-lms (commented):

jenswi-linaro (commented):

Issue #597, opened 2022-10-14
Build error with huge $PATH


wamserma:
It seems that the changes in https://github.com/OP-TEE/build/commit/e1ece8328f9f6102bda339aef4e8bf46975a67bd accidentally dropped the fixes from https://github.com/OP-TEE/build/pull/393, so building fails again on systems that have many entries in their PATH variable.

@etienne-lms Maybe you can have another look and try adopt #393 again.

etienne-lms:
You are right. I'll create a P-R and discuss accuracy through reviews.

Issue #389, opened 2019-10-07
[Buildroot/Tests] Build error in optee_test.mk for certain build setups


wamserma:
In certain build setups the contents of variable `TARGET_CONFIGURE_OPTS` can become a very long string (e.g. due to a long PATH in a nix-shell).
With the enumeration of tests done by the `foreach` function in https://github.com/OP-TEE/build/blob/03b920bded265525663b813037ff6e95272249f7/br-ext/package/optee_test/optee_test.mk#L11 this multiplies into a very very long string held in `OPTEE_TEST_BUILD_TAS`, up to the point where the string exceeds Linux' limit for `execv()` calls, which is 128KiB (even if ARG_MAX is 2MiB).
A straightforward solution could be to push the enumeration further down the call stack (from make to the invoked shell), that is replacing the `$(foreach ...)` with a shell loop.
The issue is not present in the corresponding INSTALL statement, and did not show up yet for with the examples as their number is lower than the number of tests.

Shall I make a PR for that or will this soon be obsolete via https://github.com/OP-TEE/build/pull/353?

jbech-linaro:
@wamserma , feel free to make PR for it right away, since I believe #353 will run with slow progress (and I wonder if we ever get anywhere with it).

PR #599, opened 2022-10-17
Fix build error with huge $PATH

etienne-lms:
Fix for Issue #597.

### DIFF
--- a/br-ext/package/optee_examples_ext/optee_examples_ext.mk
+++ b/br-ext/package/optee_examples_ext/optee_examples_ext.mk
@@ -8,20 +8,23 @@ OPTEE_EXAMPLES_EXT_SDK = $(BR2_PACKAGE_OPTEE_EXAMPLES_EXT_SDK)
 OPTEE_EXAMPLES_EXT_CONF_OPTS = -DOPTEE_EXAMPLES_SDK=$(OPTEE_EXAMPLES_EXT_SDK)
 
 define OPTEE_EXAMPLES_EXT_BUILD_TAS
-	@$(foreach f,$(wildcard $(@D)/*/ta/Makefile), \
-		echo Building $f && \
-			$(MAKE) CROSS_COMPILE="$(shell echo $(BR2_PACKAGE_OPTEE_EXAMPLES_EXT_CROSS_COMPILE))" \
-			O=out TA_DEV_KIT_DIR=$(OPTEE_EXAMPLES_EXT_SDK) \
-			PYTHON3=$(HOST_DIR)/bin/python3 \
-			$(TARGET_CONFIGURE_OPTS) -C $(dir $f) all &&) true
+	@for f in $(@D)/*/ta/Makefile; \
+	do \
+	  echo Building $$f && \
+	  $(MAKE) CROSS_COMPILE="$(shell echo $(BR2_PACKAGE_OPTEE_EXAMPLES_EXT_CROSS_COMPILE))" \
+	    O=out TA_DEV_KIT_DIR=$(OPTEE_EXAMPLES_EXT_SDK) \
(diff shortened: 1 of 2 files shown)


jenswi-linaro:
> making the dependencies list being likely built twice

From my point of view, this looks like a degradation. I know this is an old fix, but it still looks strange to change random stuff until something reportedly works again. How is this problem reproduced upstream?

jforissier:
I think a better option would be to generate one make target per TA, with the additional benefits that parallel build would be possible, and the first failure would stop the build.

etienne-lms:
> > making the dependencies list being likely built twice
>
> From my point of view, this looks like a degradation. 

True, buiding twice some pieces is not nice, but build env find out nothing more needs to be done for those components and it does not affect overall build time. Yet, this part and the way I use a local shell variable to create the list to deps paths is not very clean.

I have not reproduced the inital issue.
@wamserma, do you confirm https://github.com/OP-TEE/build/issues/389 ris still affecting build environments?

etienne-lms:
We have a nice _ta/Makefile_ here that knowns which TAs and order for build. We could use it, but it's scratched by CMake generated _Makefile_ file. Maybe rename the original one?

jforissier:
> > I think a better option would be to generate one make target per TA ...
> 
> We have a nice _ta/Makefile_ **~here~ in _optee_test.git_** that knowns which TAs and order for build. We could use it, but it's scratched by CMake generated _Makefile_ file. Maybe rename the original one?
> 
> **(edited)**

Yes, I think we should do both. One rule per TA here, as well as use the TA Makefiles.

wamserma:
> > > making the dependencies list being likely built twice
> > 
> > 
> > From my point of view, this looks like a degradation.
> 
> True, buiding twice some pieces is not nice, but build env find out nothing more needs to be done for those components and it does not affect overall build time. Yet, this part and the way I use a local shell variable to create the list to deps paths is not very clean.
> 
> I have not reproduced the inital issue. @wamserma, do you confirm #389 ris still affecting build environments?

I can confirm. I tried to set up a nice reproducible dev environment with Nix and ran into the issue. On a pure Ubuntu builds work as expected.
Drilling down into the Makefiles led to opening #597.

etienne-lms:
@jforissier, I'm not shure I understand your suggestions.

> Yes, I think we should do both. One rule per TA here, 

Did you mean adding 1 buildroot make/install rule per TA implemented in the package,
in both _br-ext/.../optee_test_ext.mk_ and _optee_examples_ext.mk_ ?

> as well as use the TA Makefiles.

Buildroot recipe would invoke each TA _Makefile_ individually? or invoke a single _Makefile_ to make all repository TAs?

I think the buildroot recipe (or a yocto recipe) should run a single build that builds the package. We should not need to update a recipe because a TA was added in a optee repository and is embedded (unless there is specific configuration things needed in BR).
I would be in favor of renaming _ta/Makefile_ to _ta/make-install-tas.mk_ in optee_test and in optee_examples and have the buildroot recipe _optee_xxx_ext.mk_ to run a single `make -f make-install-tas.mk` or like command.

jforissier:
I mean generate Makefile targets instead of generating a list of build commands:
```
A:
        make -C A/
all: A

B:
        make -C B/
all: B
```
...instead of:
```
all:
        ...
        make -C A/
        make -C B/
```

etienne-lms:
With buildroot integration, we are expected to define 1 make variable named `OPTEE_TEST_EXT_POST_BUILD_HOOKS` to add commands to a build make target that is handled from buildroot env. We cannot add targets on a package make recipe in buildroot, like _package/foo/foo.mk_.
Same for installation.

etienne-lms:
TA dev kit does not support CMake for now. We would need to launch a gnu make command from a CMake build env.

PR #601, opened 2022-11-08
trusted-keys.exp: fix error detection

jforissier:
The check_keyctl_result() procedure doesn't deal with all error cases properly. For example the following error is not reported:
```
 # keyctl add trusted kmk "new 32" @u
 [    7.293607] trusted-key-tee optee-ta-f04a0fe7-1f5d-4b9b-abf7-619b85b4ce8c: key shm register failed
 add_key: Invalid argument
 #
```
To make things simpler, rely on the exit status of the commands sent to the guest. The || shell construct is used to trigger the display of an error message ("FAILED"). To avoid having the message in the command themselves, which would make the parsing more complex, a shell function is introduced ("fail"). Then all the check_keyctl_result() procedure needs to do is wait for the next prompt ("# "), while dealing with errors (the "FAILED" word appearing anywhere), key ID (parsing is unchanged), and timeouts (which I have reduced to 5 seconds, more than enough for any operation in this test suite).

### DIFF
--- a/trusted-keys.exp
+++ b/trusted-keys.exp
@@ -4,25 +4,18 @@
 # source. The return code is 0 for success, >0 for error.
 #
 
-set timeout 60
+set timeout 5
 set tk_id 0
 set ek_id 0
+# Wait for next prompt, dealing with key ID, failure message and timeout
 proc check_keyctl_result arg {
 	expect {
-		-re "(\n)(\\d+)\r" {
-			set ::$arg $expect_out(2,string)
+		-re {(\d+)\r} {
+			set ::$arg $expect_out(1,string)
 			exp_continue
 		}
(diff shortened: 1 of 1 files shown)


jforissier:
CC @b49020

b49020:
Thanks @jforissier, FWIW:

`Reviewed-by: Sumit Garg <sumit.garg@linaro.org>`

jforissier:
Tags applied + new commit added, please have a look.

b49020:
For commit: `trusted-keys.exp: skip test if trusted keys are not supported`

`Reviewed-by: Sumit Garg <sumit.garg@linaro.org>`

jenswi-linaro:

PR #602, opened 2022-11-14
Add support for ROCK 4 (aka Rockpi4)

jforissier:
Add rockpi4.mk and as well as U-Boot and Linux kernel config fragments plus one TF-A patch (fix console baudrate) and one U-Boot patch (enable OP-TEE in the DT).

Similar to other platforms such as HiKey, typical build commands are:

  $ make toolchains
  $ make -j$(nproc)
  $ make flash

The bootable image is out/rockpi4.img (128 MB). "make flash" transfers it to the onboard eMMC, but it may also be written to a micro SD card. Note however that the board tries to boot from eMMC first, so once "make flash" has been used, the eMMC has to be cleared to be able to boot from micro SD again. This can be done with "make nuke-emmc".

Tested on Rockpi4B.

Link: https://wiki.radxa.com/Rock4


rockpi4.mk
```diff
@@ -0,0 +1,239 @@
...
+else
+TF_A_LOGLVL ?= 40
+TF_A_OUT = $(TF_A_PATH)/build/rk3399/debug
+endif
+
+TF_A_FLAGS ?= ARCH=aarch64 PLAT=rk3399 SPD=opteed DEBUG=$(TF_A_DEBUG) \
+	      LOG_LEVEL=$(TF_A_LOGLVL)
+
+$(TF_A_PATH)/.stamp_patched:
+	patch -d $(TF_A_PATH) -p1 <$(ROOT)/build/rockpi4/tfa/0001-rk3399-set-console-baudrate-to-1500000.patch
```
jbech-linaro, line 60: It this something in the works? I.e., looks like this is something that should be sent upstream, or?

rockpi4.mk
```diff
@@ -0,0 +1,239 @@
...
+UBOOT_FLAGS ?= CROSS_COMPILE=$(CROSS_COMPILE_NS_KERNEL) \
+	       CC=$(CROSS_COMPILE_NS_KERNEL)gcc \
+	       HOSTCC="$(CCACHE) gcc"
+
+UBOOT_EXPORTS ?= BL31=$(TF_A_OUT)/bl31/bl31.elf TEE=$(OPTEE_OS_BIN)
+
+u-boot-defconfig: $(UBOOT_PATH)/.config
+
+$(UBOOT_PATH)/.stamp_patched:
+	patch -d $(UBOOT_PATH) -p1 <$(ROOT)/build/rockpi4/u-boot/0001-rockpi4-add-OP-TEE-nodes.patch
```
jbech-linaro, line 88: Ditto, upstream the patch at some point?

jbech-linaro:
With or without comments addressed:

rockpi4.mk
jenswi-linaro: Hmm, someone has been "tainted" by yocto? ;-)

This must be quite annoying when working on TF-A.

jforissier: > Hmm, someone has been "tainted" by yocto? ;-)

Nooooo! I swear! :wink: 

> It this something in the works? I.e., looks like this is something that should be sent upstream, or?

Absolutely, I plan on sending a patch, but it may take some time until this gets into a release. Especially since I am not familiar with TF-A upstreaming at all (Gerrit etc.). In the meantime I thought this was an OK-ish solution -- but note that even without this patch things should work, just without the initial TF-A and OP-TEE logs.

rockpi4.mk
jforissier: That one is much more problematic, because I think the node must be present only when OP-TEE is present (same problem as for the kernel, you don't want U-Boot to invoke an SMC if there's no EL3/SEL1).

jenswi-linaro:
I'm not overjoyed with these patches applied on top of gits. I would prefer using forks just as we do with the Linux kernel.

jforissier:
> I'm not overjoyed with these patches applied on top of gits. I would prefer using forks just as we do with the Linux kernel.

You know what, let's drop both patches. The TF-A console patch doesn't bring much, we can leave without the BL* and OP-TEE banners until it's fixed upstream. As for the OP-TEE DT nodes, they don't have to be in U-Boot, because U-Boot doesn't invoke OP-TEE in our dev config; it's good enough to patch the Linux DT which can be done with a proper commit in our fork. I will add a commit to https://github.com/linaro-swg/linux/pull/106.

rockpi4.mk
```diff
@@ -0,0 +1,231 @@
+COMPILE_NS_USER ?= 64
+override COMPILE_NS_KERNEL := 64
+COMPILE_S_USER ?= 64
+COMPILE_S_KERNEL ?= 64
+
+CFG_TEE_CORE_LOG_LEVEL ?= 2
```
jenswi-linaro, line 6: Why is this needed?

rockpi4.mk
```diff
@@ -0,0 +1,231 @@
...
+	$(MAKE) -C $(RKDEVELOPTOOL_PATH) clean
+
+rkdeveloptool-distclean:
+	$(MAKE) -C $(RKDEVELOPTOOL_PATH) clean
+
+clean: rkdeveloptool-clean
+
+$(LOADER_BIN):
+	cd $(BINARIES_PATH) && \
+		wget https://dl.radxa.com/rockpi/images/loader/$(notdir $(LOADER_BIN))
```
jenswi-linaro, line 200: We should probably remove the downloaded file if this fails for some reason, network timeout in the middle for instance.

rockpi4.mk
```diff
@@ -0,0 +1,231 @@
...
+
+nuke-emmc: $(LOADER_BIN) $(RKDEVELOPTOOL_BIN)
+	@echo
+	@echo "** WARNING: this command will make the onboard eMMC unbootable!"
+	@echo "It can be used to boot from the SD card again."
+	$(call flash-help)
+	dd if=/dev/zero of=$(BINARIES_PATH)/zero.img bs=1M count=64
+	$(RKDEVELOPTOOL_BIN) db $(LOADER_BIN)
+	sleep 1
+	$(RKDEVELOPTOOL_BIN) wl 0 $(BINARIES_PATH)/zero.img
```
jenswi-linaro, line 231: Is it possible to persuade `wl` to read from `stdin` instead? That way we wouldn't need a temporary file.

rockpi4.mk
jforissier: It is not. Will remove.

rockpi4.mk
jforissier: I agree. I will add `.DELETE_ON_ERROR` to take care of that.

rockpi4.mk
jforissier: Apparently not. The `rkdeveloptool` command needs a file name, it won't read from stdin. Worse: it needs a seekable file -- it does `lseek(... SEEK_END)` -- so bash process substitution `<()` won't work :-/

jenswi-linaro: OK, thanks for looking into it.

rockpi4.mk
```diff
@@ -195,6 +193,9 @@ rkdeveloptool-distclean:
 
 clean: rkdeveloptool-clean
 
+# In case wget fails mid-way
+.DELETE_ON_ERROR:
```
jenswi-linaro, line 197: This is a global setting, affecting all targets that can fail. I'm not sure where it's best to put it, but this close to the end of the file should make it hard to notice.

jforissier: > This is a global setting, affecting all targets that can fail.

Correct.

> I'm not sure where it's best to put it, but this close to the end of the file should make it hard to notice.

I can move it up

jforissier:
Updated

jenswi-linaro:

TaihuLight:
@jforissier @jbech-linaro @jenswi-linaro 
 
The newest plat-rockchip: [RK3588](https://www.rutronik.com/fileadmin/Rutronik/Downloads/printmedia/products/01_semiconductors/Rockchip_Flyer_Microprocessor_Highlights_Online.pdf) also support ARM Trustzone. It is interesting and necessary to build and run OP-TEE on the new board [ROCK5](https://wiki.radxa.com/Rock5) with RK3588 for several developers and users of OP-TEE. However, does the current version of OP-TEE support RK3588? How to build it on Rock 5 running Ubuntu 20.04?

The following issues may be helpful?
#https://github.com/OP-TEE/optee_os/issues/4197

[Rockchip_RK3588_Highlights.pdf](https://github.com/OP-TEE/optee_os/files/10128303/Rockchip_RK3588_Highlights.pdf)

PR #603, opened 2022-11-15
ffa: Allow multiple SPs with same UUID

jellesels-arm:
Change the naming scheme of the SP binaries to make sure we can add multiple SPs with the same UUID to the system. 
This is need to support [PR](https://github.com/OP-TEE/optee_os/pull/5650)

### DIFF
--- a/fvp/bl2_sp_list.dtsi
+++ b/fvp/bl2_sp_list.dtsi
@@ -48,4 +48,8 @@ test_sp3 {
 	load-address = <0x7a20000>;
 };
 
+test_sp4 {
+	uuid = "23eb0100-e32a-4497-9052-2f11e584afa6";
+	load-address = <0x7a30000>;
+};
 #endif
--- a/fvp/bl2_sp_list.dtsi
+++ b/fvp/bl2_sp_list.dtsi
@@ -3,6 +3,7 @@
  * Copyright (c) 2022, Arm Limited. All rights reserved.
  */
 
+#ifndef CFG_SPMC_TESTS
 internal_trusted_storage {
 	uuid = "dc1eef48-b17a-4ccf-ac8b-dfcff7711b14";
 	load-address = <0x7a00000>;
@@ -29,3 +30,22 @@ smm_gateway {
 	uuid = "ed32d533-99e6-4209-9cc0-2d72cdd998a7";
 	load-address = <0x7e00000>;
 };
+
+#else
+
+test_sp1 {
+	uuid = "5c9edbc3-7b3a-4367-9f83-7c191ae86a37";
+	load-address = <0x7a00000>;
+};
+
+test_sp2 {
+	uuid = "7817164c-c40c-4d1a-867a-9bb2278cf41a";
+	load-address = <0x7a10000>;
+};
+
+test_sp3 {
+	uuid = "23eb0100-e32a-4497-9052-2f11e584afa6";
+	load-address = <0x7a20000>;
+};
+
+#endif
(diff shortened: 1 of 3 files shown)


jbech-linaro:
Perhaps this has been discussed in the past, but why do we want to allow multiple SP's to have the same UUID in the first place? It's sounds wrong to me, but still I can see that this might make it easier from a (non-secure) user client point of view. I.e, a client sees "an interface" that belongs to a certain UUID, but under the hood on the secure side it has been split up into different SPs.

fvp/spmc_manifest.dts
```diff
@@ -63,6 +64,25 @@
 			uuid = <0x33d532ed 0x0942e699 0x722dc09c 0xa798d9cd>;
 			load-address = <0x0 0x7e00000>;
 		};
+
+#else /*BUILD_SPMC_TESTS*/
```
etienne-lms, line 68: update according to line 40.

fvp/spmc_manifest.dts
```diff
@@ -37,6 +37,7 @@
 #ifdef ARM_BL2_SP_LIST_DTS
 	sp_packages {
 		compatible = "arm,sp_pkg";
+#if !SPMC_TESTS
```
etienne-lms, line 40: `#ifndef CFG_SPMC_TESTS`?

fvp/bl2_sp_list.dtsi
```diff
@@ -3,6 +3,7 @@
  * Copyright (c) 2022, Arm Limited. All rights reserved.
  */
 
+#if !SPMC_TESTS
```
etienne-lms, line 6: `ifndef CFG_SPMC_TESTS`?

fvp/bl2_sp_list.dtsi
```diff
@@ -48,4 +48,9 @@ test_sp3 {
 	load-address = <0x7a20000>;
 };
 
+test_sp4 {
+	/* FF-A UUID */
```
etienne-lms, line 52: remove this inline comment? (for consistency within this DTSI file)

fvp/spmc_manifest.dts
```diff
@@ -81,6 +81,11 @@
 			load-address = <0x0 0x7a20000>;
 		};
 
+		test_sp4 {
+			/* SP binary UUID */
```
etienne-lms, line 85: Ditto

etienne-lms (commented):
Commit "FF-A: Add test SPs to the fip package", in commit message:
s/fip/FIP/g
s/include/includes/
s/doesn't has/doesn't have/


Commit "FF-A: Add spm-test4 test sp", in commit message.
"... ~~it's~~ **its** binary ~~then them~~ **than** the one ..."

For both commits I suggest to use "fvp-psa-sp: " instead of "FFA: " in commit message header line.

fvp/spmc_manifest.dts
jellesels-arm: Ack

fvp/spmc_manifest.dts
jellesels-arm: Ack

fvp/bl2_sp_list.dtsi
jellesels-arm: Ack

fvp/bl2_sp_list.dtsi
jellesels-arm: Ack

fvp/spmc_manifest.dts
jellesels-arm: Ack

PR #606, opened 2022-11-29
rockpi4: allow changing root FS size

jforissier:
Compute the sizes and offsets required when creating the boot image from BR2_TARGET_ROOTFS_EXT2_SIZE instead of assuming it is fixed to 112M. This allows enabling more Buildroot packages that would not fit into the default image size.


jbech-linaro:

PR #608, opened 2022-12-02
rockpi4: build and install kernel modules when LINUX_MODULES=y

jforissier:
Many things need kernel modules to work properly (for example, network access). Add support for this with "make LINUX_MODULES=y". It also takes care of configuring the ethernet port using DHCP.

This setting is off by default because for simple OP-TEE testing and development, it is not required, and we'd rather have a smaller image that builds and boots faster.

The kernel log level is set to 6 (KERN_INFO) instead of the default 7 (KERN_DEBUG) to avoid many messages from drivers which otherwise would disturb the session. They can still be seen with dmesg.


jbech-linaro:

jenswi-linaro:

PR #613, opened 2023-01-24
Alignment with trusted-services changes

imre-kis-arm:
This pull request multiple changes which are related to the recent changes around the trusted-services project.

- Setting MEASURED_BOOT=n causes TF-A build to fail as it expects a 0/1 value. The first patch fixes this issue.
- Block storage SP was added.
- The trusted-services deployments have been restructured, so the build repo has to follow the new directory structure and enable selecting different SP configurations.
- After updating TF-A to v2.8 the measured boot feature can be enabled with FIP stored SPs.
- Disable UEFI and SPMC tests by default because they require debugging features enabled.


jenswi-linaro (commented):

PR #619, opened 2023-01-30
qemu, qemu_v8: pass arguments to qemu_check.exp for example to exclude tests

jforissier:
This commit adds a way to exclude tests when running the xtest suite:
- qemu_check.exp now supports "--xtest-args <args>", which passes <args> to the xtest command,
- The CHECK_ARGS make variable is passed to qemu_check.exp.

For example to run the whole test suite but exclude xtest regression_1031:

 $ make check CHECK_ARGS="--xtest-args '-x regression_1031'"

### DIFF
--- a/qemu-check.exp
+++ b/qemu-check.exp
@@ -8,13 +8,14 @@
 #   -q        Suppress output to stdout (quiet)
 #   --tests   Type of tests to run, values: all, xtest and trusted-keys
 #   --timeout Timeout for each test (sub)case, in seconds [480]
+#   --xtest-args Optional arguments to xtest
 
 set bios "../out/bios-qemu/bios.bin"
-set cmd "xtest"
 set cmd1 "cd /mnt/host/build/qemu_v8/xen"
 set cmd2 "xl create guest.cfg"
 set cmd3 "xl console domu"
 set quiet 0
+set xtest_args ""
 
 # The time required to run some tests (e.g., key generation tests [4007.*])
 # can be significant and vary widely -- typically, from about one minute to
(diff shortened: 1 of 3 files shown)


etienne-lms (commented):
Works fine. By the way, maybe add it also in xen case in `qemu-check.exp`.
`Acked-by: Etienne Carriere <etienne.carriere@linaro.org>`.

However we must play with quotes. How about such syntax:
```shell
$ make check XTEST_ARGS="-t regression 1004 1013 1031"
```
with:
```
check-args += --xtest-args $(XTEST_ARGS)
```
?

jforissier:
> Works fine. By the way, maybe add it also in xen case in `qemu-check.exp`.

Right, I forgot that one. Will fix.

> `Acked-by: Etienne Carriere <etienne.carriere@linaro.org>`.
> 
> However we must play with quotes. How about such syntax:
> 
> ```shell
> $ make check XTEST_ARGS="-t regression 1004 1013 1031"
> ```
> 
> with:
> 
> ```
> check-args += --xtest-args $(XTEST_ARGS)
> ```
> 
> ?

Yes, I wanted to be as generic as possible but it's probably not a good idea. Better simplify the most common use case. I will rework as suggested, thanks!

jforissier:
Update

etienne-lms (commented):
thanks

jforissier:
Tag applied. Note: I added missing double quotes: `check-args += --xtest-args "$(XTEST_ARGS)"` and also a conditional which is not strictly necessary but looks better for consistency with `CHECK_TESTS` and because it might be a problem to have `--xtest-args` with no parameter the day we add another option to `qemu-check.exp`.

PR #632, opened 2023-03-09
fvp: enable passing extra arguments

gabor-toth-arm:
Introduce a new variable to allow passing extra arguments to the FVP with environment variable or on the make command line. As an example this allows passing trace related configuration.


gabor-toth-arm:
> 

Corrected

PR #633, opened 2023-03-10
qemu_v8: qemu: configure slirp

ldts:
Without the development package installed in the host, the build system will not include slirp support.

This will cause an error when trying to configure netdev's 'user' during qemu boot.


jenswi-linaro:
So this will result in a configure error instead of silently ignoring the problem?

jenswi-linaro:
Good catch!

ldts:
thanks @jenswi-linaro. added your tag

PR #635, opened 2023-03-20
Enhance and fix TS NWd build integration 

gyuri-szing:
This PR:
-  enhances Trusted Services integration by allowing to pass extra arguments to cmake when building NWd executables. This comes handy when a layer above the build repo (e.g. CI system) needs to control the arguments.
- fixes the libts dependency handling of NWd executables. The current implementation is buggy and the intended build optimization does not work. See the commit message of the second commit for details.

### DIFF
--- a/trusted-services.mk
+++ b/trusted-services.mk
@@ -160,9 +160,15 @@ ffa-$1:
 	CROSS_COMPILE=$(subst $(CCACHE),,$(CROSS_COMPILE_NS_USER)) cmake -G"Unix Makefiles" \
 		-S $(TS_PATH)/deployments/$1/arm-linux -B $(TS_BUILD_PATH)/$1 \
 		-DCMAKE_INSTALL_PREFIX=$(TS_INSTALL_PREFIX) \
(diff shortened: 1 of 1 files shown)


jenswi-linaro:
Please drop the `Change-Id:` lines and add a `ts: ` prefix as the subject prefix, with that:

gyuri-szing:
> Please drop the `Change-Id:` lines and add a `ts: ` prefix as the subject prefix, with that: `Acked-by: Jens Wiklander <jens.wiklander@linaro.org>`
Fixed!
Thanks for the review!

jforissier:
I have replaced `TS:` with `ts:` (we prefer to always use lowercase in the subsystem prefix, while "normal" case should be used everywhere else) and merged the commits. Thanks!

gyuri-szing:
> I have replaced `TS:` with `ts:` (we prefer to always use lowercase in the subsystem prefix, while "normal" case should be used everywhere else) and merged the commits. Thanks!

Thank you for the review and the fix! Will use lowercase for the prefix next time.

PR #636, opened 2023-03-20
ffa: Allow multiple SPs with same UUID

jellesels-arm:
Change the naming scheme of the SP binaries to make sure we can add multiple SPs with the same UUID to the system. Added the possibility to compile all TS SPs with the same FF-A UUID (protocol UUID).
This is need to support [5898](https://github.com/OP-TEE/optee_os/pull/5898)
This is a follow-up on [603](https://github.com/OP-TEE/build/pull/603)

### DIFF
--- a/fvp-psa-sp.mk
+++ b/fvp-psa-sp.mk
@@ -12,6 +12,8 @@ SPMC_TESTS			?= n
 # Enable the "HArdware Volatile Entropy Gathering and Expansion" daemon to
 # overcome low-entropy conditions in the FVP
 BR2_PACKAGE_HAVEGED		?= y
+SPMC_TESTS			?= y
+TS_RPC_UUID			?= y
 
 # TS SP configurations
 DEFAULT_SP_CONFIG		?= default-opteesp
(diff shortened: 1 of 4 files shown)


jenswi-linaro:
In the commit message for "FF-A: Add test SPs to the fip package" you say:
_TF-A doesn't has a limited capacity_
I guess you mean:
_TF-A doesn't have an unlimited capacity_

For commit "FF-A: Add spm-test4 test sp" please use capital letters for SP
and also write SPs instead of sp's in the commit message.

For commit "TS: Use rpc protocol UUID", please use capital letters for RPC.

With that fixed:

jforissier:
> For commit "TS: Use rpc protocol UUID", please use capital letters for RPC.

...and aren't lowercase subsystem prefixes preferred? "ts:"

jenswi-linaro:
> > For commit "TS: Use rpc protocol UUID", please use capital letters for RPC.
> 
> ...and aren't lowercase subsystem prefixes preferred? "ts:", "ff-a:" (or even "ffa:")

Yes, good point.

jellesels-arm:
> > > For commit "TS: Use rpc protocol UUID", please use capital letters for RPC.
> > 
> > 
> > ...and aren't lowercase subsystem prefixes preferred? "ts:", "ff-a:" (or even "ffa:")
> 
> Yes, good point.

Done

PR #639, opened 2023-03-24
ts: Generate sp_layout.json after SP builds

imre-kis-arm:
TF-A requires to have the SPs in the same order in the FIP as it is defined in ARM_BL2_SP_LIST_DTS. The sp_layout.json controls the order of SPs in the FIP package. Previously the individual JSON fragments were appended to the sp_layout.json file at the end of the SP build. The final order of SP definitions in the sp_layout.json depended on the SP build order which was non-deterministic because of the parallel make jobs. To resolve this issue the JSON merging step is now done after all SPs have been built. The order will follow the order of build-sp calls in the fvp-psa-sp.mk which matches the order of SPs in ARM_BL2_SP_LIST_DTS.

### DIFF
--- a/trusted-services.mk
+++ b/trusted-services.mk
@@ -78,15 +78,21 @@ ffa-sp-all-clean: ffa-$1-sp-clean
 ffa-sp-all-realclean: ffa-$1-sp-realclean
 
 optee_os_sp_paths += $(TS_INSTALL_PREFIX)/$(SP_DIR)/bin/$3.$(SP_EXT)
+fip_sp_json_paths += $(TS_INSTALL_PREFIX)/$(SP_DIR)/json/$1.json
 endef
 
 ifeq ($(SP_PACKAGING_METHOD),embedded)
 # Add the list of SP paths to the optee_os config
 OPTEE_OS_COMMON_EXTRA_FLAGS += SP_PATHS="$(optee_os_sp_paths)"
 else ifeq ($(SP_PACKAGING_METHOD),fip)
+$(TS_INSTALL_PREFIX)/sp_layout.json: ffa-sp-all
+	python $(TS_PATH)/tools/python/merge_json.py $@ $(fip_sp_json_paths)
+
+optee-os-common: $(TS_INSTALL_PREFIX)/sp_layout.json
+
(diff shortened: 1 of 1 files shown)


jforissier (commented):

trusted-services.mk
```diff
@@ -78,12 +78,18 @@ ffa-sp-all-clean: ffa-$1-sp-clean
...
 endef
 
 ifeq ($(SP_PACKAGING_METHOD),embedded)
 # Add the list of SP paths to the optee_os config
 OPTEE_OS_COMMON_EXTRA_FLAGS += SP_PATHS="$(optee_os_sp_paths)"
 else ifeq ($(SP_PACKAGING_METHOD),fip)
+$(TS_INSTALL_PREFIX)/opteesp/json/sp_layout.json: ffa-sp-all
+	python $(TS_PATH)/tools/python/merge_json.py $@ $(fip_sp_json_paths)
+
+optee-os-common: $(TS_INSTALL_PREFIX)/opteesp/json/sp_layout.json
```
jenswi-linaro, line 91: This is a strange dependency. I can understand the `optee-os-common: ffa-sp-all` dependency above to avoid making changes in the main make file. However, this dependency is even broken, I assume that `ffa-sp-all` must complete before the `sp_layout.json` file can be generated.

jforissier: Isn't it the purpose of the recipe just above? (although the command does not contain the target which is ugly but I assume the Python script has a hard-coded output path?)

jenswi-linaro: You're right, I missed that.

jenswi-linaro:

imre-kis-arm:
I've added the Acked-by tags and I also made a slight modification in the `sp_layout.json`'s path. I moved it to `$(TS_INSTALL_PREFIX)` from `$(TS_INSTALL_PREFIX)/opteesp/json` because in case of a raw binary SPs, trusted-services would use `$(TS_INSTALL_PREFIX)/sp` as it's install directory and it would be misleading to place the `sp_layout.json` into the `opteesp` directory.
Please let me know if this is fine and the Acked-by tags are still stand for the PR.

imre-kis-arm:
Thank you!

PR #642, opened 2023-04-05
fvp: SP configuration fixes

imre-kis-arm:
There are couple issues around TS SPs which this patch intends to fix.

- Aligning selection logic for SPs across fvp-psa-sp.mk, bl2_sp_list.dtsi and spmc_manifest.dts so they all include the same set of SPs for a given configuration.
- Pass DTC_CPPFLAGS to TF-A build because it preprocesses bl2_sp_list.dtsi and spmc_manifest.dts which only works if the necessary defines are forwarded properly.
- Increase SPM test SP memory area because they overlap with the current 64kB setting.
- Make RPC protocol UUID feature backwards compatible

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -24,6 +24,7 @@ endef
 
 SHELL := bash
 BASH ?= bash
+PYTHON3 ?= python3
 ROOT ?= $(shell pwd)/..
 
 UNAME_M				:= $(shell uname -m)
@@ -344,7 +345,7 @@ buildroot: optee-os optee-rust
 	@rm -f ../out-br/build/optee_*/.stamp_*
 	@rm -f ../out-br/extra.conf
 	@$(call append-br2-vars,../out-br/extra.conf)
-	@(cd .. && python build/br-ext/scripts/make_def_config.py \
+	@(cd .. && $(PYTHON3) build/br-ext/scripts/make_def_config.py \
 		--br buildroot --out out-br --br-ext build/br-ext \
 		--top-dir "$(ROOT)" \
 		--br-defconfig build/br-ext/configs/optee_$(BUILDROOT_ARCH) \
@@ -373,7 +374,7 @@ buildroot-domu: optee-os
 	@rm -f ../out-br-domu/build/optee_*/.stamp_*
 	@rm -f ../out-br-domu/extra.conf
 	@$(call append-br2-vars,../out-br-domu/extra.conf)
-	@(cd .. && python build/br-ext/scripts/make_def_config.py \
+	@(cd .. && $(PYTHON3) build/br-ext/scripts/make_def_config.py \
 		--br buildroot --out out-br-domu --br-ext build/br-ext \
 		--top-dir "$(ROOT)" \
 		--br-defconfig build/br-ext/configs/optee_$(BUILDROOT_ARCH) \
--- a/fvp-psa-sp.mk
+++ b/fvp-psa-sp.mk
@@ -36,6 +36,19 @@ TF_A_FLAGS ?= \
 include fvp.mk
 include trusted-services.mk
 
+# The macros used in bl2_sp_list.dts and spmc_manifest.dts has to be passed to
+# TF-A because it handles the preprocessing of these files.
+define add-dtc-define
+DTC_CPPFLAGS+=-D$1=$(subst y,1,$(subst n,0,$($1)))
+endef
+
+ifeq ($(SP_PACKAGING_METHOD),fip)
+$(eval $(call add-dtc-define,SPMC_TESTS))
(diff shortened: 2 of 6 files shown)


jenswi-linaro (commented):

gyuri-szing:
Before merging these I would like to discuss a python related issue. There is no commit in this stack yet, but if maintainers and Imre agrees, a new one could be added.

Currently the way python is called from makefiles is a bit inconsistent. At some places the script is executed directly and the "shebang" will select the interpreter while at other places `python` or in case of build-root `${PYTHON3}` is used. 
The current op-tee build environment set-up instructions recommend installing "python-is-python3"  on Ubuntu 22.04 and in this environment all python script will run with python3. On 20.04 if python2 is installed (no explicate requirement for this) scripts called like `python <script name>`  will run with python2. (I don't have a 20.04 system at hand and I might be wrong here). This is not an issue in most cases as scripts called with `python <name>` seem to be Pyhton2 compatible. **Except the TS specific `merge_json.py`.** But on "recent" enough system things play out well even for that script.

The above leaves a bit too much flux in how python is called and leaves too much unspecified behavior to the environment. I mean if someone is not using the recommended Ubuntu versions with the recommended settings, thing may easily go south. I understand that "non recommended" build environment are up to the user to get working, but having a wrapper in makefiles allowing to adapt to the environment and to control which interpreter is called would be a big help. 

**My points are**:
  - the `merge_json.py` is called incorrectly and this needs a fix
  - while fixing that I suggest introducing a PYTHON3 variable and to call al python scripts using that. The default value could be `python` for backwards compatibility or `python3` to be compliant to [PEP394](https://peps.python.org/pep-0394/)
  - I did not checked other repositories (like `optee_os`) how those use `python` and some changes in those might be needed too. 


---

**Some background info**:

_This is how python is called in the build repo_:
```
gyuri@Ubul:~/work/ts/v321-test$grep -nir -e "\.py" -e "python" --exclude "*.py" build
build/br-ext/package/optee_examples_ext/optee_examples_ext.mk:6:OPTEE_EXAMPLES_EXT_DEPENDENCIES = optee_client_ext host-python-cryptography
build/br-ext/package/optee_examples_ext/optee_examples_ext.mk:15:                       PYTHON3=$(HOST_DIR)/bin/python3 \
build/br-ext/package/optee_test_ext/optee_test_ext.mk:6:OPTEE_TEST_EXT_DEPENDENCIES = optee_client_ext openssl host-python-cryptography
build/br-ext/package/optee_test_ext/optee_test_ext.mk:36:                       PYTHON3=$(HOST_DIR)/bin/python3 \
build/hikey.mk:350:     $(ROOT)/burn-boot/hisi-idt.py --img1=$(LLOADER_PATH)/recovery.bin
build/qemu_v8.mk:249:   export WORKSPACE=$(EDK2_PATH) PYTHON3_ENABLE=TRUE
build/trusted-services.mk:89:   python $(TS_PATH)/tools/python/merge_json.py $@ $(fip_sp_json_paths)
build/common.mk:347:    @(cd .. && python build/br-ext/scripts/make_def_config.py \
build/common.mk:376:    @(cd .. && python build/br-ext/scripts/make_def_config.py \
build/common.mk:490:            $(LAUNCH_TERMINAL) "$(BUILD_PATH)/soc_term.py $(1)" &
build/common.mk:498:    $(gnome-terminal) -t $(2) -x $(BUILD_PATH)/soc_term.py $(1) &
build/common.mk:504:    $(xterm) -title $(2) -e $(BASH) -c "$(BUILD_PATH)/soc_term.py $(1)" &
build/zynqmp.mk:58:     ${OPTEE_OS_PATH}/scripts/gen_tee_bin.py --input ${OPTEE_OS_PATH}/out/arm/core/tee.elf --out_tee_raw_bin ${OPTEE_OS_PATH}/out/arm/core/tee_raw.bin
build/toolchain.mk:33:  @(cd .. && python build/br-ext/scripts/make_def_config.py \
```

As you can see Buildroot already uses the PYTHON3 variable although that is a separate "name space" from OP-TEE/build makefiles.

_Motivation_

I started investigating this because my really old build container uses ubuntu 18.04 and there `python` refers to python2 which makes executing `merge_joson.py` fail. As as it seems I can not easily fix that with the current makefiles. I do have a workaround but so the issues is not blocking, but a fix merged before the next release would be nice.

jforissier:
@gyuri-szing totally makes sense, and optee_os already uses the `$(PYTHON3) script.py` syntax everywhere, with `PYTHON3 ?= python3` in `mk/config.mk`.

gyuri-szing:
Ok. We are off till Tuesday. @imre-kis-arm please add a new commit to the PR fixing the makefiles. Thanks!

PR #643, opened 2023-04-05
rpi3: don't strip kernel8.img

jbech-linaro:
In one of the final build stages, buildroot will strip [1] binaries by running the following on all binaries under ./out-br/target:

  $ aarch64-linux-gnu-strip --remove-section=.comment \
    --remove-section=.note

When doing that on kernel8.img (which should be a 1:1 copy of ./linux/arch/arm64/boot/Image), the expected "magic" from the kernel header disappears and so does a few other things as well. In short, the kernel image becomes corrupt. Without the magic, U-boot won't pass execution to Linux and will instead fail with the error message: "Bad Linux ARM64 Image magic!"

To prevent kernel8.img from being stripped, we include it in the BR2_STRIP_EXCLUDE_FILES Makefile parameter.

Why this suddenly started to show, I haven't been able to figure out.

Link: [1] https://github.com/buildroot/buildroot/blob/71a2706320dbc22b16397b70b2d53b903bde0752/Makefile#L620-L650
Fixes: https://github.com/OP-TEE/optee_os/issues/5891

Tested-by: Joakim Bech <joakim.bech@linaro.org>

### DIFF
--- a/rpi3.mk
+++ b/rpi3.mk
@@ -22,6 +22,10 @@ BR2_PACKAGE_OPENSSH ?= y
 BR2_PACKAGE_OPENSSH_SERVER ?= y
 BR2_PACKAGE_OPENSSH_KEY_UTILS ?= y
 
+# We don't want buildroot to strip kernel8.img, since that will remove the
+# "magic" from the kernel header.
+BR2_STRIP_EXCLUDE_FILES ?= kernel8.img
+
 OPTEE_OS_PLATFORM = rpi3
 
 include common.mk


jenswi-linaro:
Why do we store the kernel image under `out-br/target`? If we store all sorts of binaries there and buildroot thinks it should try to strip stuff there we may be in for a game of cat and mouse.

jforissier:
FWIW, for RockPi4 we assemble the root FS image and the kernel into the final image in the `boot-img` recipe: https://github.com/OP-TEE/build/blob/1e36c12aaece8968a37c90d2cdaf4fd3446615f5/rockpi4.mk#L178-L200

jbech-linaro:
> Why do we store the kernel image under `out-br/target`? 

I checked the commit where this was introduced and it was I who did it, but I cannot remember why.

> If we store all sorts of binaries there and buildroot thinks it should try to strip stuff there we may be in for a game of cat and mouse.

Indeed. I think we should rework this a bit. In this setup we create "one" rootfs file, which "has it all". But when we actually create the partitions, we extract the boot stuff from that rootfs file into a "boot" partition and then we create another "rootfs" partition with the rootfs itself. Which is a bit weird. I could and should have boot partition files in one file/location and the rootfs in another location.

What I would suggest is that we merge this patch so we have a working setup for 3.21 and then I'll make the updates as described above, what do you think about that?

jforissier:

jenswi-linaro:

jbech-linaro:
Squashed, rebased and tag(s) applied (should be) ready for merge! Thanks! After thinking a bit more about the upcoming approach, I want to make it so you simply just can dd an image to the SD card, that will create both partitions for you. That'll be my goal. I'll check what Jerome have above, but I think we've done similar things for other platforms in the past.

PR #647, opened 2023-05-04
toolchain.mk: do not check server certificate when downloading

jforissier:
"make toolchains" is now failing with the following error:

 curl: (60) SSL certificate problem: unable to get local issuer certificate
 More details here: https://curl.se/docs/sslcerts.html

 curl failed to verify the legitimacy of the server and therefore could not
 establish a secure connection to it. To learn more about this situation and
 how to fix it, please visit the web page mentioned above.

Although I could download the GCC package [1] just fine with Firefox, the same URL does indeed cause an error when downloaded with curl on my laptop (I am using Ubuntu 22.04.2). The same error happens in the OP-TEE CI loop.

It looks like a required CA certificate ("GlobalSign RSA OV SSL CA 2018") is missing in /etc/ssl/certs/ca-certificates.crt.

Add -k to the curl command to disable certificate verification.

Link: [1] https://developer.arm.com/-/media/Files/downloads/gnu/11.3.rel1/binrel/arm-gnu-toolchain-11.3.rel1-x86_64-aarch64-none-linux-gnu.tar.xz

### DIFF
--- a/toolchain.mk
+++ b/toolchain.mk
@@ -14,7 +14,7 @@ define dltc
 	@if [ ! -d "$(1)" ]; then \
 		echo "Downloading $(3) ..."; \
 		mkdir -p $(1); \
-		curl --retry 5 -s -S -L $(2) -o $(TOOLCHAIN_ROOT)/$(3).tar.xz || \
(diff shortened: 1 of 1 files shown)


jenswi-linaro (commented):

PR #650, opened 2023-05-15
common.mk: Remove deprecated legacy toolchains

gagachang:
The legacy toolchain support is added in commit:
92582b5 ("Support legacy compiler")
However, it is removed later in commit:
6a0fc2b ("toolchain: Remove legacy toolchains")

Since legacy toolchain is removed, we also remove unnecessary variables in common.mk.


jenswi-linaro:
@jforissier, @jbech-linaro, it should be OK to remove the legacy toolchain now, right?

jenswi-linaro:

jbech-linaro:
Yes, agree, this should be OK, thanks!

PR #651, opened 2023-05-16
toolchain.mk: Check whether the file exists when making file links

gagachang:
Not all toolchains have *-none-linux* format executables. If no file matches that pattern, the error occurs when the next "ln" command is executed. Fix it by checking whether any file name matches the pattern or not before executing "ln" command.


toolchain.mk
```diff
@@ -19,7 +19,11 @@ define dltc
 		tar xf $(TOOLCHAIN_ROOT)/$(3).tar.xz -C $(1) --strip-components=1 || \
 			{ rm $(TOOLCHAIN_ROOT)/$(3).tar.xz; echo Downloaded file is damaged; \
 			cd $(TOOLCHAIN_ROOT) && rm -rf $(1); exit 1; }; \
-		(cd $(1)/bin && for f in *-none-linux*; do ln -s $$f $${f//-none} ; done;) \
+		(cd $(1)/bin && for f in *-none-linux*; do \
+			if [ -e $$f ]; then \
+				ln -s $$f $${f//-none}; \
+			fi; \
+		done;) \
```
jforissier, line 26: There is a slightly simpler way I think, could you try this?
```suggestion
		(cd $(1)/bin && shopt -s nullglob && for f in *-none-linux*; do ln -s $$f $${f//-none} ; done;) \
```

gagachang:
Apply simpler solution.

jenswi-linaro:

gagachang:
Tags applied!

PR #653, opened 2023-06-08
rpi3: create a bootable image

jbech-linaro:
At [1] we temporarily fixed the issue with the stripped header from the kernel image. There we also suggested to make a future change making it possible to just simply 'dd' an image to the SD-card.

This change introduces a bash script that generates a rootfs partition and a boot partition. The rootfs partition will contain the root filesystem as created by Buildroot, while the boot partition will contain the firmware files placed in a temporary folder, that will be used to created the boot partition. To create the boot partition, we leverage the 'genimage' tool, hence we enable and build that as host tool in Buildroot.

The change removes the need to manually create, mount partitions and copy files to them. With this, it is sufficient to simply 'dd' the generated image to the SD-card. The 'make img-help' target has been kept, but now it's basically just a 'dd' line instead. People can also use other tools to write the image to a SD-card, such as Balena Etcher and similar tools.

Link: [1] https://github.com/OP-TEE/build/pull/643

Tested-by: Joakim Bech <joakim.bech@linaro.org>

### DIFF
--- a/rpi3.mk
+++ b/rpi3.mk
@@ -22,13 +22,18 @@ BR2_PACKAGE_OPENSSH ?= y
 BR2_PACKAGE_OPENSSH_SERVER ?= y
 BR2_PACKAGE_OPENSSH_KEY_UTILS ?= y
 
-# We don't want buildroot to strip kernel8.img, since that will remove the
-# "magic" from the kernel header.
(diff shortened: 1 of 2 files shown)


jforissier (approved):
LGTM

jbech-linaro:

PR #655, opened 2023-06-12
Update hikey.mk

TaihuLight:
Fix the issue about building OP-TEE 3.21.0 for Hikey with the default toolchain version 11.3

Fixed by @jforissier
Tested by Longlong LIAO [liaoll@hku.hk](mailto:liaoll@hku.hk)

### DIFF
--- a/hikey.mk
+++ b/hikey.mk
@@ -134,6 +134,7 @@ endef
 
 .PHONY: edk2
 edk2:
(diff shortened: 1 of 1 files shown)


jforissier:
Please fix both the Hikey and Hikey960 issues in the same commit in the same PR (654) and make the subject of the commit:
"hikey, hikey960: fix EDK2 compilation error with GCC 11.3".
We also need your `Signed-off-by: Your Name <your@email.address>`

PR #657, opened 2023-06-14
FVP, TS: optimize Linux and Buildroot config

balint-dobszay-arm:
Tweak the Linux and Buildroot configs to only build the minimal required feature set for this particular use case. This should significantly speed up the overall build.


jforissier (commented):

fvp-psa-sp.mk
```diff
@@ -12,6 +12,22 @@ SPMC_TESTS			?= n
...
+# Disable packages not used by this configuration
+BR2_PACKAGE_HOST_E2FSPROGS	?= n
+BR2_PACKAGE_KEYUTILS		?= n
+BR2_PACKAGE_MMC_UTILS		?= n
+BR2_PACKAGE_OPENSC		?= n
+BR2_PACKAGE_OPTEE_EXAMPLES_EXT	?= n
+BR2_PACKAGE_STRACE		?= n
+
+# Building xtest is not necessary if we don't want to run the SPMC tests
+ifeq ($(SPMC_TESTS),n)
```
etienne-lms, line 25: I would rather suggest `ifneq ($(SPMC_TESTS),y)`

etienne-lms (commented):
my 2 cents here

fvp-psa-sp.mk
balint-dobszay-arm: Done

PR #665, opened 2023-07-18
qemu_v8: restore ENABLE_SME_FOR_SWD=1 in TF_A_FLAGS

jenswi-linaro:
ENABLE_SME_FOR_SWD=1 was previously removed on the faulty assumption that 337ff4f1dd66 ("fix(qemu): enable SVE and SME") was from now on setting ENABLE_SME_FOR_SWD=1 by default. That's not correct, so restore ENABLE_SME_FOR_SWD=1 in TF_A_FLAGS.

Fixes: d217c656517d ("qemu_v8: remove ENABLE_SME_FOR_*=1 from TF_A_FLAGS")

This should take care of the recent CI timeouts.


jforissier:

jenswi-linaro:
Tag applied

PR #668, opened 2023-07-23
Fix download cross compiler toolchain failed.

Ricardo-Tu:
The previous execution of "make -f toolchain.mk  toolchains" command failed to download the compiler toolchains,and this commit fixed  this error.


jforissier:
`curl` is in the [prerequisites](https://optee.readthedocs.io/en/latest/building/prerequisites.html) to build OP-TEE, `wget` is not. Therefore I think this change is not a good idea.

PR #669, opened 2023-07-25
Support RISC-V targets in common.mk

gagachang:
This PR adds some variables for RISC-V architecture.
The `TARGET_ARCH` variable can be used to specify the architecture to be built.
Some configuration file for buildroot are also added.


common.mk
```diff
@@ -517,13 +512,54 @@ endef
...
+
 OPTEE_OS_COMMON_FLAGS ?= \
 	$(OPTEE_OS_COMMON_EXTRA_FLAGS) \
 	PLATFORM=$(OPTEE_OS_PLATFORM) \
 	CROSS_COMPILE=$(CROSS_COMPILE_S_USER) \
 	CROSS_COMPILE_core=$(CROSS_COMPILE_S_KERNEL) \
 	CROSS_COMPILE_ta_arm64="$(CCACHE)$(AARCH64_CROSS_COMPILE)" \
 	CROSS_COMPILE_ta_arm32="$(CCACHE)$(AARCH32_CROSS_COMPILE)" \
+	CROSS_COMPILE_ta_rv64="$(CCACHE)$(RISCV64_CROSS_COMPILE)" \
+	CROSS_COMPILE_ta_rv32="$(CCACHE)$(RISCV32_CROSS_COMPILE)" \
```
jforissier, line 562: It doesn't make much sense to have both arm and riscv flags here. I suggest defining `$(TA_CROSS_COMPILE_FLAGS)` to deal with that.

common.mk
```diff
@@ -28,6 +28,7 @@ PYTHON3 ?= python3
 ROOT ?= $(shell pwd)/..
 
 UNAME_M				:= $(shell uname -m)
+TARGET_ARCH			?= arm
```
jforissier, line 31: Could this be just `$(ARCH)`? I know the Linux kernel uses the same name and may need a different value (the Linux `ARCH` may be `arm` or `arm64` when this `TARGET_ARCH` is `arm`) but I don't think it is an issue at all since `ARCH` is explicitly provided in `LINUX_*_FLAGS`.
What do others think?

jforissier (commented):
Comments below for "common.mk: Add TARGET_ARCH variable and add riscv related variables". Please s/riscv/RISC-V/ in the commit subject. With that this commit is:

jforissier:
For commit "buildroot: Add toolchain configuration file for RISC-V 64-bit system":

jforissier:
For commit "buildroot: Add optee configuration file for RISC-V 64-bit system":

common.mk
gagachang: How about this instead of defining new variable

```
OPTEE_OS_COMMON_EXTRA_FLAGS += CROSS_COMPILE_ta_arm64="$(CCACHE)$(AARCH64_CROSS_COMPILE)"
OPTEE_OS_COMMON_EXTRA_FLAGS += CROSS_COMPILE_ta_arm32="$(CCACHE)$(AARCH32_CROSS_COMPILE)"
```

jforissier: That would work too but I like this option less.

gagachang: OK I will add another variable

gagachang: Added in reviewed commit

common.mk
gagachang: I am OK about this

gagachang: Changed in latest commits

jforissier:
@gagachang thanks, LGTM:

gagachang:
> @gagachang thanks, LGTM:
> 
> `Acked-by: Jerome Forissier <jerome.forissier@linaro.org>`

Thanks!
Tags applied

PR #670, opened 2023-07-27
qemu_v8: u-boot: fix load address of kernel and RAM disk

jforissier:
This commit changes the load address of the kernel and of the RAM disk when U-Boot is used (make UBOOT=y). It fixes the following boot error with v2023.07.02:

  ** Reading file would overwrite reserved memory **

- The kernel load address, ${kernel_addr_r}, is moved from 0x40400000 (the default defined in u-boot/board/emulation/qemu-arm/qemu-arm.env) to 0x42200000 because the current kernel size is too big to fit in the available space:
  0x40400000 + 38783552 (current uImage size, ~37 M) = 0x446FCA40
  Output of the U-Boot 'bdinfo':
   => bdinfo
   ...
   lmb_dump_all:
    memory.cnt = 0x1 / max = 0x10
    memory[0]      [0x40000000-0x820fffff], 0x42100000 bytes flags: 0
    reserved.cnt = 0x4 / max = 0x10
    reserved[0]    [0xe100000-0xeffffff], 0x00f00000 bytes flags: 4
    reserved[1]    [0x42000000-0x421fffff], 0x00200000 bytes flags: 4
    reserved[2]    [0x7f99b000-0x820fffff], 0x02765000 bytes flags: 0
    reserved[3]    [0x8099b3e0-0x820fffff], 0x01764c20 bytes flags: 0
  There is indeed an overlap with reserved[1].
- The RAM disk is loaded at 0x44000000 by default (${ramdisk_addr_r}).
With the above change it needs to be moved to a higher address, after
the end of the kernel (0x446FCA40). Let's take some margin and choose
0x45000000.


jforissier:
Depends on https://github.com/OP-TEE/manifest/pull/243 which needs to be merged first.

jforissier:
See 671.

PR #671, opened 2023-07-27
QEMU_v8: remove EDK2

jforissier:
~~Please ignore the first commit ("qemu_v8: u-boot: fix load address of kernel and RAM disk") which is submitted as https://github.com/OP-TEE/build/pull/670 and review the second one only ("qemu_v8: remove EDK2").~~ Edit: let's review both commits here.


jforissier:
As mentioned in the description of the second commit, this PR combined with the update of U-Boot and the removal of EDK2 from the manifest (https://github.com/OP-TEE/manifest/pull/244) does indeed fix the OP-TEE OS CI. Tested in my own fork: in https://github.com/jforissier/optee_os/actions/runs/5682619384.

jenswi-linaro:

PR #677, opened 2023-08-24
qemu_v8: add uImage target

jforissier:
"make linux" builds the kernel but is not sufficient to prepare all files for "make run-only". On the other hand, "make run" does a bit too many things to be used comfortably when working on the kernel.

Therefore, add target uImage which generates the kernel image ready to be used by U-Boot. With that, kernel development and testing can be done with the following sequence:

 1. make
 2. Edit kernel sources
 3. make -j$(nproc) uImage
 4. make run-only
 5. Go to 2 -- repeat as needed

### DIFF
--- a/qemu_v8.mk
+++ b/qemu_v8.mk
@@ -394,6 +394,9 @@ $(KERNEL_UIMAGE): u-boot linux | $(BINARIES_PATH)
 				-n "Linux kernel" \
 				-d $(BINARIES_PATH)/linux.bin $(KERNEL_UIMAGE)
 
+.PHONY: uImage
+uImage: $(KERNEL_UIMAGE)
+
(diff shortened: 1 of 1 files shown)


qemu_v8.mk
```diff
@@ -394,6 +394,8 @@ $(KERNEL_UIMAGE): u-boot linux | $(BINARIES_PATH)
 				-n "Linux kernel" \
 				-d $(BINARIES_PATH)/linux.bin $(KERNEL_UIMAGE)
```
jenswi-linaro, line 396: `.PHONY: uImage`

jforissier: Fixed

jenswi-linaro:

PR #678, opened 2023-08-24
qemu_v8: use QEMU_VIRT in check target

jenswi-linaro:
Export QEMU_VIRT in the check target and update the expect script to use QEMU_VIRT when adding the "virtualization" to QEMU. This enables the check target with Hafnium, that is, SPMC_AT_EL=2.


jforissier:

jenswi-linaro:
Tag applied

PR #682, opened 2023-09-13
qemu_v8: set CFG_TEE_CORE_NB_CORE equals QEMU_SMP

PeikanTsai:
Currently qemu_armv8a use default CFG_TEE_CORE_NB_CORE = 4. User need to override it as below command to bring-up more than 4 cores.

make run QEMU_SMP=6 CFG_TEE_CORE_NB_CORE=6

Let CFG_TEE_CORE_NB_CORE=$(QEMU_SMP) if QEMU_SMP is set to simplify the command as following.

make run QEMU_SMP=6


jforissier (commented):
One suggestion below, then make the subject line "qemu_v8: set CFG_TEE_CORE_NB_CORE equal QEMU_SMP by default", and in the description:
"
Set CFG_TEE_CORE_NB_CORE to $(QEMU_SMP) by default if
QEMU_SMP is set to simplify the command as following.

make run QEMU_SMP=6
"

With that:

PR #683, opened 2023-09-22
FVP TS fixes

balint-dobszay-arm:

### DIFF
--- a/fvp-psa-sp.mk
+++ b/fvp-psa-sp.mk
@@ -107,5 +107,25 @@ $(eval $(call build-ts-app,psa-api-test/initial_attestation))
 endif
 ifeq ($(TS_UEFI_TESTS),y)
 $(eval $(call build-ts-app,uefi-test))
+
+# uefi-test uses MM Communicate via the arm-ffa-user driver and the message
+# payload is forwarded in a carveout memory area. Adding reserved-memory node to
+# the device tree to prevent Linux from using the carveout area for other
+# purposes.
+
+ORIGINAL_DTB := $(FVP_LINUX_DTB)
+CARVEOUT_ENTRY = $(ROOT)/build/fvp/mm_communicate_carveout.dtsi
+FVP_LINUX_DTB = $(ROOT)/out/fvp_with_mm_carveout.dtb
+
+$(FVP_LINUX_DTB): $(CARVEOUT_ENTRY) | linux
+	{ dtc -Idtb -Odts $(ORIGINAL_DTB); cat $(CARVEOUT_ENTRY); } | dtc -Idts -Odtb -o $(FVP_LINUX_DTB)
+
+boot-img: $(FVP_LINUX_DTB)
+
+.PHONY: carveout-dtb-clean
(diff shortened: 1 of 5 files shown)


fvp-psa-sp.mk
```diff
@@ -107,5 +107,25 @@ $(eval $(call build-ts-app,psa-api-test/initial_attestation))
...
+# payload is forwarded in a carveout memory area. Adding reserved-memory node to
+# the device tree to prevent Linux from using the carveout area for other
+# purposes.
+
+ORIGINAL_DTB := $(FVP_LINUX_DTB)
+CARVEOUT_ENTRY = $(ROOT)/build/fvp/mm_communicate_carveout.dtsi
+FVP_LINUX_DTB = $(ROOT)/out/fvp_with_mm_carveout.dtb
+
+$(FVP_LINUX_DTB): linux $(CARVEOUT_ENTRY)
+	{ dtc -Idtb -Odts $(ORIGINAL_DTB); cat $(CARVEOUT_ENTRY); } | dtc -Idts -Odtb -o $(FVP_LINUX_DTB)
```
jenswi-linaro, line 121: Any chance that the carveout entry may end up being added more than once?

balint-dobszay-arm: Hmm interesting question. I think it's not possible, because the starting point is always the unmodified `$(ORIGINAL_DTB)`. So if the rule is ran more than once for some reason, the fact that the previous output dtb already has the entry doesn't matter, since we always start from scratch.

jenswi-linaro: The target `linux` depends on the `.PHONY` target `linux-common`, I suspect that causes the `$(FVP_LINUX_DTB)` to always be outdated.

jforissier:
```make
$(FVP_LINUX_DTB): $(CARVEOUT_ENTRY) | linux
```
...perhaps?

balint-dobszay-arm: > The target linux depends on the .PHONY target linux-common, I suspect that causes the $(FVP_LINUX_DTB) to always be outdated.

That's true, but I don't see how it could cause adding the carveout entry more than once. In this recipe we take the unmodified original from e.g. `linux/arch/arm64/boot/dts/arm/fvp-base-revc.dtb`, concatenate the carveout entry to it and write the results to `$(ROOT)/out/fvp_with_mm_carveout.dtb`. Running the rule multiple times only repeats this, overwriting the output with the same content. Or am I missing something?

balint-dobszay-arm: > ```$(FVP_LINUX_DTB): $(CARVEOUT_ENTRY) | linux```
...perhaps?

Good point, this fixes unnecessarily executing this rule on each run.

balint-dobszay-arm: Done

jenswi-linaro:

balint-dobszay-arm:
Any other comments on this?

jforissier:
@balint-dobszay-arm this is good, thanks!

PR #688, opened 2023-10-02
QEMU enablement for Trusted Services

stormset:
Add new build files to configure OP-TEE and TF-A for the S-EL1 SPMC scenario on QEMU platform for Trusted Services.

### DIFF
--- a/qemu-psa-sp.mk
+++ b/qemu-psa-sp.mk
@@ -1,0 +1,139 @@
+OUT_PATH		?= $(ROOT)/out
+
+QEMU_VIRTFS_HOST_DIR	?= $(ROOT)
+QEMU_VIRTFS_AUTOMOUNT	?= y
+
+QEMU_USE_CUSTOM_DTB		?= y
+
+MEASURED_BOOT			?= y
+MEASURED_BOOT_FTPM		?= n
+TS_SMM_GATEWAY			?= y
+TS_UEFI_TESTS			?= y
+# Supported values: embedded (fip is currently not supported on QEMU platform)
+SP_PACKAGING_METHOD		?= embedded
+SPMC_TESTS			?= n
+
+# Disable packages not used by this configuration
+BR2_PACKAGE_HOST_E2FSPROGS	?= n
+BR2_PACKAGE_KEYUTILS		?= n
+BR2_PACKAGE_MMC_UTILS		?= n
+BR2_PACKAGE_OPENSC		?= n
+BR2_PACKAGE_OPTEE_EXAMPLES_EXT	?= n
+BR2_PACKAGE_STRACE		?= n
+
+# Building xtest is not necessary if we don't want to run the SPMC tests
+ifneq ($(SPMC_TESTS),y)
+BR2_PACKAGE_OPTEE_TEST_EXT	?= n
+BR2_PACKAGE_LIBOPENSSL		?= n
+BR2_PACKAGE_OPENSSL		?= n
+endif
+
+# TS SP configurations
+DEFAULT_SP_CONFIG		?= default-opteesp
+SP_BLOCK_STORAGE_CONFIG		?= $(DEFAULT_SP_CONFIG)
+SP_PSA_ITS_CONFIG		?= $(DEFAULT_SP_CONFIG)
+SP_PSA_PS_CONFIG		?= $(DEFAULT_SP_CONFIG)
+SP_PSA_CRYPTO_CONFIG		?= $(DEFAULT_SP_CONFIG)
+SP_PSA_ATTESTATION_CONFIG	?= $(DEFAULT_SP_CONFIG)
+SP_SMM_GATEWAY_CONFIG		?= $(DEFAULT_SP_CONFIG)
+
+# SPMC runs at EL1
+SPMC_AT_EL ?= 1
+# Use GICv2 or GICv3
+GICV3 ?= y
+# Use TF-A's Trusted Board Boot mechanism?
+TF_A_TRUSTED_BOARD_BOOT ?= y
+
+TF_A_FLAGS ?= \
+	BL32=$(OPTEE_OS_PAGER_V2_BIN) \
+	BL33=$(BL33_BIN) \
+	PLAT=qemu \
+	SPD=spmd \
+	$(TF_A_FIP_SP_FLAGS) \
+	QEMU_USE_GIC_DRIVER=$(TFA_GIC_DRIVER) \
+	QEMU_TOS_FW_CONFIG_DTS=$(ROOT)/build/qemu_v8/spmc_el1_manifest.dts
+
+QEMU_EXTRA_ARGS ?= \
+	-cpu cortex-a57 \
+	-semihosting-config enable=on,target=native,userspace=on \
+	-d trace:'__*'
+
+include qemu_v8.mk
+include trusted-services.mk
+
+# The macros used in bl2_sp_list.dts and spmc_manifest.dts has to be passed to
+# TF-A because it handles the preprocessing of these files.
+define add-dtc-define
+DTC_CPPFLAGS+=-D$1=$(subst y,1,$(subst n,0,$($1)))
+endef
+
+ifeq ($(SP_PACKAGING_METHOD),fip)
+$(eval $(call add-dtc-define,SPMC_TESTS))
+$(eval $(call add-dtc-define,TS_SMM_GATEWAY))
+
+TF_A_EXPORTS += DTC_CPPFLAGS="$(DTC_CPPFLAGS)"
+endif
+
+OPTEE_OS_COMMON_EXTRA_FLAGS += \
+	CFG_SECURE_PARTITION=y \
+	CFG_CORE_HEAP_SIZE=131072 \
+	CFG_DT=y \
+	CFG_MAP_EXT_DT_SECURE=y
+
+SP_SMM_GATEWAY_EXTRA_FLAGS +=-DMM_COMM_BUFFER_ADDRESS="0x00000000 0x42000000"
+TS_APP_COMMON_FLAGS +=-DMM_COMM_BUFFER_ADDRESS="0x42000000"
+
+# The boot order of the SPs is determined by the order of calls here. This is
+# due to the SPMC not (yet) supporting the boot order field of the SP manifest.
+ifneq ($(SPMC_TESTS),y)
+# PSA SPs
+$(eval $(call build-sp,block-storage,config/$(SP_BLOCK_STORAGE_CONFIG),63646e80-eb52-462f-ac4f-8cdf3987519c,$(SP_BLOCK_STORAGE_EXTRA_FLAGS)))
+$(eval $(call build-sp,internal-trusted-storage,config/$(SP_PSA_ITS_CONFIG),dc1eef48-b17a-4ccf-ac8b-dfcff7711b14,$(SP_PSA_ITS_EXTRA_FLAGS)))
+$(eval $(call build-sp,protected-storage,config/$(SP_PSA_PS_CONFIG),751bf801-3dde-4768-a514-0f10aeed1790,$(SP_PSA_PS_EXTRA_FLAGS)))
+$(eval $(call build-sp,crypto,config/$(SP_PSA_CRYPTO_CONFIG),d9df52d5-16a2-4bb2-9aa4-d26d3b84e8c0,$(SP_PSA_CRYPTO_EXTRA_FLAGS)))
+ifeq ($(MEASURED_BOOT),y)
+$(eval $(call build-sp,attestation,config/$(SP_PSA_ATTESTATION_CONFIG),a1baf155-8876-4695-8f7c-54955e8db974,$(SP_PSA_ATTESTATION_EXTRA_FLAGS)))
+endif
+ifeq ($(TS_SMM_GATEWAY),y)
+$(eval $(call build-sp,smm-gateway,config/$(SP_SMM_GATEWAY_CONFIG),ed32d533-99e6-4209-9cc0-2d72cdd998a7,$(SP_SMM_GATEWAY_EXTRA_FLAGS)))
+endif
+else
+# SPMC test SPs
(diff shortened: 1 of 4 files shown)


jenswi-linaro:
I'm going to ask you the same thing as I asked the author of https://github.com/OP-TEE/build/pull/686 before he closed that PR.

It would be nice with a bit of background to this. Do you represent a company? Are you going to maintain this? Do you expect to contribute more?

stormset:
Hi! I was working at Arm recently on the TS project. I left, so I am making these contributions privately. I am available to maintain it and make other contributions. I thought it would be great to be able to use the project on QEMU too, instead of the proprietary FVP.

jenswi-linaro:
Got it, thanks for contributing.

jenswi-linaro:

qemu-psa-sp.mk
```diff
@@ -0,0 +1,143 @@
...
+# Building xtest is not necessary if we don't want to run the SPMC tests
+ifneq ($(SPMC_TESTS),y)
+BR2_PACKAGE_OPTEE_TEST_EXT	?= n
+BR2_PACKAGE_LIBOPENSSL		?= n
+BR2_PACKAGE_OPENSSL		?= n
+endif
+
+# TS SP configurations
+DEFAULT_SP_CONFIG		?= default-opteesp
+SP_BLOCK_STORAGE_CONFIG	?= $(DEFAULT_SP_CONFIG)
```
etienne-lms, line 31: indent?

qemu-psa-sp.mk
```diff
@@ -0,0 +1,143 @@
+OUT_PATH		?= $(ROOT)/out
+
+QEMU_VIRTFS_HOST_DIR = $(ROOT)
+QEMU_VIRTFS_AUTOMOUNT = y
```
etienne-lms, line 4: `?=` ?

qemu-psa-sp.mk
```diff
@@ -0,0 +1,143 @@
...
+UBOOT ?= n
```
etienne-lms, line 5: indent

qemu-psa-sp.mk
```diff
@@ -0,0 +1,143 @@
...
+	CFG_CORE_SEL1_SPMC=y \
+	CFG_CORE_HEAP_SIZE=131072 \
+	CFG_DT=y \
+	CFG_MAP_EXT_DT_SECURE=y
+
+TS_APP_COMMON_FLAGS +=-DMM_COMM_BUFFER_ADDRESS="0x42000000"
+
+# The boot order of the SPs is determined by the order of calls here. This is
+# due to the SPMC not (yet) supporting the boot order field of the SP manifest.
+ifeq ($(SPMC_TESTS),n)
```
etienne-lms, line 89: `ifneq ($(SPMC_TESTS),y`)?

qemu-psa-sp.mk
```diff
@@ -0,0 +1,143 @@
...
+# SPMC test SPs
+OPTEE_OS_COMMON_EXTRA_FLAGS	+= CFG_SPMC_TESTS=y
+$(eval $(call build-sp,spm-test1,opteesp,5c9edbc3-7b3a-4367-9f83-7c191ae86a37,$(SP_SPMC_TEST_EXTRA_FLAGS)))
+$(eval $(call build-sp,spm-test2,opteesp,7817164c-c40c-4d1a-867a-9bb2278cf41a,$(SP_SPMC_TEST_EXTRA_FLAGS)))
+$(eval $(call build-sp,spm-test3,opteesp,23eb0100-e32a-4497-9052-2f11e584afa6,$(SP_SPMC_TEST_EXTRA_FLAGS)))
+$(eval $(call build-sp,spm-test4,opteesp,423762ed-7772-406f-99d8-0c27da0abbf8,$(SP_SPMC_TEST_EXTRA_FLAGS)))
+endif
+
+# Linux user space applications
+ifeq ($(SPMC_TESTS),n)
```
etienne-lms, line 111: `ifneq ($(SPMC_TESTS),y)`?

etienne-lms (commented):
some comment.
Otherwise  `Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>`.

qemu-psa-sp.mk
stormset: Fixed

qemu-psa-sp.mk
stormset: Fixed

qemu-psa-sp.mk
stormset: Fixed

qemu-psa-sp.mk
stormset: The same pattern is used for the build file for FVP (`fvp-psa-sp.mk`), so it might be better to keep it for easier maintenance.

qemu-psa-sp.mk
stormset: The same pattern is used for the build file for FVP (`fvp-psa-sp.mk`), so it might be better to keep it for easier maintenance.

qemu-psa-sp.mk
etienne-lms: ok

qemu-psa-sp.mk
etienne-lms: ok

jforissier: Testing against `y` is better IMO. I wouldn't mind a subsequent commit to fix all occurrences.

jforissier:
@stormset do you plan to address the review comments? Thanks.

stormset:
> @stormset do you plan to address the review comments? Thanks.

Hi! I'm really sorry for the delay. I noticed another issue with the PR, namely the handling of the reserved memory location (for the carveout) in the device tree. I have a solution that dumps out the device tree (which is generated by QEMU when the 'virt' platform is used) and passes the modified one to QEMU. Looking at [this](https://patchwork.kernel.org/project/qemu-devel/patch/20210926183410.256484-1-sjg@chromium.org) patch, I don't see any better solution. I still haven't tested it properly, but I will do it this week and also address the other comments.

jforissier:
> > @stormset do you plan to address the review comments? Thanks.
> 
> Hi! I'm really sorry for the delay. I noticed another issue with the PR, namely the handling of the reserved memory location (for the carveout) in the device tree. I have a solution that dumps out the device tree (which is generated by QEMU when the 'virt' platform is used) and passes the modified one to QEMU. Looking at [this](https://patchwork.kernel.org/project/qemu-devel/patch/20210926183410.256484-1-sjg@chromium.org) patch, I don't see any better solution. I still haven't tested it properly, but I will do it this week and also address the other comments.

Thanks for the update. Given what I could read in the qemu-devel thread, I think your approach of dumping the QEMU-generated DT and modifying it is sensible, however it should be done at runtime so that the base DT file does not need to be maintained as QEMU evolves. A similar suggestion was made in another PR by the way (https://github.com/OP-TEE/build/pull/695#issuecomment-1801336966).

stormset:
Addressed the mentioned issue, also rebased.

qemu-psa-sp.mk
```diff
@@ -0,0 +1,141 @@
...
+ifeq ($(TS_UEFI_TESTS),y)
+$(eval $(call build-ts-app,uefi-test))
+
+# uefi-test uses MM Communicate via the arm-ffa-user driver and the message
+# payload is forwarded in a carveout memory area. Adding reserved-memory node to
+# the device tree to prevent Linux from using the carveout area for other
+# purposes.
+
+CARVEOUT_ENTRY = $(ROOT)/build/qemu_v8/mm_communicate_carveout.dtsi
+$(QEMU_DTB_PATH): $(CARVEOUT_ENTRY) dumpdtb | linux
```
jforissier, line 130: I think `dumptdb` should be replaced by the actual file path (here and in place of the `dumptdb:` target in `qemu_v8.mk`), otherwise make will rebuild the same thing each time it is invoked although nothing changed. You may still keep a `dumptdb: path/to/dtb` target for convenience.

jforissier (commented):
Hi @stormset,

One minor comment, otherwise LGTM.

qemu-psa-sp.mk
stormset: The reason I want it to be always regenerated is that the arguments (`QEMU_ARGS`) passed to QEMU will have an effect on the generated device tree.

jforissier: Ah, makes sense. Ideally you could generate a file with the value of $(QEMU_ARGS) which would change only when those args change (via some clever make macro), then use that as the true dependency. But that's perhaps for another PR (see for instance check-conf-h and mv-if-changed in the OP-TEE OS makefiles).
Could you at least add a comment here? "# Note: the DTB generated by dumpdtb depends on $(QEMU_ARGS)" or something like that. Thanks!

stormset: Done. Also created a [commit](https://github.com/stormset/build/commit/58e71d3409835b45347ff18ed820752364260ed1) using the mentioned macros. Will create a PR after this gets merged.

jforissier: > Done. Also created a [commit](https://github.com/stormset/build/commit/58e71d3409835b45347ff18ed820752364260ed1) using the mentioned macros. Will create a PR after this gets merged.

Excellent!

qemu-psa-sp.mk
```diff
@@ -0,0 +1,141 @@
...
+	SPD=spmd \
+	SPMD_SPM_AT_SEL2=0 \
+	$(TF_A_FIP_SP_FLAGS) \
+	QEMU_USE_GIC_DRIVER=$(TFA_GIC_DRIVER) \
+	QEMU_TOS_FW_CONFIG_DTS=$(ROOT)/build/qemu_v8/spmc_el1_manifest.dts
+
+QEMU_EXTRA_ARGS ?= \
+	-cpu cortex-a57 \
+	-semihosting-config enable=on,target=native,userspace=on \
+	-d trace:'__*'
```
jforissier, line 60: You might not want this trace flag always? Not sure.

qemu-psa-sp.mk
```diff
@@ -0,0 +1,141 @@
...
+SP_PSA_CRYPTO_CONFIG		?= $(DEFAULT_SP_CONFIG)
+SP_PSA_ATTESTATION_CONFIG	?= $(DEFAULT_SP_CONFIG)
+SP_SMM_GATEWAY_CONFIG		?= $(DEFAULT_SP_CONFIG)
+
+# SPMC runs at EL1
+SPMC_AT_EL ?= 1
+# Use GICv2 or GICv3
+GICV3 ?= y
+# Use TF-A's Trusted Board Boot mechanism?
+TF_A_TRUSTED_BOARD_BOOT ?= y
```
etienne-lms, line 45: This switch seems unused. Is there even an altternate to TF-A?

qemu-psa-sp.mk
```diff
@@ -0,0 +1,141 @@
...
+ifeq ($(SP_PACKAGING_METHOD),fip)
+$(eval $(call add-dtc-define,SPMC_TESTS))
+$(eval $(call add-dtc-define,TS_SMM_GATEWAY))
+
+TF_A_EXPORTS += DTC_CPPFLAGS="$(DTC_CPPFLAGS)"
+endif
+
+OPTEE_OS_COMMON_EXTRA_FLAGS += \
+	CFG_SECURE_PARTITION=y \
+	CFG_CORE_SEL1_SPMC=y \
```
etienne-lms, line 80: depends on `SPMC_AT_EL`?

qemu_v8.mk
```diff
@@ -470,6 +471,24 @@ else
...
+	-bios bl1.bin		\
+	-initrd rootfs.cpio.gz \
+	-kernel Image \
+	-append 'console=ttyAMA0,38400 keep_bootcon root=/dev/vda2 $(QEMU_KERNEL_BOOTARGS)' \
+	$(QEMU_XEN) \
+	$(QEMU_EXTRA_ARGS)
+	
+ifeq ($(QEMU_USE_CUSTOM_DTB),y)
+QEMU_DTB_ARG	?= -dtb $(QEMU_DTB_PATH)
+endif
```
etienne-lms, line 490: can replace this change with:
```makefile
ifeq ($(QEMU_USE_CUSTOM_DTB),y)
QEMU_EXTRA_ARGS += -dtb $(QEMU_DTB_PATH)
endif

qemu-psa-sp.mk
etienne-lms: https://github.com/OP-TEE/build/pull/695 also generates a DTB for Qemu. Maybe there are some factorization to prepare.

etienne-lms (commented):
Commit message header line: 
`qemu-psa-sp: Armv8-A QEMU enablement for Trusted Services`

Could you add a short description of how to build for qemu_armv8a with trusted service, e.g. `Build with "make -f qemu-pas-sp.mk"`.

It would be nice to have also the list of the available config switches: SPMC_TESTS, MEASURED_BOOT, MEASURED_BOOT_FTPM, TS_SMM_GATEWAY, TS_UEFI_TESTS.

qemu-psa-sp.mk
stormset: It is used in `qemu_v8.mk`, I am ot sure what you are referring to.

qemu-psa-sp.mk
stormset: Removed these SPMC deployment dependent arguments as they are already handled in `qemu_v8.mk` depending on the value of `SPMC_AT_EL`.

qemu_v8.mk
stormset: When QEMU receives both `dtb` and `dumpdtb` arguments it wants to boot with that DTB, so it won't dump anything, but `QEMU_EXTRA_ARGS` needs to be passed to QEMU when dumping DTB as it has an effect on the DTB.

qemu-psa-sp.mk
etienne-lms: Right, sorry. discard my comment.

PR #689, opened 2023-10-05
qemu_v8: xen: re-use Dom0 root FS for DomU

jforissier:
Do not build a specific root FS for DomU. Instead, re-use the one that is built for Dom0. While this is certainly not a typical real-world configuration, it helps reduce both build time and disk space which is beneficial in many ways, especially for CI.


jenswi-linaro:

PR #692, opened 2023-10-11
make: change from localhost to 127.0.0.1

jbech-linaro:
Depending on how to you connect to internet, the tcp.localhost might not always work. You can eventually end up with the following error message:

(qemu) qemu-system-aarch64: -serial tcp:localhost:54320: Failed to connect to 'localhost:54320': Connecti on refused
qemu-system-aarch64: -serial tcp:localhost:54320: could not connect serial device to character backend 't cp:localhost:54320'

Therefore it's better to instead use 127.0.0.1 when connecting to the serial consoles.

FYI, this just happened to me when being at the osfc.io conf. I first couldn't figure out why the "make-run" didn't work. Then I tried disable my WiFi connection and things worked as expect. But to avoid this, as said in the commit message, I think it's better to use 127.0.0.1 instead.

### DIFF
--- a/qemu.mk
+++ b/qemu.mk
@@ -166,7 +166,7 @@ run-only:
 	$(call wait-for-ports,54320,54321)
 	cd $(BINARIES_PATH) && $(QEMU_BUILD)/arm-softmmu/qemu-system-arm \
 		-nographic \
-		-serial tcp:localhost:54320 -serial tcp:localhost:54321 \
+		-serial tcp:127.0.0.1:54320 -serial tcp:127.0.0.1:54321 \
 		-smp $(QEMU_SMP) \
 		-s -S -machine virt,secure=on -cpu cortex-a15 \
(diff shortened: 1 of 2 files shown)


jforissier:
Sounds reasonable, although I would think `localhost` should always resolve to `127.0.0.1`.

etienne-lms:
I have no strong opinion on which of 127.0.0.1 or localhost is better.
That said, `Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>` with _qemu.mk_ also updated.

jenswi-linaro:
Haha, reminds me of an issue I have at home with a Samsung TV trying to claim localhost as its hostname. Internet of stupid things, here we come! :-)

jbech-linaro:
Commit message updated and qemu.mk changed as well (Thanks Etienne!).

PR #697, opened 2023-11-07
qemu_v8: tfa: set ENABLE_FEAT_HCX=2

jforissier:
Build Trusted Firmware-A with ENABLE_FEAT_HCX=2. This is needed to use SPMC_AT_EL=2 with kernel v6.5 or later (any version that has commit "KVM: arm64: initialize HCRX_EL2" [1]). The symptom is a PANIC message in the platform secure console (out/bin/serial1.log):

 I/TC: Primary CPU switching to normal world boot
 PANIC at PC : 0x000000000e0a13c4

Link: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=af94aad4c9150cca6781ad134c950fb05dff43f9 [1]


jforissier:
Note that this PR should fix all the CI failures we are seeing in optee_os pull requests (the `make check (QEMUv8, Hafnium)` job).

etienne-lms:
`Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>` :)

PR #698, opened 2023-11-08
qemu_v8: tfa: set ENABLE_TRF_FOR_NS=0

jforissier:
TF-A commit "refactor(build): reorder platform Makefile evaluation" [1] has changed the default value of ENABLE_TRF_FOR_NS from 0 to 1. This is not correct, and does not work with QEMU v8.0.0. The error message is:

 $ tail -4 ../out/bin/serial0.log
 NOTICE:  BL1: Booting BL31
 NOTICE:  BL31: v2.9.0(release):v2.9.0-917-g3547270f5
 NOTICE:  BL31: Built : 10:53:55, Nov  8 2023
 write access to unsupported AArch64 system register op0:3 op1:4 crn:1 crm:2 op2:1

Quoting a proposed fix in TF-A [2]:

 FEAT_TRF is only mandatory on Armv8.4 if an ETM Architecture trace unit
 is implemented. Some implementations (eg QEMU) that do not implement an
 ETM trace unit do not support FEAT_TRF and therefore do not handle
 writes to TRFCR_EL2.

Make sure ENABLE_TRF_FOR_NS is set to 0 in qemu_v8.mk so that nothing breaks when we decide to upgrade TF-A from the current v2.9.0 to a more recent version that enables the flag by default.

Link: https://git.trustedfirmware.org/TF-A/trusted-firmware-a.git/commit/?id=3547270f59a3e27eab316a99cdaab6a835c10e85  [1]
Link: https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/23648 [2]


jforissier:
Created as a draft because this commit is not needed now (with TF-A v2.9.0), it is needed only for current TF-A master, but might not be needed if the next release contains the fix mentioned in the description.

etienne-lms (commented):
with a comment on commit message: TF-A version ID uses only 2 numbers.
=> s/v2.9.0/v2.9/

jenswi-linaro:

jforissier:
> TF-A version ID uses only 2 numbers. => s/v2.9.0/v2.9/

Both tags exist, actually.

etienne-lms:
> > TF-A version ID uses only 2 numbers. => s/v2.9.0/v2.9/
>
> Both tags exist, actually.

Indeed, since v2.7 there are version ID triplets.
Sorry for the noise ;)

PR #703, opened 2023-11-20
qemu: make local ports in use configurable

jbech-linaro:
Make it possible to set the local ports used for the non-secure (QEMU_NW_PORT) and secure (QEMU_SW_PORT) terminal when running QEMU.

When you need to run multiple QEMU/OP-TEE instances (debugging, multi-user etc.) on the same physical (non-containerized environment) computer, this is useful because you can avoid port collisions (port already in use message) by simply setting non-clashing port numbers when running make and thereby also avoid making changes directly to the makefile.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -105,6 +105,10 @@ QEMU_VIRTFS_MOUNTPOINT	?= /mnt/host
 
 # End of QEMU shared folder settings
 
+# The ports used for the consoles that are spawned when running QEMU.
+QEMU_NW_PORT ?= 54320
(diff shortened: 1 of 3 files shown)


jforissier (commented):
Good idea!

jenswi-linaro (commented):

etienne-lms (commented):

jbech-linaro:
Thanks! Squashed, rebased and tag(s) applied (should be) ready for merge!

PR #706, opened 2023-11-27
rpi3: dependency improvements

jbech-linaro:
From time to time it looks like the Raspberry Pi 3 build completes the build with no errors, but it's still an incomplete build, since the image for the SD card for example hasn't been created. This is most likely due to a dependency error in the Makefile. After reviewing the dependencies it looks like there were at least two dependencies that was wrong/missing.

- Change so the boot filesystem also depends on linux and the U-boot environment file.
- Change so tf-a depends on u-boot and not u-boot-env.
- Change arm-tf targets to tf-a.
- Add .PHONY to a few more targets.

Fixes: https://github.com/OP-TEE/optee_os/issues/6442

With these changes I've rebuilt 4 times from scratch and all of them successfully built everything.


jbech-linaro:
@jforissier , thanks for the feedback, I've split up my patch into four different commits and force pushed it to the same branch. Hopefully the commits are more self contained now, thanks!

jforissier:
`Reviewed-by: Jerome Forissier <jerome.forissier@linaro.org>` for all commits, thanks!

jbech-linaro:

PR #711, opened 2023-12-12
stm32mp1: *_SCMI flavor for STM32MP15 ST boards

etienne-lms:
Define `157x_xxx_SCMI` platform flavors for STM32MP15 ST boards where RCC secure hardening is enabled and non-secure needs to use SCMI service. These are board DTS file ending with _-scmi.dts_.

The default stm32mp1 platform flavor is now `157C_DK2_SCMI`.

This change depends on https://github.com/OP-TEE/optee_os/pull/6541.


GseoC:
Hi,
Ok for the most part but there's a mix between '-' and '_' in the file names. Can you correct this?
e.g: DK1_SCMI

etienne-lms:
I have simplified the picture here. The 2 first commits remove the various genimage config files and bootfs file tree overlays and replaces them by files generated at build time (modif dine in boards _post-image.sh_ shell script).

The 3rd patch defined `157x_XXX_SCMI` platform flavors for the config related to _stm32mp15*-scmi.dts_ board files.

etienne-lms:
@GseoC, could you look back to these changes?

GseoC (commented):
For what it's worth
`Acked-by: Gatien Chevallier <gatien.chevallier@foss.st.com>`

with or without comment addressed

etienne-lms:
I've addressed comment (s/TFA_VERSION/TFA_BOARD_REV/g) and applied the review tag.
Thanks.

etienne-lms:
@jforissier, ok to merge these before next release tag? This change syncs _build.git/stm32mp1.mk_ with _optee_os_ regarding platform flavor names.

etienne-lms:
Note: OP-TEE OS needs a fixup, see https://github.com/OP-TEE/optee_os/pull/6620.

jforissier:
> @jforissier, ok to merge these before next release tag? This change syncs _build.git/stm32mp1.mk_ with _optee_os_ regarding platform flavor names.

It's a bit more than a bug fix so it should normally not go after -rc1 with the risk it could break things for stm32... But if you say you're OK with that, then fine.

PR #712, opened 2023-12-14
qemu: Correct initrd-end in u-boot config fragment

PeikanTsai:
smhload set ramdisk_addr_end to the end address of rootfs, but the kernel expect linux,initrd-end
to be the first address after the initrd.
Update the default bootcmd to meet the requirement.


jenswi-linaro (commented):

PR #716, opened 2024-01-03
qemu-check.exp: protect access to $::env(XEN_FFA)

jforissier:
If the XEN_FFA environment variable is not set, $::env(XEN_FFA) raises an exception. It is the case when using qemu.mk (while qemu_v8.mk correctly sets XEN_FFA always before invoking qem-check.exp):

 $ make check
 [...]
 no such variable
     (read trace on "::env(XEN_FFA)")
     invoked from within
 "if {$::env(XEN_FFA) == "y"} {
         set cmd2 "xl create guest_ffa.cfg"
 } else {
         set cmd2 "xl create guest.cfg"
 }"
     (file "/home/jerome/work/optee_repo_qemu/build/../build/qemu-check.exp" line 15)
 make: *** [Makefile:194: check] Error 1

Rather than introducing XEN_FFA in qemu.mk where is doesn't apply, guard the $::env(XEN_FFA) statement with an 'if {[info exists ::env(XEN_FFA)] ...' in qemu-check.exp.

Fixes: 4809b4f67fa8 ("qemu_v8: support for FF-A in Xen")


jenswi-linaro:

PR #718, opened 2024-01-10
Integrate the FWU SP into the OP-TEE build flow for the FVP platform.

gabor-toth-arm:
- Integrate the FWU SP into the OP-TEE build flow for the FVP platform.


fvp-psa-sp.mk
```diff
@@ -128,4 +135,29 @@ carveout-dtb-clean:
...
+
+# Copy the disk image used by FWU to the build directory to allow the FVP binary to find it.
+$(ROOT)/build/secure-flash.img:
+	cp $(ROOT)/trusted-services/components/media/disk/disk_images/multi_location_fw.img $(ROOT)/build/secure-flash.img
+
+# Add a shortcut to help manually doing the copy.
+ffa-fwu-fash-img: $(ROOT)/build/secure-flash.img
+
+# TODO: find the correct place for this dependency.
+ffa-fwu-sp: $(ROOT)/build/secure-flash.img
```
jenswi-linaro, line 154: `$(ROOT)/out/bin/secure-flash.img` perhaps?

fvp-psa-sp.mk
```diff
@@ -128,4 +135,29 @@ carveout-dtb-clean:
...
+
+endif
+
+ffa-fwu-fash-img-clean:
+	rm -rf $(ROOT)/build/secure-flash.img
```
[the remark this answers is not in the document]
gabor-toth-arm: Corrected

trusted-services.mk
```diff
@@ -9,6 +9,7 @@ TS_INSTALL_PREFIX	?= $(OUT_PATH)/ts-install
 # Secure Partitions
 ################################################################################
 .PHONY: ffa-sp-all ffa-sp-all-clean ffa-sp-all-realclean
+.NOTPARALLEL: ffa-sp-all
```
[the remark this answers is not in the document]
gabor-toth-arm: We are not sure why it was added in the original code. I removed it and faced no problems.

fvp-psa-sp.mk
gabor-toth-arm: Updated to use the out directory as you proposed.

trusted-services.mk
```diff
@@ -189,3 +189,49 @@ ffa-test-all: ffa-$1
...
+		$(TS_HOST_${FFA_$1_UC_NAME}_EXTRA_FLAGS)
+	$$(MAKE) -C $(TS_BUILD_PATH)/$1 install
+
+.PHONY: ts-host-$1-clean
+ts-host-$1-clean:
+	$$(MAKE) -C $(TS_BUILD_PATH)/$1 clean
+
+.PHONY: ts-host-$1-realclean
+ts-host-$1-realclean:
+	rm -rf $(TS_BUILD_PATH)/$1
```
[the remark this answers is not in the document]
gabor-toth-arm: Added check in build-ts-host-app to make sure the argument will never be empty.

trusted-services.mk
```diff
@@ -189,3 +189,49 @@ ffa-test-all: ffa-$1
...
 endef
+
+################################################################################
+# Trusted Services hot applications
+################################################################################
+.PHONY: ts-host-all ts-host-all-clean ts-host-all-realclean
+all: ts-host-all
+
+ts-host-all-realclean:
+	rm -rf $(TS_INSTALL_PREFIX)/linux-pc
```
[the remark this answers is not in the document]
gabor-toth-arm: "linux-pc" is hardcoded here, so this rm -rf cannot make too big mistake even with incorrect $(TS_INSTALL_PREFIX). Maybe this deletion could be moved somewhere else deeper in the build system, but that would make it more complex and more error prone. I don't really see a better option for the removal of this folder, do you have any other solution in mind?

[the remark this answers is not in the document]
jenswi-linaro: No, I'm just trying to minimize the number of `rm -rf`'s.

fvp-psa-sp.mk
```diff
@@ -95,6 +101,7 @@ $(eval $(call build-sp,spm-test3,opteesp,23eb0100-e32a-4497-9052-2f11e584afa6,$(
 $(eval $(call build-sp,spm-test4,opteesp,423762ed-7772-406f-99d8-0c27da0abbf8,$(SP_SPMC_TEST_EXTRA_FLAGS)))
 endif
 
+
```
jenswi-linaro, line 104: Leftover?

.gitignore
```diff
@@ -6,3 +6,4 @@
 gdb
 pi3.cfg
 .hafnium_checkout
+secure-flash.img
```
jenswi-linaro, line 9: Not needed any longer.

fvp-psa-sp.mk
```diff
@@ -128,4 +135,29 @@ carveout-dtb-clean:
...
+ffa-fwu-sp: ts-host-fwu-tool
+
+# Copy the disk image used by FWU to the build directory to allow the FVP binary to find it.
+$(ROOT)/out/bin/secure-flash.img:
+	mkdir -p $(ROOT)/out/bin && cp $(ROOT)/trusted-services/components/media/disk/disk_images/multi_location_fw.img $(ROOT)/out/bin/secure-flash.img
+
+# Add a shortcut to help manually doing the copy.
+ffa-fwu-fash-img: $(ROOT)/out/bin/secure-flash.img
+
+# TODO: find the correct place for this dependency.
```
jenswi-linaro, line 153: Is the comment still relevant?

fvp-psa-sp.mk
```diff
@@ -128,4 +135,29 @@ carveout-dtb-clean:
...
+ifeq ($(TS_FW_UPDATE),y)
+
+# TODO: the fwu-tool is currently not needed.
+$(eval $(call  build-ts-host-app,fwu-tool))
+
+ffa-fwu-sp: ts-host-fwu-tool
+
+# Copy the disk image used by FWU to the build directory to allow the FVP binary to find it.
+$(ROOT)/out/bin/secure-flash.img:
+	mkdir -p $(ROOT)/out/bin && cp $(ROOT)/trusted-services/components/media/disk/disk_images/multi_location_fw.img $(ROOT)/out/bin/secure-flash.img
```
jenswi-linaro, line 148: You can do the `cp` on the next line without any `&&`.

jenswi-linaro (commented):
In the commit "fvp: deploy the FWU SP" there's a "[1]" reference, but it's never defined.

trusted-services.mk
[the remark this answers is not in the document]
gabor-toth-arm: I discussed this with the the original implementer, but we found no better solution, so I left it like this.

fvp-psa-sp.mk
gabor-toth-arm: Yes. Corrected

.gitignore
gabor-toth-arm: I reverted the change where I moved the image under out and wrote an answer there, but for easier access I will also copy it here:

"Unfortunately the image will not be accessible by the tests if the file is put under out (checkImgDirAccess fails).
I think the problem is that the FVP runs from the out folder and maybe it makes sure that files are not modified there."

fvp-psa-sp.mk
gabor-toth-arm: Removed

fvp-psa-sp.mk
jenswi-linaro: `$(ROOT)/build` is not a good directory, it's better to put it in `$(ROOT)` then. By the way, FVP is executed from the `$(ROOT)/build`, I think it would make sense to cd to `$(ROOT)/out/bin` or `$(ROOT)/out just as we do with QEMU.

gabor-toth-arm: Thank you for the advice. Corrected.

fvp.mk
```diff
@@ -337,4 +338,5 @@ FVP_ARGS ?= \
 endif
 
 run-only:
-	$(FVP_PATH)/$(FVP_BIN) $(FVP_ARGS) $(FVP_EXTRA_ARGS)
+	make -p $(BINARIES_PATH)
```
jenswi-linaro, line 341: The `run-only` target should only run, not have any other side effect than what possibly the emulation itself.

.gitignore
jenswi-linaro: Not needed any longer. :-)

fvp.mk
gabor-toth-arm: Updated it. I also changed the solution a bit to use "semihosting-cwd" instead of CD-ing into the folder. This seems to be a nicer way for our change.

jenswi-linaro:

gabor-toth-arm:
Could we please merge this?

gabor-toth-arm:
Thank you :)

PR #719, opened 2024-01-16
rpi3: Increase size for root files system

jockebech:
As part of updating the manifest to using a more up-to-date Linux kernel, we need to increase the size of the root file system a bit.

With these settings, there is roughly 50MB free on the filesystem before and after running xtest.

This doesn't really affect what I tested [here](https://github.com/OP-TEE/optee_os/pull/6574#issuecomment-1891769754) for the 4.1.0 release, so I see no harm in merging this before the release. On the other hand no hurry either, so this can wait to after the release as well if that is preferable.


jforissier (commented):

jenswi-linaro:

jbech-linaro:

PR #721, opened 2024-01-19
stm32mp1: factorize RPMB test S30optee scritpt

etienne-lms:
_S30optee_ init script does not depend on target platform hence use a generic directory names (_./overlay-rootfs-rpmb/_) to store this file instead of a platform depend directory name (_./overlay(STM32MP157C_EV1-rpmb/_).


igoropaniuk:
`Reviewed-by: Igor Opaniuk <igor.opaniuk@foundries.io>`

PR #722, opened 2024-01-19
qemuv8: rust.exp: set timeout to 10 seconds

jforissier:
There is no timeout specified in rust.exp. This makes the tests unreliable, because depending on which tests have been run before, the $timeout global variable may have different values. For example, it is set to 900 by qemu-check.exp (the main file), but may be overriden by trusted-keys.exp to 5 seconds. That's not enough in some cases, causing the following error in optee_os CI:

 2024-01-19T11:58:23.5778720Z Running signature_verification-rs...
 2024-01-19T11:58:28.3431817Z !!! Timeout: Test failed

### DIFF
--- a/rust.exp
+++ b/rust.exp
@@ -4,6 +4,8 @@
 # success, >0 for error.
 #
 
+set timeout 10
+
 info "Test Rust example applications:\n"
 info "Running acipher-rs...\n"
 send -- "acipher-rs 256 teststring\r"


jforissier:
Could someone please Ack? I'd like to fix the CI before the release. Thanks!

etienne-lms (commented):

jforissier:
Thanks!

PR #724, opened 2024-01-22
buildroot: add $HOME/.cargo/bin to $PATH prior to building

jforissier:
Since commit e742dc88f51c ("qemu_v8: Enable Rust examples build by default"), the 'cargo' command installed by the optee-rust target needs to be in the $PATH or the Rust examples will fail to build.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -355,7 +355,7 @@ buildroot: optee-os optee-rust
 		$(DEFCONFIG_FTPM) \
 		--br-defconfig out-br/extra.conf \
 		--make-cmd $(MAKE))
-	@export PATH=$${HOME}/.cargo/bin:$${PATH} && $(MAKE) $(br-make-flags) -C ../out-br all
+	@export PATH=$(HOME)/.cargo/bin:$(PATH) && $(MAKE) $(br-make-flags) -C ../out-br all
 
 .PHONY: buildroot-clean
 buildroot-clean:
--- a/common.mk
+++ b/common.mk
@@ -355,7 +355,7 @@ buildroot: optee-os optee-rust
 		$(DEFCONFIG_FTPM) \
 		--br-defconfig out-br/extra.conf \
 		--make-cmd $(MAKE))
-	@$(MAKE) $(br-make-flags) -C ../out-br all
+	@export PATH=$${HOME}/.cargo/bin:$${PATH} && $(MAKE) $(br-make-flags) -C ../out-br all
 
 .PHONY: buildroot-clean
 buildroot-clean:
--- a/common.mk
+++ b/common.mk
@@ -355,7 +355,7 @@ buildroot: optee-os optee-rust
 		$(DEFCONFIG_FTPM) \
 		--br-defconfig out-br/extra.conf \
 		--make-cmd $(MAKE))
-	@export PATH=$(HOME)/.cargo/bin:$(PATH) && $(MAKE) $(br-make-flags) -C ../out-br all
+	@source $(HOME)/.cargo/env && $(MAKE) $(br-make-flags) -C ../out-br all
 
 .PHONY: buildroot-clean
 buildroot-clean:


common.mk
```diff
@@ -355,7 +355,7 @@ buildroot: optee-os optee-rust
 		$(DEFCONFIG_FTPM) \
 		--br-defconfig out-br/extra.conf \
 		--make-cmd $(MAKE))
-	@$(MAKE) $(br-make-flags) -C ../out-br all
+	@export PATH=$${HOME}/.cargo/bin:$${PATH} && $(MAKE) $(br-make-flags) -C ../out-br all
```
jenswi-linaro, line 358: Why is this somewhat more complicated way of writing `@export PATH=$(HOME)/.cargo/bin:$(PATH) ...`?

By the way, can we rely on `$(HOME)/.cargo/bin` to contain the right thing? What if the user for some reason messes around with the user-specific cargo installation? This toolchain isn't special in any way compared to the GCC toolchain we're using and we're not expecting the user to keep a sane GCC cross-toolchain ready for us.

jforissier: > By the way, can we rely on `$(HOME)/.cargo/bin` to contain the right thing?

Yes, because that path is populated by the `optee-rust` target which is a dependency of this one.

jforissier: Do you mean that `optee-rust` should use a more specific path to install the toolchain?

jenswi-linaro: To keep stuff in `~/.cargo/bin/`, that is.

jenswi-linaro: > Do you mean that `optee-rust` should use a more specific path to install the toolchain?

I'm sorry, I missed your comment. Yes, I believe a more safe choice would be to install the rust toolchain under the `$(ROOT)/toolchains` directory.

jforissier: OK. Then I think this would require some changes in https://github.com/apache/incubator-teaclave-trustzone-sdk/blob/no-std/setup.sh in order to support choosing the installation directory. @b49020

b49020: I have been thinking about moving Rust toolchain installation from `setup.sh` to rather `toolchain.mk` which aligns to the comments above. So I will work on a patch to remove `setup.sh` invocation and rather add Rust toolchain installation instructions to `toolchain.mk` instead.

b49020: Also, as per https://stackoverflow.com/questions/57765424/overriding-rust-installation-default-paths-home-cargo-and-home-rustup, we can customize Rust toolchain installation for OP-TEE buildroot setup.

common.mk
```diff
@@ -355,7 +355,7 @@ buildroot: optee-os optee-rust
 		$(DEFCONFIG_FTPM) \
 		--br-defconfig out-br/extra.conf \
 		--make-cmd $(MAKE))
-	@$(MAKE) $(br-make-flags) -C ../out-br all
+	@export PATH=$(HOME)/.cargo/bin:$(PATH) && $(MAKE) $(br-make-flags) -C ../out-br all
```
jenswi-linaro, line 358: Even though `export` seems to work I wonder if `env` wouldn't be preferable.

jforissier: `env` applies the environment to the specified command only, while `export` modifies the environment that the shell passes to any subsequent sub-process. That's more or less the same here because the shell will exit as soon as the `$(MAKE) ...` command exits (make spawns a new shell for each line in a recipe).

b49020: We should rather do `source "$HOME/.cargo/env"` here to setup path for Rust toolchain. It worked for me since I have that added as part of my `~/.bashrc`.

jenswi-linaro: > We should rather do `source "$HOME/.cargo/env"` here to setup path for Rust toolchain. It worked for me since I have that added as part of my `~/.bashrc`.

Good point.

jforissier:
Updated

b49020:
@jforissier @jenswi-linaro 

> > Do you mean that `optee-rust` should use a more specific path to install the toolchain?
> 
> I'm sorry, I missed your comment. Yes, I believe a more safe choice would be to install the rust toolchain under the `$(ROOT)/toolchains` directory.

https://github.com/OP-TEE/build/pull/725 implements this. Changes in this PR looks good to me as well, if you would like to merge it first then feel free to apply:

`Reviewed-by: Sumit Garg <sumit.garg@linaro.org>`

jforissier:
@b49020 thanks, but let's drop this in favor of #725.

PR #725, opened 2024-01-24
toolchain.mk: Add support for Rust toolchain for aarch{32/64}

b49020:
Rather than invoking custom setup.sh script provided by OP-TEE Rust SDK, add Rust toolchain installation support as part of toolchain Makefile. This allows to separate OP-TEE specific Rust toolchain installation from default Rust toolchain installation. Now all OP-TEE specific Rust toolchains will be installed under: $(ROOT)/toolchains/rust/.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -331,7 +331,7 @@ br-make-flags := -j1
 endif
 
 .PHONY: buildroot
-buildroot: optee-os optee-rust
+buildroot: optee-os
 	@mkdir -p ../out-br
 	@rm -f ../out-br/build/optee_*/.stamp_*
 	@rm -f ../out-br/extra.conf
@@ -349,7 +349,7 @@ buildroot: optee-os optee-rust
 		$(DEFCONFIG_FTPM) \
 		--br-defconfig out-br/extra.conf \
 		--make-cmd $(MAKE))
-	@$(MAKE) $(br-make-flags) -C ../out-br all
+	@source $(RUST_TOOLCHAIN_PATH)/.cargo/env && $(MAKE) $(br-make-flags) -C ../out-br all
 
 .PHONY: buildroot-clean
 buildroot-clean:
@@ -550,22 +550,6 @@ optee-os-common:
 .PHONY: optee-os-clean-common
 optee-os-clean-common:
 	$(MAKE) -C $(OPTEE_OS_PATH) $(OPTEE_OS_COMMON_FLAGS) clean
-
-################################################################################
-# OP-TEE Rust
-################################################################################
-.PHONY: optee-rust
-optee-rust: $(OPTEE_RUST_PATH)/.done
-
-$(OPTEE_RUST_PATH)/.done:
-ifeq ($(RUST_ENABLE),y)
-	@(export OPTEE_DIR=$(ROOT) && \
-	  export CARGO_NET_GIT_FETCH_WITH_CLI=true && \
-	  cd $(OPTEE_RUST_PATH) && ./setup.sh && touch .done)
-endif
-
-optee-rust-clean:
-	rm -f $(OPTEE_RUST_PATH)/.done
 
 ################################################################################
 # fTPM Rules
--- a/toolchain.mk
+++ b/toolchain.mk
@@ -57,14 +57,14 @@ AARCH64_GCC_VERSION 		?= arm-gnu-toolchain-11.3.rel1-x86_64-aarch64-none-linux-g
 SRC_AARCH64_GCC 		?= https://developer.arm.com/-/media/Files/downloads/gnu/11.3.rel1/binrel/$(AARCH64_GCC_VERSION).tar.xz
 
 .PHONY: toolchains
-toolchains: aarch32 aarch64
+toolchains: aarch32-toolchain aarch64-toolchain
 
-.PHONY: aarch32
-aarch32:
+.PHONY: aarch32-toolchain
+aarch32-toolchain:
 	$(call dltc,$(AARCH32_PATH),$(SRC_AARCH32_GCC),$(AARCH32_GCC_VERSION))
 
-.PHONY: aarch64
-aarch64:
+.PHONY: aarch64-toolchain
+aarch64-toolchain:
 	$(call dltc,$(AARCH64_PATH),$(SRC_AARCH64_GCC),$(AARCH64_GCC_VERSION))
 
 CLANG_VER			?= 12.0.0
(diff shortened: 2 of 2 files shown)


b49020:
@DemesneGH fyi..

common.mk
```diff
@@ -349,6 +349,7 @@ buildroot: optee-os optee-rust
 		$(DEFCONFIG_FTPM) \
 		--br-defconfig out-br/extra.conf \
 		--make-cmd $(MAKE))
+	@source $(RUST_TOOLCHAIN_PATH)/.cargo/env
```
jenswi-linaro, line 352: Wouldn't this be lost when this line is done?

jforissier:
@jbech-linaro I believe IBART needs to have "make toolchains" run again.

jforissier (commented):
LGTM with minor comments.

common.mk
jforissier: Good catch :) this should be on the same line as `$(MAKE) $(br-make-flags) ...`

b49020: That's true, I will fix it.

toolchain.mk
```diff
@@ -56,8 +56,10 @@ AARCH64_CROSS_COMPILE 		?= $(AARCH64_PATH)/bin/aarch64-linux-gnu-
 AARCH64_GCC_VERSION 		?= arm-gnu-toolchain-11.3.rel1-x86_64-aarch64-none-linux-gnu
 SRC_AARCH64_GCC 		?= https://developer.arm.com/-/media/Files/downloads/gnu/11.3.rel1/binrel/$(AARCH64_GCC_VERSION).tar.xz
 
+RUST_TOOLCHAIN_PATH 		?= $(TOOLCHAIN_ROOT)/rust
+
 .PHONY: toolchains
-toolchains: aarch32 aarch64
+toolchains: aarch32 aarch64 rust
```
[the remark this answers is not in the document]
b49020: Done

toolchain.mk
```diff
@@ -67,6 +69,25 @@ aarch32:
 aarch64:
 	$(call dltc,$(AARCH64_PATH),$(SRC_AARCH64_GCC),$(AARCH64_GCC_VERSION))
 
+# Download the Rust toolchain
+define dl-rust
```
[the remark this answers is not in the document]
b49020: Done

b49020:
> @jbech-linaro I believe IBART needs to have "make toolchains" run again.

Yeah that will fix the build error. BTW, shouldn't we run `make toolchains` by default in the CI? It will be a no-op if the toolchains are already present.

b49020:
Addressed comments and tag applied.

toolchain.mk
```diff
@@ -56,17 +56,37 @@ AARCH64_CROSS_COMPILE 		?= $(AARCH64_PATH)/bin/aarch64-linux-gnu-
...
-.PHONY: aarch64
-aarch64:
+.PHONY: aarch64-toolchain
+aarch64-toolchain:
 	$(call dltc,$(AARCH64_PATH),$(SRC_AARCH64_GCC),$(AARCH64_GCC_VERSION))
 
+# Download the Rust toolchain
+define dl-rust-toolchain
+	@if [ ! -d "$(1)" ]; then \
+		mkdir -p $(1); \
```
jenswi-linaro, line 75: Shouldn't the commands be separated with `&&` instead to propagate eventual errors?

b49020: Done and along with that added some error handling too.

jbech-linaro:
> > @jbech-linaro I believe IBART needs to have "make toolchains" run again.
> 

Will fix!

> Yeah that will fix the build error. BTW, shouldn't we run `make toolchains` by default in the CI? It will be a no-op if the toolchains are already present.

In theory yes, in practice no, since it's a slow operation and it rarely changes. I.e., I'd rather update that manually once a year or so, than having a few extra minutes added to every pull request.

jforissier:
For "toolchain.mk: Add *-toolchain suffix to targets":

b49020:
Tag applied, thanks.

jforissier:
> > > @jbech-linaro I believe IBART needs to have "make toolchains" run again.
> 
> Will fix!
> 
> > Yeah that will fix the build error. BTW, shouldn't we run `make toolchains` by default in the CI? It will be a no-op if the toolchains are already present.
> 
> In theory yes, in practice no, since it's a slow operation and it rarely changes. I.e., I'd rather update that manually once a year or so, than having a few extra minutes added to every pull request.

If you run `make toolchains` from the same directory every time, then it would not take any time at all because the files exist already. For example on my laptop:
```
build$ time make -j3 toolchains
[...]
real    1m19.484s
user    0m31.974s
sys     0m9.434s
build$ time make -j3 toolchains

real    0m0.056s
user    0m0.036s
sys     0m0.008s
```

toolchain.mk
```diff
@@ -56,17 +56,38 @@ AARCH64_CROSS_COMPILE 		?= $(AARCH64_PATH)/bin/aarch64-linux-gnu-
...
+	@if [ ! -d "$(1)" ]; then \
+		( mkdir -p $(1) && \
+		export RUSTUP_HOME=$(1)/.rustup && \
+		export CARGO_HOME=$(1)/.cargo && \
+		curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --no-modify-path && \
+		source "$(1)/.cargo/env" && \
+		rustup target install aarch64-unknown-linux-gnu && \
+		rustup target install arm-unknown-linux-gnueabihf && \
+		rustup default nightly-2023-12-18 ) || \
+		( echo Rust toolchain installation failed && rm -rf $(1) ); \
```
jenswi-linaro, line 83: The infamous `rm -rf`, what if `RUST_TOOLCHAIN_PATH` has been set to something bad, like `/` or `$(HOME)`?

jforissier: ...then the user is left homeless! :rofl: 
I agree. In fact the whole `||` part is not needed IMO. If something fails here, better let the user clean up :wink:

b49020: Yeah but I was following the existing practice: https://github.com/OP-TEE/build/blob/master/toolchain.mk#L22. However, I can `||` part.

jbech-linaro:
> If you run `make toolchains` from the same directory every time,

I always start from a completely clean folder and symlink to my toolchain folder, but I could of course start by adding the symlink and adding then run `make toolchains` to the flow. After seeing what's happening when running make toolchains with the rust patches, maybe I have to revise my comment above. It's seems like there was a lot going on when it was cloning the rust stuff. I.e., it's not just a compressed file as the other toolchains, so perhaps different packages for rust gets updated more frequently. I'll make some test and changes. In any case, the rust toolchain is in place now and I've restarted the build for this PR.

b49020:
Update

etienne-lms (commented):

toolchain.mk
jenswi-linaro: > Yeah but I was following the existing practice: https://github.com/OP-TEE/build/blob/master/toolchain.mk#L22. However, I can drop `||` part.

Ouch, people for some reason seem to love `rm -rf`.

DemesneGH:
`Acked-by: Yuan Zhuang <yuanz@apache.org>`

b49020:
Tags applied, thanks.

PR #729, opened 2024-01-30
common.mk: add linux-menuconfig-common target

jenswi-linaro:
Adds a linux-menuconfig-common target to run "make menuconfig" with environment and flags matching the configured linux kernel.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -364,6 +364,10 @@ buildroot-cleaner:
 ################################################################################
 LINUX_COMMON_FLAGS ?= LOCALVERSION= CROSS_COMPILE=$(CROSS_COMPILE_NS_KERNEL)
 
+.PHONY: linux-menuconfig-common
+linux-menuconfig-common: linux-defconfig
+	$(MAKE) -C $(LINUX_PATH) $(LINUX_COMMON_FLAGS) menuconfig
+
 .PHONY: linux-common
 linux-common: linux-defconfig
 	$(MAKE) -C $(LINUX_PATH) $(LINUX_COMMON_FLAGS) $(LINUX_COMMON_TARGETS)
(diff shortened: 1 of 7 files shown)


common.mk
```diff
@@ -364,6 +364,12 @@ buildroot-cleaner:
 ################################################################################
 LINUX_COMMON_FLAGS ?= LOCALVERSION= CROSS_COMPILE=$(CROSS_COMPILE_NS_KERNEL)
 
+.PHONY: linux-menuconfig-common
+linux-menuconfig-common: linux-defconfig
+	$(MAKE) -C $(LINUX_PATH) \
+		$(filter-out Image script_gdb uImage zImage,\
```
jbech-linaro, line 370: Why do we have to filter-out these? Wouldn't just calling menuconfig work? In any case.

jenswi-linaro: The kernel would be built first before running "make menuconfig" so it's a bit inefficient, especially since we're likely to rebuild right after.

jenswi-linaro:
Tag applied

common.mk
jforissier: TBH none of `Image script_gdb uImage zImage` should be in `$(LINUX_COMMON_FLAGS)` because they are targets, not flags. Some small refactoring would be in order here I believe :wink:

jenswi-linaro: Makes sense, I'll update this PR.

jenswi-linaro:
Updated with "Introduce LINUX_COMMON_TARGETS"

etienne-lms (commented):
For commit "common.mk: add linux-menuconfig-common target", in commit message:
```patch
-Adds a linux-menuconfig-common target to run "make menuconfig" with
+Add a linux-menuconfig-common target to run "make menuconfig" with
```
`Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>` for the 2 commits.

jforissier (approved):

jenswi-linaro:
Comment addressed, and tags applied.

PR #730, opened 2024-01-30
common: target optee-os-devkit builds the devkit

etienne-lms:
Adds build target `optee-os-devkit` to get the TA devkit be built and installed. It can be later used as a dependency when building external TAs that need to build linked to OP-TEE as early TAs where building OP-TEE core depends on such TAs being already built. This applies to ftpm hence replace the dependency of `ftpm` make target with a dependency on `optee-os-devkit`.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -551,6 +551,10 @@ optee-os-common:
 optee-os-clean-common:
 	$(MAKE) -C $(OPTEE_OS_PATH) $(OPTEE_OS_COMMON_FLAGS) clean
 
+.PHONY: optee-os-devkit
+optee-os-devkit:
+	$(MAKE) -C $(OPTEE_OS_PATH) $(OPTEE_OS_COMMON_FLAGS) ta_dev_kit
+
 ################################################################################
 # fTPM Rules
 ################################################################################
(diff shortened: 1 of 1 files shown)


common.mk
```diff
@@ -568,7 +572,7 @@ FTPM_FLAGS ?= 						\
 .PHONY: ftpm
 ftpm:
 ifeq ($(MEASURED_BOOT_FTPM),y)
-ftpm: optee-os
+ftpm: optee-os-devkit
```
jforissier, line 575: This dependency looks good, but I don't see where/how the ftpm TA is linked into OP-TEE as an early TA. Perhaps this extra information should be omitted from the commit description.

jforissier (commented):
One comment, then:

jenswi-linaro:
With Jerome's comment addressed:

common.mk
etienne-lms: I was considering one building with `EARLY_TA_PATHS=/path/to/ftpm/build/bc50d971-d4c9-42c4-82cb-343fb7f37896.stripped.elf`.

Nevertheless, I'll remove the related information in the commit message.

etienne-lms:
I've removed the pointed sentence from the commit message and slightly rephrased it.
Review tags applied.

PR #732, opened 2024-02-06
Buildroot: enable Fortran and OpenMP in external toolchains

jforissier:
In preparation for upgrading Buildroot to v2023.11.1, set BR2_EXTERNAL_TOOLCHAIN_FORTRAN=y and
BR2_EXTERNAL_TOOLCHAIN_OPENMP=y, not because Fortran or OpenMP are needed but to silence some errors that would come up with the newer version:

 Fortran support is not selected but is available in external toolchain
 OpenMP support is not selected but is available in external toolchain


jenswi-linaro:

PR #735, opened 2024-02-19
rockpi4: increase root FS size from 384 to 640M when LINUX_MODULES=y

jforissier:
make LINUX_MODULES=y fails due to insufficient space for the root FS. Therefore increase the default size from 384 to 640M. With that the available space is 35.5M:
```
 # df -h /
 Filesystem                Size      Used Available Use% Mounted on
 /dev/root               629.5M    562.0M     35.5M  94% /
```


jbech-linaro (commented):

Issue #744, opened 2024-04-16
optee_test_ext: /bin/bash: Argument list too long


imre-kis-arm:
The hooks in [optee_test_ext.mk](https://github.com/OP-TEE/build/blob/6b9359d7933485fd8c01b391f6f04c8582d911f9/br-ext/package/optee_test_ext/optee_test_ext.mk) chains each TA's build step into a single command (note `&&` at the end of the foreach body). `$(TARGET_CONFIGURE_OPTS)` contains the whole environment, including `PATH` and the absolute path of each compiler tool (CC, LD, AS, OBJCOPY, etc.). This causes the hook to generate a command which is around 100k long (checked by removing `@` before the foreach). Depending on the length of the build directory path and the system's limitation on the maximal command length this can cause a build failure.

```
[100%] Built target xtest
/opt/cmake-3.18.4-Linux-x86_64/bin/cmake -E cmake_progress_start /[...]/optee-spmc-test/out-br/build/optee_test_ext-1.0/CMakeFiles 0
make[2]: /bin/bash: Argument list too long
make[2]: *** [package/pkg-generic.mk:284: /[...]/optee-spmc-test/out-br/build/optee_test_ext-1.0/.stamp_built] Error 127
make[1]: *** [Makefile:23: _all] Error 2
make[1]: Leaving directory '/[...]/optee-spmc-test/out-br'
make: *** [common.mk:341: buildroot] Error 2
make: Leaving directory '/[...]/optee-spmc-test/build'
```

It would be better to run each TA build as a separate command or have separate hooks for each TA if that's possible.

Although I didn't have problems with [optee_examples_ext](https://github.com/OP-TEE/build/blob/6b9359d7933485fd8c01b391f6f04c8582d911f9/br-ext/package/optee_examples_ext/optee_examples_ext.mk), it uses the same pattern, so it might worth changing that too.

Tested with 4.2.0.

jforissier:
Hi @imre-kis-arm,

The build process for TAs in `optee_test_ext.mk` is quite convoluted, in order to address the issue I decided to delegate the build to the makefile in `optee_test/ta`. Please see https://github.com/OP-TEE/optee_test/pull/740 which is a prerequisite then https://github.com/OP-TEE/build/pull/745.

imre-kis-arm:
Hi @jforissier,

[...]

PR #745, opened 2024-04-17
optee_test_ext.mk: use the optee_test Makefile to build TAs

jforissier:
Rather than enumerating all the TAs and invoking each Makefile individually (optee_test/ta/<ta_name>/Makefile), user the higher level Makefile (optee_test/ta/Makefile.gmake). This makes things simpler here and less fragile if something changes in optee_test. In addition, this should fix an "Argument list too long" error that can happen depending on the environment [1]. Finally, TAs can now be built in parallel (make -jN).

Link: https://github.com/OP-TEE/build/issues/744 [1]

### DIFF
--- a/br-ext/package/optee_test_ext/optee_test_ext.mk
+++ b/br-ext/package/optee_test_ext/optee_test_ext.mk
@@ -29,30 +29,18 @@ define OPTEE_TEST_EXT_PREPARE_GP_SUITE
 endef
 
 define OPTEE_TEST_EXT_BUILD_TAS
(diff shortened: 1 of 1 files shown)


jforissier:
Depends on https://github.com/OP-TEE/optee_test/pull/740.

jenswi-linaro (commented):

jforissier:
@jenswi-linaro are you OK with the fixup patch? I have tested this in combination with optee_test 740 and with the GP test suite, and it does work. I don't like very much the output paths, I would prefer a single out/ directory under the main optee_test tree but I see no easy way and I don't want to risk breaking builds for us or users of the optee_test Makefiles.

jenswi-linaro:
Yes, looks good.

PR #750, opened 2024-05-20
Hafnium support

sudeep-holla:
This add support to run FVP with SPMC_AT_EL={n,1,2} very similar to QEMU support. Most of the work is based on the QEMU changes. PAUTH and MEMTAG in OPTEE is not add and also other features like SME and SVE yet.


fvp.mk
```diff
@@ -141,6 +146,18 @@ else
 		      EVENT_LOG_LEVEL=20
 endif
 
+TF_A_FLAGS_BL32_OPTEE  = BL32=$(OPTEE_OS_HEADER_V2_BIN)
+TF_A_FLAGS_BL32_OPTEE += BL32_EXTRA1=$(OPTEE_OS_PAGER_V2_BIN)
+TF_A_FLAGS_BL32_OPTEE += BL32_EXTRA2=$(OPTEE_OS_PAGEABLE_V2_BIN)
+TF_A_FLAGS_BL32_OPTEE += ARM_TSP_RAM_LOCATION=tdram
+
+TF_A_FLAGS_SPMC_AT_EL_n  = $(TF_A_FLAGS_BL32_OPTEE) SPD=opteed
+TF_A_FLAGS_SPMC_AT_EL_1  = BL32=$(OPTEE_OS_PAGER_V2_BIN) SPD=spmd
```
jenswi-linaro, line 166: Why is this different from `qemu_v8.mk`?

fvp/spmc_el1_optee_manifest.dts
```diff
@@ -0,0 +1,28 @@
...
+		spmc_id = <0x8000>;
+		maj_ver = <0x1>;
+		min_ver = <0x1>;
+		exec_state = <0x0>;
+		load_address = <0x0 0x6000000>;
+		entrypoint = <0x0 0x6000000>;
+		binary_size = <0x80000>;
+	};
+
+	/* VIRT_SECURE_MEM */
```
jenswi-linaro, line 23: This is a reference to a line in `hw/arm/virt.c` in the QEMU source tree. We should have something different here.

jenswi-linaro (commented):
Nice cleanup.

fvp/spmc_el1_optee_manifest.dts
sudeep-holla: Right I will fix that, it is copy paste error.

fvp.mk
sudeep-holla: For some reason passing the header doesn't work on FVP.

sudeep-holla: Thanks. I asked Balint to take a look. fvp-psa-sp.mk should be fine but can get a cleanup as well with these changes. If these changes don't break anything there(a quick look suggested so but would like to hear from Balint), then the cleanup there can be a separate change.

jenswi-linaro:

fvp/spmc_el1_optee_manifest.dts
balint-dobszay-arm, line 1: The `spmc_manifest.dts` file in the same directory already contains this. Could we use that instead? Maybe just do a rename.

fvp.mk
balint-dobszay-arm: Currently this is breaking `fvp-psa-sp.mk`. Because `SPMC_AT_EL` is not set there, it'll default to `n` so the `FF-A`-less TF-A flags are appended to the flags defined in `fvp-psa-sp.mk`

fvp/spmc_el1_optee_manifest.dts
sudeep-holla: Sure I think it makes sense. spmc_el1_partitions_manifests.dts ?

sudeep-holla:
Ah right then I will have to make the changes for `fvp-psa-sp.mk` in this set itself. I miss read the TF_A_FLAGS assignment before including` fvp.mk` in `fvp-psa-sp.mk`, assumed it won't override what  `fvp.mk` does. I will fix it as per your suggestions.

sudeep-holla:
Note the below 3 commits can be squashed into one if bisectablility is important. i.e. After first 2 commits, `fvp-psa-sp.mk` would not build for the issue @balint-dobszay-arm mentioned.

6800dcda4515 fvp: ts: Update TS/PSA makefile to use introduced SPMC_AT_EL flag
a72725c971e8 fvp: Rename and update manifest for S-EL1 SPMC
bbed154f43e2 fvp: Introducing SPMC_AT_EL

fvp/spmc_el1_optee_manifest.dts
balint-dobszay-arm: I think that's a bit misleading, since this file is not related to the partition manifests.
How about `spmc_el1_manifest.dts`?

fvp/optee_sp_manifest.dts
```diff
@@ -0,0 +1,54 @@
...
+	/* Properties */
+	description = "op-tee";
+	ffa-version = <0x00010001>; /* 31:16 - Major, 15:0 - Minor */
+	uuid = <0xe0786148 0xe311f8e7 0x02005ebc 0x1bc5d5a5>;
+	id = <1>;
+	execution-ctx-count = <8>;
+	exception-level = <2>; /* S-EL1 */
+	execution-state = <0>; /* AARCH64 */
+	load-address = <0x6280000>;
+	mem-size = <0xd00000>; 	/* OP-TEE specific extension */
```
odeprez, line 26: This mem-size value relates to qemu memory layout.
On FVP Trusted DRAM spans 0x6000000-0x7ffffff so mem-size might expand up to 0x1d80000

fvp/optee_sp_manifest.dts
```diff
@@ -0,0 +1,54 @@
...
+	entrypoint-offset = <0x4000>;
+	xlat-granule = <0>; /* 4KiB */
+	boot-order = <0>;
+	messaging-method = <0x3>; /* Direct request/response supported. */
+	ns-interrupts-action = <1>; /* NS_ACTION_ME */
+	notification-support;
+	managed-exit;
+	run-time-model = <1>; /* SP pre-emptible. */
```
odeprez, line 34: run-time-model isn't consumed by hafnium.

fvp/spmc_el2_optee_sp_manifest.dts
```diff
@@ -0,0 +1,110 @@
...
+	};
+
+	hypervisor {
+		compatible = "hafnium,hafnium";
+		vm1 {
+			is_ffa_partition;
+			load_address = <0x6280000>;
+			debug_name = "op-tee";
+			vcpu_count = <8>;
+			mem_size = <0xd00000>;
```
odeprez, line 33: Check if this could be extended?

fvp/spmc_el1_partitions_manifest.dts
```diff
@@ -13,13 +13,19 @@
...
 		load_address = <0x0 0x6000000>;
 		entrypoint = <0x0 0x6000000>;
 		binary_size = <0x80000>;
 	};
 
+	/* Secure world memory */
+	memory@6000000 {
+		device_type = "memory";
+		reg = <0x0 0x6000000 0x1000000>; /* Trusted DRAM */
+	};
```
odeprez, line 27: Is this consumed by OP-TEE SPMC?

jenswi-linaro: Currently not, but I assume Hafnium does and it would be a good idea to do the same when OP-TEE is the SPMC.

fvp/optee_sp_manifest.dts
sudeep-holla: I wasn't aware of that, I think I copied it from TF-A FVP or from QEMU. I will drop it. I also noticed
notification-support and managed-exit are also not needed. I have dropped those 2 in my change request to
TF-A (https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/29047 https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/29041).

The idea is to get them updated or added in TF-A and drop them here when we move to the TF-A release containing those files so that only one copy can be maintained.

fvp/spmc_el1_partitions_manifest.dts
sudeep-holla: Again I can fix that for spmc_el1_partitions_manifest.dts, but I assumed the load address used in that file to load trusted services partitions(driven from fvp-psa-sp.mk) must not overlap. Sorry I don't understand all the details of these manifest files. I assumed either the copy here or in TF-A are up-to-date but it looks like they are not :(

sudeep-holla:
Any updates on this ? I (mostly work done by Olivier) have patches to upgrade TF-A an Hafnium to v2.11. I just don't want to expand the scope of this MR to include that too. I will post them separate once this is merged.

jforissier:
For "fvp: ts: Update TS/PSA makefile to use introduced SPMC_AT_EL flag": LGTM, please apply:

Then this will be ready to merge. Thanks!

PR #754, opened 2024-05-24
stm32mp1: define rootfs from its partition UUID

etienne-lms:
Change stm32mp1 SDcard image to define the rootfs partition by its GPT partition UUID instead of its mmcblk device and partition indices. This is required for Linux kernel v6.2 onward that do not guarantee device indices when several devices of a same driver are initialized.


etienne-lms:
Tag applied. Thanks.

PR #755, opened 2024-05-24
stm32mp1: use tee-supplicant option --unique-rpmb

etienne-lms:
Use **tee-supplicant** recently added option `--unique-rpmb` to let **tee-supplicant** find the unique RPMB partition device file since Linux kernels v6.2 and later do not guaranty the mmcblk device index passed to OP-TEE configuration option `CFG_RPMB_FS_DEV_ID` is reliable.

This change depends on https://github.com/OP-TEE/optee_client/pull/381.


etienne-lms:
I close this P-R. https://github.com/OP-TEE/optee_client/pull/381#issuecomment-2166032817 proposes an alternate way to identify device file of the embedded eMMC.

PR #757, opened 2024-05-30
qemu_v8: enable max CPU for Xen

jenswi-linaro:
Enable the QEMU max CPU for Xen. With this change simplify the configuration a little.


jforissier (commented):

jenswi-linaro:
Tag applied

PR #758, opened 2024-06-07
qemu_v8: fix non-FF-A Xen configuration

jenswi-linaro:
Xen doesn't work with SME enabled. For SPMC_AT_EL == 1, 2, or 3 SME is already disabled when starting QEMU, but not for SPMC_AT_EL=n and XEN_BOOT=y, so fix this with another check.

Fixes: f68f33e84595 ("qemu_v8: enable max CPU for Xen")


jforissier (commented):

jenswi-linaro:
Tags applied.

PR #760, opened 2024-06-14
stm32mp1: retrieve unique eMMC CID for tee-supplicant

etienne-lms:
Retrieve the eMMC CID if only one eMMC device is found. We expect it to be used for OP-TEE RPMB secure storage. This change allows OS based on Linux kernel v6.2 or later to be more reliable on identifying the eMMC mmcblk device as since Linux kernel v6.2 the device index assigned by the kernel may change from one cold boot to another when they are several mmcblk devices on the platform, for example a plugged SD card and a soldered eMMC device.


jforissier (commented):

etienne-lms:
Tag applied.
Thanks again for the pointer. It seems these instruction are deployed in Yocto recipes since quite a while :)

PR #762, opened 2024-06-14
toolchains.mk, get_clang.sh: upgrade to Clang 18.1.6

jforissier:
Update the Clang version from 12.0.0 to 18.1.6. Unfortunately, newer versions of the LLVM toolchain do not make available the pre-built compiler-rt packages for arm64 and armhf. Therefore, the way we obtain the toolchain is changed. I created a project with a Makefile and a Dockerfile [1] which allows to build Clang from scratch with the proper compiler-rt libraries for OP-TEE cross-compilation. Nothing really special, yet not a configuration that is availale pre-built in the major Linux distributions (Unbuntu). The problem is that on an x84_64 system, there is no way to install the compiler-rt builtins for a different architecture (libclang_rt.builtins-{aarch64,armhf}.a).

The new method to install Clang is thererfore to pull a pre-built image (called jforissier/clang-<version>-<arch>) from the Docker Hub, then simply extract the files contained inside, which is what the modified get_clang.sh does.

Link: https://github.com/jforissier/docker_clang [1]

### DIFF
--- a/get_clang.sh
+++ b/get_clang.sh
@@ -7,43 +7,19 @@
 # Clang is configured to be able to cross-compile to all the supported
 # architectures by default (see <clang path>/bin/llc --version) which is great,
 # but compiler-rt is included only for the host architecture. Therefore we need
-# to combine several packages into one, which is the purpose of this script.
+# a custom build of clang, which is achieved by [1].
+# This script retrieves a pre-compiled image from the Docker Hub and extracts
+# the compiler.
 
 [ "$1" ] || { echo "Usage: get_clang.sh version [path]"; exit 1; }
 
 VER=${1}
 DEST=${2:-./clang-${VER}}
-X86_64=clang+llvm-${VER}-x86_64-linux-gnu-ubuntu-16.04
-AARCH64=clang+llvm-${VER}-aarch64-linux-gnu
-ARMV7A=clang+llvm-${VER}-armv7a-linux-gnueabihf
 
-set -x
+ARCH=$(uname -m)
+set -e -x
 
-TMPDEST=${DEST}_tmp${RANDOM}
-
-if [ -e ${TMPDEST} ]; then
-  echo Error: ${TMPDEST} exists
-  exit 1
-fi
-
-function cleanup() {
-  rm -f ${X86_64}.tar.xz ${AARCH64}.tar.xz ${ARMV7A}.tar.xz
-  rm -rf ${AARCH64} ${ARMV7A}
-}
-
-trap "{ exit 2; }" INT
-trap cleanup EXIT
-
-(wget -nv https://github.com/llvm/llvm-project/releases/download/llvmorg-${VER}/${X86_64}.tar.xz && tar xf ${X86_64}.tar.xz) &
-pids=$!
-(wget -nv https://github.com/llvm/llvm-project/releases/download/llvmorg-${VER}/${AARCH64}.tar.xz && tar xf ${AARCH64}.tar.xz) &
-pids="$pids $!"
-(wget -nv https://github.com/llvm/llvm-project/releases/download/llvmorg-${VER}/${ARMV7A}.tar.xz && tar xf ${ARMV7A}.tar.xz) &
-pids="$pids $!"
-
-wait $pids || exit 1
-
-mv ${X86_64} ${TMPDEST} || exit 1
-cp ${AARCH64}/lib/clang/${VER}/lib/linux/* ${TMPDEST}/lib/clang/${VER}/lib/linux || exit 1
-cp ${ARMV7A}/lib/clang/${VER}/lib/linux/* ${TMPDEST}/lib/clang/${VER}/lib/linux || exit 1
-mv ${TMPDEST} ${DEST}
+id=$(docker create jforissier/clang-${VER}-${ARCH})
+docker cp $id:/root/clang-${VER} ${DEST}
+docker rm ${id}
+docker rmi jforissier/clang-${VER}-${ARCH}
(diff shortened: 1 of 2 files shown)


jbech-linaro:
Commit message:
s/Unbuntu/Ubuntu/

Seems like a bit complicated way to get the stuff needed, but let's use until we have found a better way.

get_clang.sh
```diff
@@ -7,43 +7,19 @@
...
-wait $pids || exit 1
-
-mv ${X86_64} ${TMPDEST} || exit 1
-cp ${AARCH64}/lib/clang/${VER}/lib/linux/* ${TMPDEST}/lib/clang/${VER}/lib/linux || exit 1
-cp ${ARMV7A}/lib/clang/${VER}/lib/linux/* ${TMPDEST}/lib/clang/${VER}/lib/linux || exit 1
-mv ${TMPDEST} ${DEST}
+id=$(docker create jforissier/clang-${VER}-${ARCH})
+docker cp $id:/root/clang-${VER} ${DEST}
+docker rm ${id}
+docker rmi jforissier/clang-${VER}-${ARCH}
```
jbech-linaro, line 25: I suppose we need to Update the `Dockerfile` as well for [this](https://optee.readthedocs.io/en/latest/building/prerequisites.html#prerequisites), to show "docker", since our instructions actually is for people building from the host environment.

jenswi-linaro:
If I understand correctly, you're distributing a binary you've built yourself. Can we build the compiler using buildroot instead?

jforissier:
> If I understand correctly, you're distributing a binary you've built yourself.

Yes

> Can we build the compiler using buildroot instead?

Good idea. It may be a bit much (I mean, it will take significant time to compile on a typical development PC), but we already do a similar build for GCC on Arm64 hosts, and it would certainly make maintenance easier.

jenswi-linaro:
> > If I understand correctly, you're distributing a binary you've built yourself.
> 
> Yes

Using that approach in CI is fine, but recommending others to install that on their development machines is something else.

> 
> > Can we build the compiler using buildroot instead?
> 
> Good idea. It may be a bit much (I mean, it will take significant time to compile on a typical development PC), but we already do a similar build for GCC on Arm64 hosts, and it would certainly make maintenance easier.

Easier maintenance sounds like a selling point. :-)

jforissier:
> > > If I understand correctly, you're distributing a binary you've built yourself.
> > 
> > 
> > Yes
> 
> Using that approach in CI is fine, but recommending others to install that on their development machines is something else.

We're not installing anything, just unpacking the Clang binaries in `<proj_root>/clang-18.1.6` in the same way we used to download and extract the Clang 12 binary packages into `<proj_root>/clang-12.0.0` in the previous version of `make clang-toolchains`. The main difference being, instead of the binaries coming from llvm.org (which doesn't have everything we need anymore), they come from a Docker image I built myself and stored on the Docker Hub.

> 
> > > Can we build the compiler using buildroot instead?
> > 
> > 
> > Good idea. It may be a bit much (I mean, it will take significant time to compile on a typical development PC), but we already do a similar build for GCC on Arm64 hosts, and it would certainly make maintenance easier.
> 
> Easier maintenance sounds like a selling point. :-)

:)

jenswi-linaro:
> > > > If I understand correctly, you're distributing a binary you've built yourself.
> > > 
> > > 
> > > Yes
> > 
> > 
> > Using that approach in CI is fine, but recommending others to install that on their development machines is something else.
> 
> We're not installing anything, just unpacking the Clang binaries in `<proj_root>/clang-18.1.6` in the same way we used to download and extract the Clang 12 binary packages into `<proj_root>/clang-12.0.0` in the previous version of `make clang-toolchains`. The main difference being, instead of the binaries coming from llvm.org (which doesn't have everything we need anymore), they come from a Docker image I built myself and stored on the Docker Hub.

That's not my point. The point is that the binaries are executing on the development machines.

jforissier:
I could not find a way to build Clang with Buildroot, not with all we need anyways. There is a `clang` command in the `host` directory but it is not a cross compiler. Then I can find `libclang.so` and `llvm-*` tools under `target` but no `clang` command and no `libclang_rt` libraries. The documentation is quite scarce... Moreover it takes forever to build because Buildroot first builds a specific version of GCC with whatever GCC is available, then it builds LLVM.
I think people who want to build everything themselves are better off re-using the Dockerfile: https://github.com/jforissier/docker_clang/blob/master/Dockerfile

get_clang.sh
jforissier: Good catch. TBH I also find it a bit much to require Docker to simply download a toolchains package, especially since it brings some complications when running inside Docker such as in CI (add `-v /var/run/docker.sock:/var/run/docker.sock`, then if the user in the container is not root you need to map the docker group from the host).
Perhaps we could find a place to store a `.tar.bz2`?

jbech-linaro: Never tried it, but I wonder if https://git-lfs.com/ is a solution for this problem?

jforissier: Ha! I'm just reading https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-git-large-file-storage :wink: but unfortunately there are very strict limits (1GB/month).
https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-storage-and-bandwidth-usage

jforissier: I believe I should first add the Clang build to this repository (such as add the proper Dockerfile and a new target `make clang-toolchains-build`). Then when we find a place to store the binary we can update `make clang-toolchains`. I will create a new PR for that.

PR #765, opened 2024-06-18
toolchain.mk: add clang-toolchains-build to build Clang from sources

jforissier:
Add a makefile target to download and build the Clang 18.1.7 sources and install the compiler into <PRJ_ROOT>/toolchains/clang-18.1.7/bin.

Usage example:

 $ make clang-toolchains-build
 $ PATH=$(pwd)/../toolchains/clang-18.1.7/bin:$PATH \
     make -j$(nproc) COMPILER=clang check

The Clang build takes some significant time (about 12 minutes on my machine, a 24-core Intel Core i9-12900K) so may want to store a pre-built package somewhere in the future and have "make clang-toolchains" download that package. In the mean time, this new target allows to test OP-TEE with a recent LLVM toolchain.

### DIFF
--- a/clang/Dockerfile
+++ b/clang/Dockerfile
@@ -1,0 +1,82 @@
+ARG VER_MAJ
+ARG VER
+ARG INSTALL_DIR=/root/install
+
+FROM ubuntu:22.04 as clang-build
+ARG VER_MAJ
+ARG VER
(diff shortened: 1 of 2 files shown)


jenswi-linaro:

PR #766, opened 2024-07-04
AArch64 host fixes

jenswi-linaro:


jforissier:
LGTM but TBH the configuration is error-prone. Perhaps we should have a few comments here and there to remind about what needs to be consistent? Anyways:

jenswi-linaro:
> LGTM but TBH the configuration is error-prone. Perhaps we should have a few comments here and there to remind about what needs to be consistent?

I agree, I'll see what I can do in another PR.

Tag applied.

PR #767, opened 2024-07-04
Refactor and comment downloaded and compiled toolchains

jenswi-linaro:
Move the common parts of br-ext/configs/toolchain-aarch32-sdk and br-ext/configs/toolchain-aarch64-sdk into
br-ext/configs/toolchain-common-sdk.
br-ext/configs/toolchain-aarch32-sdk was previously using BR2_TOOLCHAIN_EXTERNAL_GCC_11=y and is now by using br-ext/configs/toolchain-common-sdk updated to use the correct BR2_TOOLCHAIN_EXTERNAL_GCC_14=y.

Add comments where to update if compiled or downloaded toolchain is updated.


toolchain.mk
```diff
@@ -57,11 +57,14 @@ endef
 
 ifeq ($(UNAME_M),x86_64)
 ifeq ($(ARCH),arm)
+# Please keep in sync with br-ext/configs/toolchain-aarch32
+# and below for x86_64 host
```
jforissier, line 61: You mean for aarch64 host?

br-ext/configs/sdk-common
```diff
@@ -1,3 +1,5 @@
+# Please keep GCC version and headers in sync with
+# br-ext/configs/toolchain-external-sdk
```
jforissier, line 2: You mean `toolchain-common-sdk`?

jforissier (commented):
Two comments below, other than that looks good.

toolchain.mk
jenswi-linaro: You're right, I'll update.

jenswi-linaro:
Comments addressed and tag applied.

PR #768, opened 2024-07-05
QEMU U-Boot version update

igoropaniuk:
Depends on https://github.com/OP-TEE/manifest/pull/285

### DIFF
--- a/kconfigs/u-boot_qemu_virt_v7.conf
+++ b/kconfigs/u-boot_qemu_virt_v7.conf
@@ -1,4 +1,4 @@
-CONFIG_SYS_TEXT_BASE=0x60000000
+CONFIG_TEXT_BASE=0x60000000
 CONFIG_BOOTCOMMAND="fdt addr ${fdt_addr} && fdt resize 1000 && smhload zImage ${kernel_addr_r} && smhload rootfs.cpio.gz ${ramdisk_addr_r} ramdisk_addr_end &&  setenv bootargs console=ttyAMA0,115200 earlyprintk=serial,ttyAMA0,115200 && setexpr ramdisk_addr_end ${ramdisk_addr_end} + 1 && fdt chosen ${ramdisk_addr_r} ${ramdisk_addr_end} && bootz ${kernel_addr_r} - ${fdt_addr}"
 CONFIG_SEMIHOSTING=y
 CONFIG_ENV_IS_NOWHERE=y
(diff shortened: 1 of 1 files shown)


jenswi-linaro:
This is a bit late in the release process, can it wait until after the release?

igoropaniuk:
@jenswi-linaro sure, it was planned that way :)

igoropaniuk:
any comments/suggestions?

kconfigs/u-boot_qemu_virt_v7.conf
```diff
@@ -1,5 +1,5 @@
 CONFIG_TEXT_BASE=0x60000000
-CONFIG_BOOTCOMMAND="fdt addr ${fdt_addr} && fdt resize 1000 && smhload zImage ${kernel_addr_r} && smhload rootfs.cpio.gz ${ramdisk_addr_r} ramdisk_addr_end &&  setenv bootargs console=ttyAMA0,115200 earlyprintk=serial,ttyAMA0,115200 && setexpr ramdisk_addr_end ${ramdisk_addr_end} + 1 && fdt chosen ${ramdisk_addr_r} ${ramdisk_addr_end} && bootz ${kernel_addr_r} - ${fdt_addr}"
+CONFIG_BOOTCOMMAND="fdt addr ${fdt_addr} && fdt resize 1000 && load hostfs - ${kernel_addr_r} zImage && load hostfs - ${ramdisk_addr_r} rootfs.cpio.gz && setenv bootargs console=ttyAMA0,115200 earlyprintk=serial,ttyAMA0,115200 && fdt chosen ${ramdisk_addr_r} ${filesize} && bootz ${kernel_addr_r} - ${fdt_addr}"
```
jenswi-linaro, line 2: When is `${filesize}` set?

igoropaniuk: it will be set set by `load` cmd, in our case we use the value from the latest `load` execution (`load hostfs - ${ramdisk_addr_r} rootfs.cpio.gz`)
from: https://github.com/u-boot/u-boot/blob/v2024.07/doc/usage/semihosting.rst
> The load command will set the filesize variable with the size of the file. The fdt chosen command has been updated to take a size instead of an end address.

jenswi-linaro: Got it, thanks.

jenswi-linaro:
Does this depends on a newer version of U-Boot or can it be used with the already present version?

igoropaniuk:
@jenswi-linaro yes, both changes require new U-Boot

jenswi-linaro:
> @jenswi-linaro yes, both changes require new U-Boot

I could have read the description of the PR. :-)

igoropaniuk:
@jenswi-linaro tag is applied, thanks!

PR #772, opened 2024-07-23
qemu: add tmux support.

jxlpzqc:
Whether it is gnome-terminal, konsole or xterm, they all have key limitations, i,e, they require desktop enviroment, which make it not convenient to develop and test through ssh connection.

Some terminal multiplexers like tmux, screen are more generic and they are supposed to be supported.

This patch add tmux support for `run-only` target.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -458,9 +458,25 @@ define launch-terminal
 		$(LAUNCH_TERMINAL) "$(BUILD_PATH)/soc_term.py $(1)" &
 endef
 else
+tmux := $(TMUX)
+tmux_window := $(shell echo OPTEE_$$RANDOM)
 gnome-terminal := $(shell command -v gnome-terminal 2>/dev/null)
 konsole := $(shell command -v konsole 2>/dev/null)
 xterm := $(shell command -v xterm 2>/dev/null)
+
+ifdef tmux
+define launch-terminal
+	@if tmux list-windows -F '#W' | grep -q $(tmux_window); then \
+		nc -z 127.0.0.1 $(1) || \
+			tmux split-window -d -h -t $(tmux_window) "$(BUILD_PATH)/soc_term.py $(1)" ; \
+	else \
+		nc -z 127.0.0.1 $(1) || \
+			tmux new-window -d -n $(tmux_window) "$(BUILD_PATH)/soc_term.py $(1)" ; \
+	fi
+
+	@echo "* $(2)'s terminal has been spawned in $(tmux_window)."
+endef
+else
 ifdef gnome-terminal
 define launch-terminal
 	@nc -z  127.0.0.1 $(1) || \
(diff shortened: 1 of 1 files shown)


common.mk
```diff
@@ -458,9 +458,25 @@ define launch-terminal
 		$(LAUNCH_TERMINAL) "$(BUILD_PATH)/soc_term.py $(1)" &
 endef
 else
+tmux := $(TMUX)
+tmux_window := $(shell echo OPTEE_TEST_$$RANDOM)
```
jforissier, line 462: I suggest `OPTEE_$$RANDOM` because it's shorter and more generic (`optee_test` is the name of the xtest project)

jforissier (commented):
Hi @jxlpzqc,

Thanks for the patch! As a `tmux` user myself, I find this quite useful. I have a minor suggestion, please see below. In any case:

Edit: preferably, remove the full stop in the commit subject (`qemu: add tmux support`).

common.mk
```diff
@@ -458,9 +458,25 @@ define launch-terminal
...
 konsole := $(shell command -v konsole 2>/dev/null)
 xterm := $(shell command -v xterm 2>/dev/null)
+
+ifdef tmux
+define launch-terminal
+	@if tmux list-windows -F '#W' | grep -q $(tmux_window); then \
+		nc -z 127.0.0.1 $(1) || \
+			tmux split-window -d -h -t $(tmux_window) "$(BUILD_PATH)/soc_term.py $(1)" ; \
+	else \
+		nc -z 127.0,0.1 $(1) || \
```
etienne-lms, line 473: typo: s/127.0,0.1/127.0.0.1/

etienne-lms (commented):
I failed to use this change with tmux.
I tried `TMUX=/usr/bin/tmux make run-only` and `TMUX=tmux make run-only`. Both fail with message `error connecting to /usr/bin/tmux (Permission denied)` despite:
```shell
$ ls -l `which tmux`
-rwxr-xr-x 1 root root 692192 Jan 31  2023 /usr/bin/tmux
```
Did I miss something?

common.mk
jxlpzqc: Hi, @etienne-lms, when in tmux environment, $TMUX env is automatically set by tmux.
This feature helps developer who is using tmux to open terminal in new tmux window, as the picture below, Two terminals are created in `OPTEE_12293`, but does not support running outside tmux.

20240724150636_rec_

Besides, thanks for reviewing and your finding typo mistake, I will fix it right now.

etienne-lms: Thanks, I better understand now. Works ok (as you would have expected I guess).

common.mk
```diff
@@ -484,6 +500,7 @@ endif
 endif
 endif
 endif
+endif
```
etienne-lms, line 503: Maybe `endif #tmux` to help not getting to lost in the condition cases ?
```make
...
check-terminal := @echo "Error: could not find gnome-terminal, konsole nor xterm" ; false
endif
endif
endif
endif #tmux
endif #LAUNCH_TERMINAL

etienne-lms (commented):
`Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>` (FWIW)

common.mk
jxlpzqc: Good idea 👍

PR #776, opened 2024-08-14
 qemuv8: fix to enable Rust by default on x86_64 hosts only

bingzhux:
When building on x86_64 host, the rust application won't be built by default, this is because the "UNAME_M" is not yet defined/assigned. 

We can move "include common.mk" line above, because in common.mk file the "UNAME_M" is assigned with correct value.

### DIFF
--- a/qemu_v8.mk
+++ b/qemu_v8.mk
@@ -29,13 +29,13 @@ GICV3 = y
 QEMU_VIRTFS_AUTOMOUNT = y
 endif
 
-include common.mk
-
 # Option to enable Rust examples
 # Currently supported only on x86_64 hosts
-ifeq ($(UNAME_M),x86_64)
+ifeq ($(shell uname -m),x86_64)
 RUST_ENABLE ?= y
 endif
+
+include common.mk
 
 DEBUG ?= 1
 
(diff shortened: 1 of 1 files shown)


jbech-linaro:
@jforissier , I might be wrong, be didn't we make it behave like this intentionally, since building Rust added a lot of build time?

qemu_v8.mk
```diff
@@ -29,14 +29,14 @@ GICV3 = y
 QEMU_VIRTFS_AUTOMOUNT = y
 endif
 
+include common.mk
+
```
jforissier, line 33: Please leave the include where it is and replace `$(UNAME_M)` below with `$(shell uname -m)`. Reason being, `common.mk` uses `$(RUST_ENABLE)`.

jforissier:
> @jforissier , I might be wrong, be didn't we make it behave like this intentionally, since building Rust added a lot of build time?

No, it doesn't take that long (~ 30 seconds on my build machine). It used to be very slow when Rust support was introduced but not anymore.

jbech-linaro:
> No, it doesn't take that long (~ 30 seconds on my build machine). It used to be very slow when Rust support was introduced but not anymore.

Ok, glad to hear that I was wrong :)

qemu_v8.mk
bingzhux: good catch. (I thought the rust build was OK, but actually you're right, the rust examples still not built after checking the build logs)

I delete the "patch-1" branch by mistake just now, which resulted this PR invalid... I will go create a new PR, thanks.

PR #777, opened 2024-08-15
qemuv8: fix to enable Rust by default on x86_64 hosts only

bingzhux:
Change (UNAME_M) to (shell uname -m) to detect host, so that "RUST_ENABLE" can be correctly set on x86_64 hosts.


jforissier (commented):

bingzhux:
Thanks，updated with signed-off, tested-by, and reviewed-by. 
It should be ready to merge.

PR #780, opened 2024-09-25
Ftpm

jenswi-linaro:


common.mk
```diff
@@ -271,6 +275,10 @@ ifeq ($(XEN_BOOT),y)
 DEFCONFIG_XEN=--br-defconfig build/br-ext/configs/xen.conf
 endif
 
+ifneq ($(CFG_MS_TPM_20_REF),y)
```
jbech-linaro, line 278: I guess this should be as line 567? I.e., 
```
ifneq ($(CFG_MS_TPM_20_REF),)
```
Since the line (47) setting `CFG_MS_TPM_20_REF` is a path and not `y`.

jenswi-linaro: Yes, you're right.

jenswi-linaro:
Replaced by https://github.com/OP-TEE/build/pull/787

PR #781, opened 2024-10-04
FVP: bump TF-A/Hafnium v2.11

odeprez:
Build changes for bumping TF-A to v2.11 tag.
Hafnium pinned to a known good hash for building with clang 18.
clang toolchain was removed from hafnium prebuilts. Expect clang to be provided in PATH


common.mk
```diff
@@ -51,7 +51,10 @@ WITH_CXX_TESTS			?= y
 endif
 endif
 
-CCACHE ?= $(shell which ccache) # Don't remove this comment (space is needed)
+CCACHE_present := $(shell which ccache)
+ifneq ($(CCACHE_present),)
+CCACHE ?= $(CCACHE_present) # Don't remove this comment (space is needed)
```
jenswi-linaro, line 56: What problem are you trying to solve with this?

jenswi-linaro (commented):
Comments for commit "common.mk: fix CCACHE handling"

Please add a description of the problem you're fixing in the commit message.

jenswi-linaro (commented):
Comments for "fvp: use BRANCH_PROTECTION"

https://trustedfirmware-a.readthedocs.io/en/latest/design/firmware-design.html#armv8-3-a says that `CTX_INCLUDE_PAUTH_REGS` must be set to 1 for the secure world to use pointer authentication.

common.mk
odeprez: Hi Jens, TF-A v2.11 introduced some build system hardening implying CROSS_COMPILE must contain the exact path to the toolchain. For the case when ccache is not installed, this statement:
https://github.com/OP-TEE/build/blob/b80d8cb05ef6b789e536631f6e8d88118e8483b4/common.mk#L195
breaks TF-A build because CROSS_COMPILE gets an extra heading space character. TF-A attempts invoking $(CROSS_COMPILE)gcc but but resolves to 
CROSS_COMPILE=" <...>/build/../toolchains/aarch64/bin/aarch64-linux-gnu-"  instead of
CROSS_COMPILE="<...>/build/../toolchains/aarch64/bin/aarch64-linux-gnu-"

odeprez:
> Comments for "fvp: use BRANCH_PROTECTION"
> 
> https://trustedfirmware-a.readthedocs.io/en/latest/design/firmware-design.html#armv8-3-a says that `CTX_INCLUDE_PAUTH_REGS` must be set to 1 for the secure world to use pointer authentication.

BRANCH_PROTECTION shall be the user facing build option when it comes to enabling PAuth/BTI or combination of both.

Other toggles CTX_INCLUDE_PAUTH_REGS, ENABLE_PAUTH, ENABLE_BTI, BP_OPTION have become internal build system flags.
e.g. CTX_INCLUDE_PAUTH_REGS is governed by BRANCH_PROTECTION:
https://git.trustedfirmware.org/plugins/gitiles/TF-A/trusted-firmware-a/+/refs/heads/master/Makefile#527
https://git.trustedfirmware.org/plugins/gitiles/TF-A/trusted-firmware-a/+/refs/heads/master/Makefile#545

I appreciate documentation doesn't convey that clearly as you pointed out.

common.mk
jenswi-linaro: Thanks for the detailed explanation, please add a short comment in the makefile to help others understand what's happening. Perhaps:
```
# Only set CCACHE if it's pointing to something to avoid prefixing CROSS_COMPILE
# with whitespace. TF-A will not build with whitespace first in CROSS_COMPILE.
```

jenswi-linaro:
Got it, thanks.

jenswi-linaro:

jforissier (commented):
- For "common.mk: fix CCACHE handling": please add a commit description. You may use the same text that Jens suggested (add it both to the description and as a comment).
- For commit " fvp: specify hafnium PLATFORM build option": mention that TFA v2.11 needs this in the commit description
- For commit " fvp: add secure uart interrupt in device region": s/uart/UART/ in the subject

With that and for all patches:

common.mk
odeprez: Done.

jforissier:
@odeprez please add the R-b tags, it makes things easier for me :wink:

odeprez:
Hi,

> @odeprez please add the R-b tags, it makes things easier for me 😉
Sorry what is this :) ?

Before merging, let's have the confirmation that nothing breaks with the docker image you shared in another thread.
I have in mind that using clang 18 in the docker image may break qemu because its manifest is still pinned to v2.10
We have the following mapping:
hafnium v2.10 => clang 15
hafnium v2.11 => clang 18

In this eventual case, we may have to merge qemu manifest + build changes (bumping to TF-A/hafnium v2.11) altogether.

jforissier:
> Hi,
> 
> > @odeprez please add the R-b tags, it makes things easier for me 😉
> 
> Sorry what is this :) ?

I mean adding the `Reviewed-by: xxx xxx <xxx@xxx>` lines that Jens and I gave above. Copy them after your `Signed-off-by:`. In all commits.

> 
> Before merging, let's have the confirmation that nothing breaks with the docker image you shared in another thread. I have in mind that using clang 18 in the docker image may break qemu because its manifest is still pinned to v2.10 We have the following mapping: hafnium v2.10 => build with clang 15 hafnium v2.11 => build with clang 18
> 
> In this eventual case, we may have to merge qemu manifest + build changes (bumping to TF-A/hafnium v2.11) altogether.

Yes that's what we usually do.

> Or are there separate docker configs for fvp and qemu manifests?

No. We don't test FVP in the OP-TEE CI. There is just a build of the optee_os project only, i.e., "make PLATFORM=vexpress-fvp" in optee_os as part of the "make (multi-platform)" job ([link](https://github.com/OP-TEE/optee_os/blob/fc80dabbd5a7d2d09834fc3f6175ef7eac3d6fd7/.github/workflows/ci.yml#L57)). But there is no full build + run like we have for QEMUv7 or QEMUv8, i.e., running "make check" in the build directory ([link](https://github.com/OP-TEE/optee_os/blob/fc80dabbd5a7d2d09834fc3f6175ef7eac3d6fd7/.github/workflows/ci.yml#L352)).

odeprez:
> > Hi,
> > > @odeprez please add the R-b tags, it makes things easier for me 😉
> > 
> > 
> > Sorry what is this :) ?
> 
> I mean adding the `Reviewed-by: xxx xxx <xxx@xxx>` lines that Jens and I gave above. Copy them after your `Signed-off-by:`. In all commits.

Done.

> 
> > Before merging, let's have the confirmation that nothing breaks with the docker image you shared in another thread. I have in mind that using clang 18 in the docker image may break qemu because its manifest is still pinned to v2.10 We have the following mapping: hafnium v2.10 => build with clang 15 hafnium v2.11 => build with clang 18
> > In this eventual case, we may have to merge qemu manifest + build changes (bumping to TF-A/hafnium v2.11) altogether.

Actually scratch that. qemu v8 is pinned to TF-A/Hafnium v2.10 and qemu_v8.mk adjusts PATH to use clang toolchain sourced from hafnium prebuilts submodule. So this should be fine eventually.

> 
> Yes that's what we usually do.
> > Or are there separate docker configs for fvp and qemu manifests?
> No. We don't test FVP in the OP-TEE CI. There is just a build of the optee_os project only, i.e., "make PLATFORM=vexpress-fvp" in optee_os as part of the "make (multi-platform)" job ([link](https://github.com/OP-TEE/optee_os/blob/fc80dabbd5a7d2d09834fc3f6175ef7eac3d6fd7/.github/workflows/ci.yml#L57)). But there is no full build + run like we have for QEMUv7 or QEMUv8, i.e., running "make check" in the build directory ([link](https://github.com/OP-TEE/optee_os/blob/fc80dabbd5a7d2d09834fc3f6175ef7eac3d6fd7/.github/workflows/ci.yml#L352)).

I was able to build FVP (SPMC_AT_EL=2) configuration using the docker file sourcing clang 18.

PR #785, opened 2024-10-14
QEMU v8: upgrade TF-A to v2.11

jenswi-linaro:


jforissier (commented):
For commit "qemu_v8: build hafnium using OP-TEE's clang toolchain":

I suggest to replace "Use the toolchain provided in PATH instead" with "Instead, use OP-TEE's clang toolchain ('make clang-toolchains') if it exists, or the one already in PATH".

qemu_v8.mk
```diff
@@ -161,9 +161,7 @@ TF_A_FLAGS ?= \
 	PLAT=qemu \
 	QEMU_USE_GIC_DRIVER=$(TFA_GIC_DRIVER) \
 	ENABLE_SVE_FOR_NS=2 \
-	ENABLE_SME_FOR_NS=2 \
 	ENABLE_SVE_FOR_SWD=1 \
-	ENABLE_SME_FOR_SWD=1 \
```
jforissier, line 166: Comment on "qemu_v8: enable SME for SPMC_AT_EL=2".
Why are these two lines not needed when SPMC_AT_EL=n? We still have QEMU_SME=on in that case.

jenswi-linaro: That's a good question. I'll restore the lines and remove the corresponding line below.

jenswi-linaro:
Tags applied

jforissier (commented):
For "qemu_v8: enable SME for SPMC_AT_EL=2":

jenswi-linaro:
Squashed and tag applied.

jenswi-linaro:
Ready to merge?

PR #787, opened 2024-10-23
Ftpm2

jenswi-linaro:
This PR replaces https://github.com/OP-TEE/build/pull/780


common.mk
```diff
@@ -556,6 +557,7 @@ endif
 
 CFG_IN_TREE_EARLY_TAS ?= trusted_keys/f04a0fe7-1f5d-4b9b-abf7-619b85b4ce8c
 
+
```
etienne-lms, line 560: can discard this change

common.mk
```diff
@@ -582,26 +588,22 @@ optee-os-devkit:
...
-	$(FTPM_FLAGS) $(MAKE) -C $(FTPM_PATH)
+	$(FTPM_FLAGS) $(MAKE) -C $(OPTEE_FTPM_PATH)
 endif
 
 .PHONY: ftpm-clean
 ftpm-clean:
 ifeq ($(MEASURED_BOOT_FTPM),y)
 ftpm-clean:
-	-$(FTPM_FLAGS) $(MAKE) -C $(FTPM_PATH) clean
+	-$(FTPM_FLAGS) $(MAKE) -C $(OPTEE_FTPM_PATH)
```
etienne-lms, line 608: restore `clean` target.

common.mk
```diff
@@ -566,6 +568,10 @@ OPTEE_OS_COMMON_FLAGS ?= \
 	DEBUG=$(DEBUG) \
 	CFG_IN_TREE_EARLY_TAS="$(CFG_IN_TREE_EARLY_TAS)"
 
+ifeq ($(MEASURED_BOOT_FTPM),y)
+OPTEE_OS_COMMON_EXTRA_FLAGS += EARLY_TA_PATHS=$(OPTEE_FTPM_PATH)/out/bc50d971-d4c9-42c4-82cb-343fb7f37896.stripped.elf
+optee-os-common: ftpm
```
etienne-lms, line 573: Shouldn't `.PHONY: optee-os-common` be set before 1st occurrence of the `optee-os-common` target?

common.mk
```diff
@@ -582,26 +588,22 @@ optee-os-devkit:
...
-# OPTEE_OS_TA_DEV_KIT_DIR variable.
 FTPM_FLAGS ?= 						\
-	TA_CPU=cortex-a9				\
-	TA_CROSS_COMPILE=$(AARCH32_CROSS_COMPILE)	\
-	TA_DEV_KIT_DIR=$(OPTEE_OS_PATH)/out/arm/export-ta_arm32 \
-	CFG_TA_DEBUG=y CFG_TEE_TA_LOG_LEVEL=4 CFG_TA_MEASURED_BOOT=y
+	CROSS_COMPILE=$(CROSS_COMPILE_S_USER)	\
+	TA_DEV_KIT_DIR=$(OPTEE_OS_TA_DEV_KIT_DIR) \
+	CFG_MS_TPM_20_REF=$(MS_TPM_20_REF_PATH) \
+	O=out
```
etienne-lms, line 595: As per current https://github.com/OP-TEE/optee_ftpm/blob/master/README.md, `CFG_TA_MEASURED_BOOT` is default disabled. I think it should be enabled here, as previously done. 

Should `CFG_TA_DEBUG` be set when `DEBUG!=0`?

qemu_v8.mk
```diff
@@ -58,6 +61,7 @@ PAUTH ?= n
 # Option to configure Memory Tagging Extension
 MEMTAG ?= n
 
+
```
etienne-lms, line 64: Remove this extra empty line?

common.mk
jenswi-linaro: It doesn't matter (I tested to create an `optee-os-common` file) and the gnu make documentation doesn't mention it explicitly, but it's good practice. I'll move it.

common.mk
jenswi-linaro: > As per current https://github.com/OP-TEE/optee_ftpm/blob/master/README.md, `CFG_TA_MEASURED_BOOT` is default disabled. I think it should be enabled here, as previously done.

OK

> 
> Should `CFG_TA_DEBUG` be set when `DEBUG!=0`?

Sure, that makes sense.

jenswi-linaro:
Comments addressed.

jenswi-linaro:
Squashed and rebased

jforissier:

Could we have at least a basic "sanity check" test?

jenswi-linaro:
Tag applied.


> Could we have at least a basic "sanity check" test?

Yes. The simplest case I can think of is checking that `/dev/tpm0` is available if the fTPM TA is enabled.
I think it's best to check if the fTPM TA is enabled by trying to open a session to it. I'll add a test in xtest.

jforissier:
> Tag applied.
> 
> > Could we have at least a basic "sanity check" test?
> 
> Yes. The simplest case I can think of is checking that `/dev/tpm0` is available if the fTPM TA is enabled. I think it's best to check if the fTPM TA is enabled by trying to open a session to it. I'll add a test in xtest.

Sounds good!

PR #790, opened 2024-11-27
qemu_v8: add CC=... to TF_A_EXPORTS

jenswi-linaro:
Add CC="$(CCACHE)$(AARCH64_CROSS_COMPILE)gcc" to TF_A_EXPORTS for compiling TF-A v2.12.

Older and newer versions of TF-A can be compiled without issue if both CROSS_COMPILE and CC are set in TF_A_EXPORTS.

This dependency was introduced by the TF-A commit 3789c3c00090 ("build: determine toolchain tools dynamically")

### DIFF
--- a/qemu_v8.mk
+++ b/qemu_v8.mk
@@ -145,7 +145,8 @@ include toolchain.mk
 # ARM Trusted Firmware
 ################################################################################
 TF_A_EXPORTS ?= \
-	CROSS_COMPILE="$(CCACHE)$(AARCH64_CROSS_COMPILE)"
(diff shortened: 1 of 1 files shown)


jforissier:
> Older and newer versions of TF-A can be compiled without issue if both CROSS_COMPILE and CC are set in TF_A_EXPORTS.

That's interesting (meaning: kind of unexpected). But if it works then it's fine :)

jenswi-linaro:
Yeah :-)

Tag applied, thanks.

PR #791, opened 2024-11-28
qemu_v8: spmc_el3_manifest.dts: FF-A 1.1 and 32 cores

jenswi-linaro:
Update the manifest for the EL3 SPMC used when loading OP-TEE to FF-A version 1.1 and set execution-ctx-count to 32 in order to match the default number of cores supported in TF-A compiled for QEMU virt.


jforissier (commented):

jenswi-linaro:
Tag applied

PR #794, opened 2024-12-09
stm32mp1: support build for TF-A v2.11 and later

etienne-lms:
Building TF-A v2.11-rc0 and later fails when CROSS_COMPILE command includes space characters as when using ccache directives. This seems to be fixed by setting CC to the ccache and cross-toolchain directive.

### DIFF
--- a/stm32mp1.mk
+++ b/stm32mp1.mk
@@ -192,7 +192,9 @@ optee-os-clean: optee-os-clean-common
 ################################################################################
 # TrustedFirmware-A
 ################################################################################
-TFA_EXPORTS ?= CROSS_COMPILE="$(CCACHE)$(AARCH32_CROSS_COMPILE)"
+TFA_EXPORTS ?= CROSS_COMPILE="$(CCACHE)$(AARCH32_CROSS_COMPILE)" \
+	       CC="$(CCACHE)$(AARCH32_CROSS_COMPILE)gcc" \
+	       LD="$(CCACHE)$(AARCH32_CROSS_COMPILE)ld"
 
 TFA_DEBUG ?= $(DEBUG)
 ifeq ($(TFA_DEBUG),0)


stm32mp1.mk
```diff
@@ -192,7 +192,8 @@ optee-os-clean: optee-os-clean-common
 ################################################################################
 # TrustedFirmware-A
 ################################################################################
-TFA_EXPORTS ?= CROSS_COMPILE="$(CCACHE)$(AARCH32_CROSS_COMPILE)"
+TFA_EXPORTS ?= CROSS_COMPILE="$(CCACHE)$(AARCH32_CROSS_COMPILE)" \
+	       CC="$(CCACHE)$(AARCH32_CROSS_COMPILE)gcc"
```
jenswi-linaro, line 196: You should probably also add a `LD="$(CCACHE)$(AARCH32_CROSS_COMPILE)ld"` like in https://github.com/OP-TEE/build/pull/793/.

jforissier:
With `LD` set as suggested by @jenswi-linaro

stm32mp1.mk
etienne-lms: I saw that change but it seems TF-A does not need it. That said, maybe better in the long term, I'll add it. Thanks.

etienne-lms:
I've updated the change.
I don't really understand how TF-A deals with `CC` and `LD` known that I cross-compile from a x86 host and TF-A still successfully builds me x86 host tools like `fiptool` and few others.

stm32mp1.mk
jforissier: Did you try building on an arm64 server? It does not seem to be needed when building on x86_64, although it may be because the default linker may be different on the two machines that I tested. Therefore setting LD in all cases is more likely to be correct.

jforissier:
> I've updated the change.

Thanks.

> I don't really understand how TF-A deals with `CC` and `LD`

Neither do I :hand_over_mouth: 

>  known that I cross-compile from a x86 host and TF-A still successfully builds me x86 host tools like `fiptool` and few others.

I suspect the build might pick up a different LD than the one that comes witth the toolchain pointed to by CC.

jenswi-linaro:

etienne-lms:
Review tags applied. Thanks.

PR #795, opened 2024-12-11
Revert "qemu_v8.mk: enable MEASURED_BOOT_FTPM by default"

jforissier:
This reverts commit 9bf2bbe57184bb2c6ba73eacb2b11b7eeee40962. CI has become quite unstable since that commit was merged, see for instance the multiple failures in [1] and the absence of failure in [2]. I do not see any direct relationship betweem fTPM and the errors reported in the CI runs, but in order to make CI useful again we need to have it run reliably so disable MEASURED_BOOT_FTPM temporarily.

In my tests, 'make -j$(nproc) check-only XTEST_ARGS="-l 14"' always failed with a TA panic when MEASURED_BOOT_FTPM=y, and never when it is =n. The panic sometimes happens in xtest regression_4006, or sometimes in pkcs11_1021 or pkcs11_1023.

Link: [1] https://github.com/OP-TEE/optee_os/pull/7170
Link: [2] https://github.com/OP-TEE/optee_os/pull/7175

### DIFF
--- a/qemu_v8.mk
+++ b/qemu_v8.mk
@@ -35,9 +35,6 @@ ifeq ($(shell uname -m),x86_64)
 RUST_ENABLE ?= y
 endif
 
-# Enable fTPM
-MEASURED_BOOT_FTPM ?= y
-
 include common.mk
 
 DEBUG ?= 1


etienne-lms:

jenswi-linaro:

PR #801, opened 2024-12-18
qemu-check.exp: catch more RCU messages

jforissier:
Update the regular expression that catches RCU errors from the kernel in order to report an error upon:

  [   79.434969] rcu: INFO: rcu_preempt detected stalls on CPUs/tasks:

Without this fix, the test framework would either report a timeout (if the command causing the RCU message has hung), or report nothing (if the RCU message is triggered by a different thread of execution).


jenswi-linaro:

PR #802, opened 2025-01-03
qemu_v8: add Trusted Services support

jenswi-linaro:
Add support for building the SPMC test Secure Partitions (SPs) implemented by the Trusted Services project.

The build system is configured to build SPs with SEL0_SPS=y:
- the SPs are embedded into the OP-TEE binary
- the ffa_spmc testsuite is enabled in in xtest
- the out-of-tree linux-arm-ffa-user kernel module is build
- a S30test-arm-ffa-user init-script loads the kernel module during boot

Depends on https://github.com/OP-TEE/manifest/pull/306

Note that the ffa_spmc suite doesn't pass with this yet since SP1 doesn't boot due to an invalid address in the SP1 manifest in the Trusted Services git.


jenswi-linaro:
The problem is here:
https://git.trustedfirmware.org/plugins/gitiles/TS/trusted-services.git/+/refs/tags/v1.1.0/deployments/spm-test1/opteesp/default_spm_test1.dts.in#29
@imre-kis-arm @gyuri-szing @gabor-toth-arm @balint-dobszay-arm how can we fix this for QEMU? Removing "test-region" seems to work, but perhaps that's not what we want.

gyuri-szing:
Pls cherry-pick this TS change [1], then add `-DCFG_TEST_MEM_REGION_ADDRESS=<address>` to `SP_SPMC_TEST_EXTRA_FLAGS` in `qemu_v8.mk`.

The TS change allows platform specific configuration of the base address of the memory region used for testing. The platform specific value can be overridden from the CMake command line like above.

For base-address please use any secure physical address which is unused in the specific configuration of qemu_v8. The address must be aligned to the translation granule, and the test is using a single page only.

I tested the cmake code and it correctly modifies the manifest, but I did not actually run the tests.

1: https://review.trustedfirmware.org/c/TS/trusted-services/+/34314

jenswi-linaro:
Thanks @gyuri-szing, it works. We don't have a nice way of carving out secure memory in OP-TEE so I'm using the last secure page and keeping my fingers crossed that it's unused.

jenswi-linaro:
@gyuri-szing, by the way, I noticed the TS patch is now in the integration branch. Do you know when it's scheduled for the master branch? I'm asking because I'd like to update the qemu_v8.xml manifest to point to this commit, but I assume we should wait until it's on the master branch. It would be nice to include this in the coming 4.5.0 OP-TEE release.

gyuri-szing:
> @gyuri-szing, by the way, I noticed the TS patch is now in the integration branch. Do you know when it's scheduled for the master branch? I'm asking because I'd like to update the qemu_v8.xml manifest to point to this commit, but I assume we should wait until it's on the master branch. It would be nice to include this in the coming 4.5.0 OP-TEE release.

Will push an update today.

Do you plan to enable the SPMC test in the CI? AFAIR the blocker before was lacking support for qemu.

jenswi-linaro:
> > @gyuri-szing, by the way, I noticed the TS patch is now in the integration branch. Do you know when it's scheduled for the master branch? I'm asking because I'd like to update the qemu_v8.xml manifest to point to this commit, but I assume we should wait until it's on the master branch. It would be nice to include this in the coming 4.5.0 OP-TEE release.
> 
> Will push an update today.

Great!

> 
> Do you plan to enable the SPMC test in the CI? AFAIR the blocker before was lacking support for qemu.

Yes, I hope to get that into the 4.5.0 release.

jenswi-linaro:
@gyuri-szing do you have any update on the TS patch?

gyuri-szing:
> @gyuri-szing do you have any update on the TS patch?

Ouch, it has fallen trough the cracks and stuck in review. It is merged now.

jenswi-linaro:
> > @gyuri-szing do you have any update on the TS patch?
> 
> Ouch, it has fallen trough the cracks and stuck in review. It is merged now.

Thanks.

jforissier (commented):

But the tests (`xtest -t ffa_spmc`) do not work if the logged in user is not `root`. That's not optimal since all other tests work with the `test` user (`uid=102(test) gid=104(test) groups=103(teeclnt),104(test)`) and although CI does run as root...
Would it be possible to change the permissions of `/sys/kernel/debug/arm_ffa_user` (currently `0644 root:root`) so that it would work as non-root?

jforissier:
> Updated to work as non-root

Good enough :) thanks

PR #804, opened 2025-01-14
imx: update fdt_addr and initrd_addr

jenswi-linaro:
The kernel has grown and overwrites fdt_addr when loaded. Fix this by updating the fdr_addr and initrd_addr variables the addresses further up in memory.


jenswi-linaro:
Tag applied

PR #806, opened 2025-01-20
rpi3: update uboot.env.txt defaults

jbech-linaro:
* Change kernel commandline to use TCP instead of UDP. Recent rpcbind defaults use TCP only (check with: 'rpcinfo -p <myip> | grep nfs'). Using TCP should give a more robust connections overall.
* Add "nfsversion" variable to ease experimentation with different NFS versions when using various systems.
* Update "nfsserverip" IP address to match examples in RPi3 documentation.

This relates to https://github.com/OP-TEE/optee_docs/pull/260.


jforissier (commented):

jbech-linaro:

PR #807, opened 2025-01-23
qemu_v8: fix "SCMI_DTB" with QEMU v9.2.0

vingu-linaro:
Commit 1977c37e46d4 ("qemu_v8: support new path to QEMU v9.2.0") allows QEMU v9.2.0 to be used for 'make run[-only]', but it does not take into account 'make $(SCMI_DTB)'.

### DIFF
--- a/qemu_v8.mk
+++ b/qemu_v8.mk
@@ -577,6 +577,10 @@ QEMU_BASE_ARGS += $(QEMU_XEN)
 QEMU_BASE_ARGS += $(QEMU_EXTRA_ARGS)
 QEMU_BASE_ARGS += -machine virt,acpi=off,secure=on,mte=$(QEMU_MTE),gic-version=$(QEMU_GIC_VERSION),virtualization=$(QEMU_VIRT)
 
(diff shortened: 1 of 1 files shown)


etienne-lms (commented):
For the commit message: "_'make $(SCMI_DTB)_'" seems as strange command line. Should it rather be "_'make all WITH_SCMI=y'_"?

vingu-linaro:
> For the commit message: "_'make $(SCMI_DTB)_'" seems as strange command line. Should it rather be "_'make all WITH_SCMI=y'_"?

only the $(SCMI_DTB): has not been updated jens updated the run: and jerome fixed the check:
The commit message is a copy/paste/replace of [0264ae3d]( https://github.com/OP-TEE/build/commit/0264ae3dba685f06f4032bb6eaa8e0414e2cb8a4)

jforissier (commented):
Please reword:
> but it does not take into account 'make $(SCMI_DTB)'.
and use instead:
> but it does not update the $(SCMI_DTB) recipe.

With that:

vingu-linaro:
Does the latest version look ok for you ?

PR #810, opened 2025-01-24
qemu_v8.mk: use DEBUG default value from common.mk

jenswi-linaro:
Let common.mk set the default DEBUG value instead of setting a default value before including common.mk. The restores the default value 0 for DEBUG used before commit 85cc2db7efeb ("qemu_v8: add Trusted Services support"). This should also fix the CI build for Hafnium.

Fixes: 85cc2db7efeb ("qemu_v8: add Trusted Services support")

### DIFF
--- a/qemu_v8.mk
+++ b/qemu_v8.mk
@@ -38,8 +38,6 @@ endif
 # Enable fTPM
 MEASURED_BOOT_FTPM ?= y
 
-DEBUG ?= 1
-
 # Option to build with GICV3 enabled
 GICV3 ?= y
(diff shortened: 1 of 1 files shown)


jbech-linaro:

jforissier:

Although the Hafnium config should also build fine with DEBUG=1, shouldn't it?

jenswi-linaro:
It should, but it doesn't. I'll fix that once this is in place.

jenswi-linaro:
Tags applied.

PR #811, opened 2025-01-27
qmeu_v8.mk: -march=armv8.3-a for TF-A PAUTH=y

jenswi-linaro:
When building TF-A with pointer authentication enabled (PAUTH=y or SPMC_AT_EL=2), the compiler will need features from Arm Architecture version 8.3 for some instructions. The TF-A port for the QEMU platform by default uses Arm Arch 8.0 so with pointer authentication we need to specify a higher version to avoid errors like:
  CC      common/backtrace/backtrace.c
/tmp/cc6mqzev.s: Assembler messages:
/tmp/cc6mqzev.s:57: Error: selected processor does not support `xpaci x0'

These errors are currently limited to configurations with TF_A_DEBUG=1 but to keep it simple, always select armv8.3-a if pointer authentication is needed.


jforissier:
s/qmeu_v8/qemu_v8/ in the commit subject. With that fixed:

jenswi-linaro:
Comment addressed and tags applied.

PR #814, opened 2025-01-30
nanopc-t6: add fit_wrapper.sh with TEE_LOAD_ADDR=0x30000000

edtubbs:
Switch the NanoPC-T6 TEE load address from `0x08400000` to `0x30000000` by adding a local `fit_args.sh` and updating the device tree overlay. This aligns with [optee_os#7254](https://github.com/OP-TEE/optee_os/pull/7254), so merge this after that PR.

### DIFF
--- a/nanopc-t6.mk
+++ b/nanopc-t6.mk
@@ -22,8 +22,10 @@ BOOT_IMG		?= $(ROOT)/out/nanopc-t6.img
 TPL_BIN			?= $(BINARIES_PATH)/rk3588_ddr_lp4_2112MHz_lp5_2400MHz_v1.18.bin
 BOOT_CMD		?= $(ROOT)/build/nanopc-t6/nanopi6.h
 BOARD_DTSO		?= $(ROOT)/build/nanopc-t6/rk3588-nanopi6-optee.dtso
+FIT_WRAPPER		?= $(ROOT)/build/nanopc-t6/fit_wrapper.sh
 LINUX_DTSI		?= $(LINUX_PATH)/arch/arm64/boot/dts/rockchip/rk3588-nanopi6-common.dtsi
 UBOOT_HEADER		?= $(UBOOT_PATH)/include/configs/nanopi6.h
+UBOOT_FIT		?= $(UBOOT_PATH)/arch/arm/mach-rockchip/fit_wrapper.sh
 
 LINUX_MODULES ?= n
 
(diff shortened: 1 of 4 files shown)


nanopc-t6.mk
```diff
@@ -132,12 +134,16 @@ u-boot-proper: $(TPL_BIN) $(UBOOT_PATH)/.config u-boot-config
 u-boot-apply-bootcmd:
 	cp $(BOOT_CMD) $(UBOOT_HEADER)
 
+.PHONY: u-boot-apply-fit-args
+u-boot-apply-fit-args:
+	cp $(FIT_ARGS) $(UBOOT_FIT_ARGS)
```
jforissier, line 139: I see that you re using the same pattern as `u-boot-apply-bootcmd`, but I think it's a bad pattern. The good one would be:
```make
$(U_BOOT_FIT_ARGS): $(FIT_ARGS)
	cp $< $@
```

nanopc-t6.mk
```diff
@@ -132,12 +134,16 @@ u-boot-proper: $(TPL_BIN) $(UBOOT_PATH)/.config u-boot-config
...
+
 .PHONY: u-boot-loader
 u-boot-loader: u-boot-proper
 	$(UBOOT_PATH)/tools/mkimage -n rk3588 -T rksd -d $(UBOOT_PATH)/tpl/u-boot-tpl.bin:$(UBOOT_PATH)/spl/u-boot-spl.bin $(UBOOT_PATH)/idbloader.img
 
 .PHONY: u-boot
-u-boot: u-boot-apply-bootcmd $(TPL_BIN) $(UBOOT_PATH)/.config u-boot-loader
+u-boot: u-boot-apply-bootcmd u-boot-apply-fit-args $(TPL_BIN) $(UBOOT_PATH)/.config u-boot-loader
```
jforissier, line 146: And `$(U_BOOT_FIT_ARGS)` should be added instead of `u-boot-apply-fit-args`.

nanopc-t6/fit_args.sh
```diff
@@ -0,0 +1,163 @@
+#!/bin/bash
+#
+# Copyright (c) 2020 Rockchip Electronics Co., Ltd
+#
+# SPDX-License-Identifier: GPL-2.0
```
jforissier, line 5: The GPL license is somewhat problematic. Besides, you are copying the whole file ([link](https://raw.githubusercontent.com/friendlyarm/uboot-rockchip/d2512f90e9756130358771d4feb1d66d13dd6a90/arch/arm/mach-rockchip/fit_args.sh)) only to change `TEE_OFFSET`. Isn't there another way to specify the TEE load address?

edtubbs:
Thanks @jforissier! I’ve added `fit_wrapper.sh` to export `TEE_LOAD_ADDR` and updated the U-Boot config and Makefile to reference it instead.

jforissier:

PR #815, opened 2025-02-06
versal_net: add board support for AMD Versal Net

jcorbier:
Based on Versal support.

    Adds versal_net.mk and bootimage generation configuration
    files.

    Build commands:
      $ make toolchains
      $ make -j$(nproc)
      $ make -j$(nproc) bootimage

    The bootable image is versal/BOOT.BIN, usable with XSCT.

### DIFF
--- a/versal_net.mk
+++ b/versal_net.mk
@@ -1,0 +1,189 @@
+################################################################################
+# Following variables defines how the NS_USER (Non Secure User - Client
+# Application), NS_KERNEL (Non Secure Kernel), S_KERNEL (Secure Kernel) and
+# S_USER (Secure User - TA) are compiled
+################################################################################
+override COMPILE_NS_USER   := 64
+override COMPILE_NS_KERNEL := 64
+override COMPILE_S_USER    := 64
+override COMPILE_S_KERNEL  := 64
+
+WITH_CXX_TESTS = n
+
+# Network support related packages:
+BR2_PACKAGE_DHCPCD ?= y
+BR2_PACKAGE_ETHTOOL ?= y
+BR2_PACKAGE_XINETD ?= y
+
+# SSH Packages :
+BR2_PACKAGE_OPENSSH ?= y
+BR2_PACKAGE_OPENSSH_SERVER ?= y
+BR2_PACKAGE_OPENSSH_KEY_UTILS ?= y
+
+# Openssl binary
+BR2_PACKAGE_LIBOPENSSL_BIN ?= y
+
+BR2_TARGET_GENERIC_GETTY_PORT ?= ttyAMA1
+
+PLATFORM = versal-net-vnx-b2197-revA
+OPTEE_OS_PLATFORM = versal-net
+OPTEE_OS_COMMON_EXTRA_FLAGS = CFG_PKCS11_TA=y CFG_USER_TA_TARGET_pkcs11=ta_arm64 O=out/arm
+
+################################################################################
+# Paths to git projects and various binaries
+################################################################################
+TF_A_PATH		?= $(ROOT)/arm-trusted-firmware
+U-BOOT_PATH		?= $(ROOT)/u-boot
+BOOTGEN_PATH		?= $(ROOT)/bootgen
+LINUX_PATH		?= $(ROOT)/linux
+
+include common.mk
+
+################################################################################
+# Targets
+################################################################################
+
+all: tfa optee-os u-boot linux dtbo buildroot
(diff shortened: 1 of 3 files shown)


versal_net.mk
```diff
@@ -0,0 +1,192 @@
...
+	$(TF_A_EXPORTS) $(MAKE) -C $(TF_A_PATH) $(TF_A_FLAGS) clean
+
+################################################################################
+# OP-TEE
+#################################################################################
+
+optee-os: optee-os-common
+
+optee-os-clean: optee-os-clean-common
+	rm -f ${OPTEE_OS_PATH}/out/arm/core/tee_raw.bin
```
jforissier, line 72: Should be taken care of by the optee_os makefiles already (`core/arch/arm/kernel/link.mk:cleanfiles += $(link-out-dir)/tee-raw.bin`). This line should not be needed.

versal_net.mk
```diff
@@ -0,0 +1,192 @@
...
+# TF-A, Linux kernel, U-Boot and OP-TEE OS/Client/... are not built from their
+# related Buildroot native package.
+BR2_TARGET_ARM_TRUSTED_FIRMWARE=n
+BR2_LINUX_KERNEL=n
+BR2_TARGET_OPTEE_OS=n
+BR2_TARGET_UBOOT=n
+BR2_PACKAGE_OPTEE_CLIENT=n
+BR2_PACKAGE_OPTEE_TEST=n
+BR2_PACKAGE_OPTEE_EXAMPLES=n
+BR2_PACKAGE_OPTEE_BENCHMARK=n
```
jforissier, line 152: `?=` everywhere please (so that variables may also be set via the environment).

jforissier (commented):
Comments below.

versal_net.mk
jcorbier: Indeed. Good catch. Thanks!

versal_net.mk
jcorbier: Done. Thanks.

jcorbier:
@nathan-menhorn FYI.

versal_net.mk
```diff
@@ -0,0 +1,191 @@
...
+BR2_TARGET_GENERIC_ISSUE ?= "OP-TEE embedded distrib for $(PLATFORM)"
+BR2_TARGET_ROOTFS_EXT2 ?= y
+BR2_PACKAGE_BUSYBOX_WATCHDOG ?= y
+
+# TF-A, Linux kernel, U-Boot and OP-TEE OS/Client/... are not built from their
+# related Buildroot native package.
+BR2_TARGET_ARM_TRUSTED_FIRMWARE ?= n
+BR2_LINUX_KERNEL ?= n
+BR2_TARGET_OPTEE_OS ?= n
+BR2_TARGET_UBOOT ?= n
```
jforissier, line 147: OK I can see how these ones are built and packaged...

versal_net.mk
```diff
@@ -0,0 +1,191 @@
...
+BR2_PACKAGE_OPTEE_CLIENT ?= n
+BR2_PACKAGE_OPTEE_TEST ?= n
+BR2_PACKAGE_OPTEE_EXAMPLES ?= n
```
jforissier, line 150: ...but these ones are not built at all? Why? How do you test the resulting image?

versal_net.mk
```diff
@@ -0,0 +1,191 @@
...
+BR2_PACKAGE_OPTEE_BENCHMARK ?= n
```
jforissier, line 151: This one is deprecated and should be removed

jforissier (commented):
Some more comments I should have made earlier, sorry.

versal_net.mk
jcorbier: They are built from BR2_PACKAGE_OPTEE_*_EXT in common.mk actually, which is why BR2_PACKAGE_OPTEE_* are disabled (also the for the original Versal port and other ports as well).

versal_net.mk
jcorbier: Ok, will do.

versal_net.mk
jforissier: Oh yes, my bad. It's the non-`_EXT` flags. Well maybe they should be set to `n` by `common.mk` for clarity, but leaving them unset should work fine, no?

jcorbier: Checked and it works as expected. I removed the defines from versal_net.mk. Thanks!

PR #817, opened 2025-02-17
Add Buildroot package for Veraison Attestation example in OP-TEE

mmxsrup:
This PR adds a Buildroot package for veraison-rust-apiclient, enabling the Veraison Attestation example application to be built within the OP-TEE Buildroot environment.

The example application, introduced in https://github.com/linaro-swg/optee_examples/pull/124, is designed to validate the attestation functionality added in https://github.com/OP-TEE/optee_os/pull/7006. It demonstrates how a Client Application (CA) interacts with a Trusted Application (TA) to obtain attestation evidence from OP-TEE’s Platform Trust Application (PTA) and forward it to a remote Veraison Verifier server for validation.


mmxsrup:
Now, I encountered issues with installing the required package. 

The directory `out-br/build/veraison_rust_apiclient_ext-3fef382ad58465c5b6b3ad779d51d8035e3b0459` is expected to be created during the build process, but it does not appear. This suggests that the package fetch process might not be working correctly. I would appreciate any insights on what might be causing this issue.

PR #819, opened 2025-04-02
versal2: update build configurations

Akshay-Belsare:
Update the build sequence of components, build args for TF-A and OP-TEE. Remove OP-TEE memory range from reserved memory node of  dtso file.


Akshay-Belsare:
@jenswi-linaro change updated with acked tag. Thanks!

PR #828, opened 2025-04-29
toolchain: remove support for downloading or building Clang

jforissier:
Since [1] the regular Clang/LLVM toolchain provided by Linux distributions such as Ubuntu can be used as-is with OP-TEE. Therefore, remove get_clang.sh (which was used to fetch and assemble Clang+LLVM binaries including the compiler-rt libraries up to Clang 12 but not after due to newer compiler-rt libraries not being published) and clang/Dockerfile (which was used to build and install more recent versions). Also remove the related recipes (clang-toolchains and clang-toolchains-build).

With Ubuntu 24.10 for example just run "sudo apt install clang llvm lld" and that's it -- "make COMPILER=clang" will find the compiler in the PATH.

Link: https://github.com/OP-TEE/optee_os/commit/3d5c2dd8b59a [1]


jenswi-linaro:

PR #830, opened 2025-04-30
qemu_v8: add hafnium-clean target

jforissier:
Add a hafnium-clean target and add it to the clean: dependencies so that Hafnium can be cleaned and its submodules re-sync'ed.


jenswi-linaro:

PR #831, opened 2025-04-30
qemu_v7: add CC=... to TF_A_EXPORTS

jenswi-linaro:
Add CC="$(CCACHE)$(AARCH32_CROSS_COMPILE)gcc" to TF_A_EXPORTS for compiling TF-A v2.12.

Older and newer versions of TF-A can be compiled without issue if both CROSS_COMPILE and CC are set in TF_A_EXPORTS.

This dependency was introduced by the TF-A commit 3789c3c00090 ("build: determine toolchain tools dynamically")

### DIFF
--- a/qemu.mk
+++ b/qemu.mk
@@ -40,7 +40,9 @@ include toolchain.mk
 # ARM Trusted Firmware
 ################################################################################
 TF_A_EXPORTS ?= \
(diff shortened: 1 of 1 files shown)


jenswi-linaro:
Tag applied

PR #832, opened 2025-05-07
qemu: Fix uninitialized variables

raymo200915:
Fix uninitialized variables in procedure 'check_test_result'.


jforissier (commented):
Please make the prefix "qemu-check.exp:" rather than "qemu:". Other than that, looks fine, thanks for the patch!

raymo200915:
Addressed review comments and added review tag.

PR #835, opened 2025-06-09
versal2: reorganize Buildroot related configs

Akshay-Belsare:
Reorganize all Buildroot configuration settings specific to the platform before including common.mk, ensuring that the platform-specific default configurations are correctly applied and take effect as intended.


Akshay-Belsare:
@jforissier tags are applied to the commit.

PR #838, opened 2025-07-03
rockpi4: increase root FS size to 720M when LINUX_MODULES=y

jforissier:
The kernel has grown. Increase the root FS image size from 640 to 720 MB to avoid TEE_ERROR_STORAGE_NO_SPACE.


jforissier:
@jenswi-linaro may I add your Acked-by? Thanks!

jenswi-linaro:
Oh, absolutely. Sorry for the delay.

PR #843, opened 2025-08-11
qemu_v8: xen-create-image: increase root fs size from 100M to 120M

jforissier:
Since the Buildroot upgrade [1], it looks like the Xen root FS is a bit bigger resulting in:

 __populate_fs: Could not allocate block in ext2 filesystem while writing file "rootfs.cpio.gz"

Therefore increase the size from 100M to 120M. This leaves 1800 * 4 = 7200 KB free according to the command:

  $ dumpe2fs -h ../out/bin/xen.ext4 | grep -E '(Free blocks|Block size)'
  Free blocks:              1800
  Block size:               4096

Link:
https://github.com/OP-TEE/manifest/commit/aab39982ed1219b33ddae75c85953036de085eb2 [1]


jenswi-linaro (commented):

PR #855, opened 2026-01-08
rpi3: Suppress RWX segment warnings in TF-A build

jockebech:
When linking bl2.elf for rpi3, ld.bfd emits:
  warning: bl2.elf has a LOAD segment with RWX permissions

This warning was introduced in binutils 2.39 via --warn-rwx-segments and is treated as fatal by TF-A's default build flags, causing the build to fail. The issue surfaced after upgrading from GCC 11.x (which came with binutils 2.38) to GCC 14.x (which includes binutils 2.44).

Root cause: TF-A's BL2 linker script maps .text, .rodata, .data, and .bss into a single PT_LOAD segment, resulting in RWX flags.

Workaround: Add LDFLAGS="--no-warn-rwx-segments" to TF_A_FLAGS in rpi3.mk. This suppresses the warning allowing the build to succeed.

Drawbacks: Suppresses a security-related warning, which may mask legitimate issues.

There is a good blog post about it here [1].

Link: https://www.redhat.com/en/blog/linkers-warnings-about-executable-stacks-and-segments [1]

Tested-by: Joakim Bech <joakim.bech@gmail.com> (Rpi3)


jenswi-linaro:

jockebech:

PR #856, opened 2026-01-12
hikey: Suppress warnings in TF-A build

jenswi-linaro:
Updating the GCC toolchain to version 14.3 triggers new warnings in the old TF-A v2.6 used with Hikey. To make it build use the following flags:
  CFLAGS="-Wno-array-bounds"
  LDFLAGS="--no-warn-rwx-segments"


etienne-lms (commented):
`Acked-by: Etienne Carriere <etienne.carriere@st.com>`

jenswi-linaro:
Tag applied.

PR #858, opened 2026-04-09
imx: Fix TF-A v2.12 ccache issue

jenswi-linaro:
TF-A build system defines CROSS_COMPILE to hold the GCC target prefix. v2.12 implements more strict value checking which needs changes here.

### DIFF
--- a/imx.mk
+++ b/imx.mk
@@ -62,7 +62,10 @@ include toolchain.mk
 ################################################################################
 # ARM Trusted Firmware
 ################################################################################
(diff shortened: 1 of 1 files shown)


jenswi-linaro:
Please review so we can include it in the release.

jforissier:
`Acked-by: Jerome Forissier <jerome.forissier@arm.com>`

jenswi-linaro:
Tag applied.

PR #857, opened 2026-03-06
versal: Enhance board support for AMD/Xilinx Versal ACAP and Versal NET

joft-mle:
This pull request is build on top of previous work by @jcorbier & Co in merged PR #815 .

This pull request extends the existing AMD/Xilinx Versal ACAP and NET build instructions regarding the following aspects:

* Deduplicate versal_net.mk into versal.mk
* Add support for easily selecting Versal PS UART1 as primary UART; and/or the "other" PS UART for OP-TEE exclusively
* Load U-Boot FIT image with Linux included as part of BOOT.BIN by default
* Add support building custom, adjusted PLM (and PSM) Firmware regarding security features

This pull request depends on and shall be merged after [optee_os.git PR #7726](https://github.com/OP-TEE/optee_os/pull/7726) has been merged.

See [optee_docs.git PR #280](https://github.com/OP-TEE/optee_docs/pull/280) for updated documentation.

PR #860, opened 2026-04-29
versal: Move generation of ELF wrapper around tee-raw.bin into ...

joft-mle:
... `optee_os` repository

This pull request is a follow-up of PR #857 and [optee_os PR #7762](https://github.com/OP-TEE/optee_os/pull/7726).

As part of [discussion](https://github.com/OP-TEE/build/pull/857#discussion_r3111730188) in [build PR #857](https://github.com/OP-TEE/build/pull/857) it was decided that it would be better to perform the generation of the ELF wrapper directly within the `optee_os/` build system and to just rely on its availability in `build/versal.mk`.

This pull request depends on [optee_os PR #7798](https://github.com/OP-TEE/optee_os/pull/7798) and shall be merged after that one.

### DIFF
--- a/versal.mk
+++ b/versal.mk
@@ -429,56 +429,24 @@ endif
 # Boot Image
 ###############################################################################
 
-OPTEE_OS_ELF ?= $(shell dirname $(OPTEE_OS_BIN))/tee.elf
-OPTEE_OS_RAWBIN ?= $(shell dirname $(OPTEE_OS_BIN))/tee-raw.bin
-$(OPTEE_OS_ELF): optee-os-common
-$(OPTEE_OS_RAWBIN): optee-os-common
-
-OPTEE_OS_RAWBIN_OBJ ?= versal/$(shell basename $(OPTEE_OS_RAWBIN)).o
-OPTEE_OS_RAWBIN_ELF ?= versal/$(shell basename $(OPTEE_OS_RAWBIN)).elf
-
-# NOTE: Since we need to reference an ELF file in the .bif file to make PLM
-#       firmware really recognize OP-TEE OS as a SEL-1 binary and provide
-#       matching "handoff" parameters to ATF (API id 0x70b,
-#       get_atf_handoff_params), we wrap the regular tee-raw.bin file in ELF
-#       format with a single .text section and the appropriate entry point.
-#
-#       The tee.elf file cannot be used directly, since OP-TEE OS entry code
-#       (entry_a64.S) depends on a struct boot_embdata placed right at symbol
-#       __data_end. The script gen_tee_bin.py does this placement while crafting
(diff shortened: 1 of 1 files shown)


versal.mk
jenswikl: Why is this still needed?

joft-mle: @jenswi-linaro, you are referring to the `.PHONY` statement in particular, right? The (template) `.bif.in` file points to various (executable) files, either by "variable" (like `%IUB_PATH%`; added in #857) or statically (like the ARM Trusted Firmware executable; as before #857). Other than that, there is template variable `%IUB_LOAD%`, which is an address, not a file, and could be set on the `make` command line (via `IUB_BIF_LOAD`).

So we could move one step ahead and also explicitly list the hard-coded files (unlike before #857). But then we also have to deal with this address potentially changing.

jenswikl: Mixing a real target with a phony is confusing and would need to be explained with a comment, like what you just said, in this file.

However, I prefer something more straightforward, so how about making a proper phony target instead, like `tfa` or `bootimage`? Or you could even fold it into the commands for the `bootimage` target.

joft-mle:
Hi @jenswi-linaro,

in today's [2026/04/29 force-push](https://github.com/OP-TEE/build/compare/5b55bfdd648c6add28c3d6d15b8cfa39af59cc3f..712e12f49924f9d5f84e4c3e64861d7f79c86974) the following issues have been addressed:

- Remove (not to be `.PONY`) target `$(BIF_PATH)` and replace by non-file target `bif`, as [suggested by](https://github.com/OP-TEE/build/pull/860#discussion_r3161726065) @jenswi-linaro.
- Fix, also remove old code from target `bootimage-clean`, reference variables and files not existing at all anymore or not existing in `build/versal/` anymore. Replace (any) `versal/bootImage-${PLATFORM}.bif` by `$(BIF_PATH)` while at it.

versal.mk
joft-mle: [2026/04/29 force-push](https://github.com/OP-TEE/build/compare/5b55bfdd648c6add28c3d6d15b8cfa39af59cc3f..712e12f49924f9d5f84e4c3e64861d7f79c86974) uses target non-file target `bif` now.

jenswi-linaro:
Looks good.

joft-mle:
Hi @jenswi-linaro,

in today's [2026/04/30 force-push](https://github.com/OP-TEE/build/compare/712e12f49924f9d5f84e4c3e64861d7f79c86974..517910a43822cb787ee8436c6048ab79233a9e99) the following issues have been addressed:

- Add `Acked-by` tag [offered by](https://github.com/OP-TEE/build/pull/860#issuecomment-4350047279) @jenswi-linaro.

PR #861, opened 2026-05-19
Fvp updates

gyuri-szing:
This MR carries two changes:
- Allow specifying a telnet terminal command to the FVP when started. The change adds automatic discovery for tmux os screen session and will start each telnet terminal (emulated serial line) of the FVP in a dedicated window. (This change was inspired by the ``--terminals`` argument of the ``runfvp`` [command of meta-arm](https://github.com/jonmason/meta-arm/blob/master/documentation/runfvp.md).)
- Make redirecting the TS Logging SP output to a file optional.

### DIFF
--- a/fvp.mk
+++ b/fvp.mk
@@ -402,7 +402,9 @@ else ifneq ($(STY),)
 	FVP_SERIAL_CMD_ARGS += -C bp.terminal_3.terminal_command="$(FVP_TELNET_CMD)"
 endif
 ifeq ($(TS_LOGGING_SP),y)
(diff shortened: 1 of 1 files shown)


jforissier:
Thanks for pushing this. The `tmux` thing in particular is convenient :)

`Reviewed-by: Jerome Forissier <jerome.forissier@arm.com>`

gyuri-szing:
> Thanks for pushing this. The `tmux` thing in particular is convenient :)
> 
> `Reviewed-by: Jerome Forissier <jerome.forissier@arm.com>`

Thanks, applied. 🙂

PR #865, opened 2026-07-02
versal: boot with default PLM by disabling fuse-based HUK (CFG_VERSAL…

ldts:
…_HUK=n)

The default Versal PLM firmware does not enable fuse (eFUSE) support. Because of that the fuse-based HUK driver core/drivers/versal_huk.c cannot work: its tee_otp_get_hw_unique_key() relies on the PMC eFUSE+AES services (versal_efuse_read_dna -> versal_sha3_384 -> aes_gcm_encrypt, with CFG_VERSAL_HUK_KEY=12) and returns TEE_ERROR_GENERIC when those services are not enabled, so no HUK is available.

Without a HUK the secure-storage key cannot be derived, tee_fs_htree_open() fails and the "Secure Storage TA" store (priority 4) returns a generic error. The TA-store search in ldelf_syscalls.c only continues past a store on TEE_ERROR_ITEM_NOT_FOUND / TEE_ERROR_STORAGE_NOT_AVAILABLE, so the generic error aborts the search before the "REE" store (priority 9) that loads /lib/optee_armtz/*.ta. The result is that no TA can load and xtest fails on every TA-based test.

The project must boot out of the box with the default PLM, so disable the fuse-based HUK: with CFG_VERSAL_HUK=n OP-TEE uses the generic __weak tee_otp_get_hw_unique_key() (a stable, non-fuse HUK). Secure storage then works and TAs load; verified with xtest on VCK190. Re-enable CFG_VERSAL_HUK once the PLM enables fuse support, which then provides a real hardware-unique key.


jenswikl:

ldts:
@jenswikl thanks, tag added (and commit message reworded a bit for clarity)

PR #868, opened 2026-07-27
Add QEMU multi-arch simulation, Bootgen for AMD Versal Gen 2

Akshay-Belsare:
Add support for building and running QEMU-based multi-architecture simulation for the AMD Versal Gen 2 platform.

Dependent on https://github.com/OP-TEE/manifest/pull/352


versal2.mk
```diff
@@ -194,3 +197,27 @@ bootgen:
...
+
+$(QEMU_BUILD)/config-host.mak:
+	mkdir -p $(QEMU_BUILD)
+	cd $(QEMU_BUILD) && ../configure --target-list="aarch64-softmmu,microblazeel-softmmu,riscv32-softmmu" \
+		--enable-fdt --disable-kvm --disable-xen --enable-gcrypt --enable-slirp
+
+qemu: $(QEMU_BUILD)/.stamp_qemu
+
+$(QEMU_BUILD)/.stamp_qemu: $(QEMU_BUILD)/config-host.mak
+	$(MAKE) -C $(QEMU_BUILD) -j4
```
jenswikl: I believe the `-j` flag should be inherited from the top make.

versal2.mk
```diff
@@ -194,3 +197,27 @@ bootgen:
...
+	touch $@
+
+qemu-clean:
+	rm -rf $(QEMU_BUILD)
```
jenswikl: What if `QEMU_BUILD=/`?
Doesn't `$(MAKE) -C $(QEMU_PATH) distclean` work?
`rm -f $(QEMU_BUILD)/.stamp_qemu` is missing.

jenswikl (commented):
Please try to follow the 80 column limit in the patches.

versal2.mk
```diff
@@ -187,6 +191,42 @@ buildroot_mkimg: buildroot
 				-C gzip \
 				-d $(ROOTFS_GZ) $(ROOTFS_SIGN)
 
+################################################################################
+# Boot Image with Bootgen
+################################################################################
+
+bootimage: bootgen tfa optee-os u-boot dtbo
+	mkdir -p $(BINARIES_PATH)
+	@test -f $(BINARIES_PATH)/bl31.elf || \
```
jenswikl: I'm a bit puzzled over when these tests can fail, given the dependencies above.

versal2.mk
```diff
@@ -225,3 +265,121 @@ $(QEMU_BUILD)/.stamp_qemu: $(QEMU_BUILD)/config-host.mak
...
+	qemu qemu-devicetrees bootgen bootimage bootimage-clean \
+	bootgen-clean run-qemu-clean
+
+# Individual QEMU target recipes
+.PHONY: run-qemu-microblaze
+run-qemu-microblaze:
+	@echo "=== Starting MicroBlaze PMC QEMU ==="
+	@echo "Working directory: $(BOOT_IMAGES_PATH)"
+	@echo "Device tree: $(DEVICE_TREE_PATH)/board-versal2-pmxc-virt.dtb"
+	@test -f $(QEMU_MICROBLAZE_BIN) || \
```
jenswikl: Can't we just agree that `make all` should be run at least once before using any of these targets?
While these error messages are friendly, it seems that they will be hard to keep accurate.

versal2.mk
Akshay-Belsare: Agreed, that makes sense. We'll remove those tests in the next update.

versal2.mk
amey-raghatate: Thanks for the comment. We can't remove these file checks because these targets deliberately have no prerequisites. Adding them would let run-qemu-direct's backgrounded sub-processes independently re-trigger the same rebuild chain concurrently, and we hit a race condition this way. These checks are the only safety net for a direct invocation of make run-qemu-microblaze.

jenswikl: I don't understand. Surely `make all -j<whatever>` should work without races, or something is broken.

Akshay-Belsare: updated the code accordingly.
Rebased to latest master.

versal2.mk
```diff
@@ -59,8 +63,14 @@ ROOTFS_SIGN	?= $(BINARIES_PATH)/rootfs.cpio.gz.u-boot
 # Targets
 ################################################################################
 
-all: tfa optee-os u-boot linux dtbo buildroot buildroot_mkimg
-clean: tfa-clean optee-os-clean u-boot-clean linux-clean dtbo-clean buildroot-clean
+all: tfa optee-os u-boot linux dtbo buildroot buildroot_mkimg bootgen \
+     qemu qemu-devicetrees bootimage
+run: all
+	$(MAKE) run-only
+run-only: bootimage run-qemu-direct
```
jenswikl: The other `run-only` target in `fvp.mk`, `qemu.mk`, and `qemu_v8.mk`, doesn't depend on anything.
Since this `run-only` target doesn't even do anything, only depend on stuff, how about removing it to avoid confusion?
By the way, if `run-only` is executed with `-j<something>`, `run-qemu-direct` might execute before `bootimage` is done.

amey-raghatate: We'll keep run-only as a standalone target rather than removing it, similar to how it's done upstream. Instead of listing bootimage run-qemu-direct as prerequisites, run-only would do an explicit existence check for each required artifact (TF-A, OP-TEE, U-Boot, DTBs, bootgen, QEMU binaries, kernel Image, rootfs) before invoking the run step directly.
Let us know if this approach works, or if you'd prefer something different — will send the updated patch once confirmed.

jenswikl: Sure, that works.

jenswikl:

Akshay-Belsare:
Thanks @jenswikl!
The tag is now applied to the changes.

PR #870, opened 2026-08-12
Enable make check support for qemu_riscv64 platform

raymo200915:
This series adds initial `qemu_riscv64` platform support to `build`, including build/test integration and RISC-V-specific CI test handling.

It introduces:

  - `qemu_riscv64.mk`
  - `make check` support for `qemu_riscv64`
  - RISC-V-specific boot/config fragments
  - Support for applying an external upstream patch series

### Note:

1. This series consumes a `qemu_riscv64` manifest [1] from `optee_manifest` with a RISC-V-specific source combination for Linux, U-Boot, OpenSBI, QEMU, Buildroot, OP-TEE test, etc.

2. The optional patch application support in `qemu_riscv64.mk` also expects an external patch-series source. In the current integration this is provided by `optee_build_qemu_riscv64_patches` [2].

3. For CI/runtime validation, the host/container environment must also provide the required RISC-V build dependencies [3].

### Other related PRs / repo together to turn on the RISC-V64 xtest:

[1] PR for `qemu_riscv64` manifest
https://github.com/OP-TEE/manifest/pull/353

[2] Repo for `optee_build_qemu_riscv64_patches`
https://github.com/raymo200915/optee_build_qemu_riscv64_patches/tree/main

[3] PR for RISC-V build dependencies in dockerfile
https://github.com/jforissier/docker_optee_os_ci/pull/1

[4] Final step: PR for turning on the CI flow
https://github.com/OP-TEE/optee_os/pull/7927


common.mk
```diff
@@ -334,13 +334,15 @@ ifneq (y,$(BR2_PER_PACKAGE_DIRECTORIES))
 br-make-flags := -j1
 endif
 
+HOST_TOOLS_ENV ?=
```
jenswikl: Why is this needed?

qemu_riscv64.mk
```diff
@@ -0,0 +1,237 @@
...
+PLAT_QEMU ?= virt
+# Keep TLS-specific xtest cases disabled on qemu_riscv64 for now.
+# Enabling WITH_TLS_TESTS currently makes os_test.ta carry
+# R_RISCV_TLS_DTPMOD64 relocations, and ldelf then fails with:
+# "e64_relocate:732 Unknown relocation type 7".
+WITH_TLS_TESTS := n
+
+# Keep Buildroot host-python isolated from user-provided OpenSSL/Python
+# library overrides without changing the outer shell environment.
+HOST_TOOLS_ENV = env -u LD_LIBRARY_PATH -u PYTHONHOME -u PYTHONPATH
```
jenswikl: Why does this need to be RISCV64-specific?

qemu-riscv64-check.exp
jenswikl, line 1: DO we need a separate file for RISCV64, or can we add some options for the common expect script?

qemu_riscv64.mk
```diff
@@ -57,12 +64,39 @@ UBOOT_SPL		?= $(UBOOT_PATH)/spl/u-boot-spl.bin
 SDCARD_IMG		?= $(ROOT)/out-br/images/sdcard.img
 ROOTFS_EXT2		?= $(ROOT)/out-br/images/rootfs.ext2
 
+define apply_optional_patch_series
```
jenswikl: I expect that the problem with non-upstream patches is temporary. How about using temporary forks until it's resolved?

raymo200915: > I expect that the problem with non-upstream patches is temporary. How about using temporary forks until it's resolved?

I think this will take for a while, depends on the bandwidth of the MPXY author @peterlin-sifive.
I would suggest keep using `apply_optional_patch_series` on top of below baselines, when all MPXY patches are merged, we just need to update to the new baselines and unset the optional patch tag in the manifest, then it will automatically skip applying when they don't exist, looks quite flexible from my point of view.

Current baselines (also attached at https://github.com/OP-TEE/manifest/pull/353)
u-boot: refs/tags/v2026.07
linux: refs/tags/v6.19-rc2
opensbi: refs/tags/v1.8

qemu-riscv64-check.exp
raymo200915: > DO we need a separate file for RISCV64, or can we add some options for the common expect script?

I considered extending the common expect script, but the current `qemu-check.exp` still carries quite a lot of QEMUv8/Arm-specific behavior (Xen handling, DomU console flow, trusted-keys/rust dispatch, guest-side library-path handling, and the serial1 console model).

For `qemu_riscv64` the boot/test flow is currently simpler and single-console only, so I kept a separate script to avoid adding another layer of conditionals to the existing common one.

I agree there is room for refactoring later if we want to extract a smaller shared core plus platform-specific options. But at the moment, I think keep a separated and simpler expect script for `qemu_riscv64` is more realistic.

common.mk
raymo200915: > Why is this needed?

It is because I reproduced the host-side Python/OpenSSL environment contamination during Buildroot generation.
So the idea is to keep Buildroot host-python isolated.

qemu_riscv64.mk
raymo200915: > Why does this need to be RISCV64-specific?

The hook in `common.mk` is intentionally generic, while the current user is only the RISC-V QEMU platform.

It should not be RISC-V-specific, however, it is only enabled in `qemu_riscv64.mk` because that is where I reproduced the host-side Python/OpenSSL environment contamination during Buildroot generation. I cannot prove that same issue will happen with Arm platforms, thus I choose to narrow down its usage for `qemu_riscv64.mk` only at the moment.

raymo200915:
Updated to make use of `PLAT_QEMU` instead of hard-coding `virt` as suggested by @maroueneboubakri.

qemu_riscv64.mk
jenswikl: It's easier if we can do it the same way everywhere, with fewer things to keep track of, etc. Is there any chance that this could cause harm if used for the Arm version too?

qemu_riscv64.mk
jenswikl: Applying patches on top of checked out gits is rather hacky, and I anticipate problems. Unless there's a good reason not to, I'd rather see temporary forks that are updated as needed until we can switch to upstream.

qemu-riscv64-check.exp
jenswikl: How about moving the common part(s) into a separate file that can be included by this file and `qemu-check.exp` as needed? I'm concerned that, for instance, `check_test_result()` could diverge and cause different behavior.

qemu_riscv64.mk
raymo200915: I will give a try on Arm for it.

qemu-riscv64-check.exp
raymo200915: Your concern is right, I can move the common logics to a new file.

qemu_riscv64.mk
raymo200915: > Applying patches on top of checked out gits is rather hacky, and I anticipate problems. Unless there's a good reason not to, I'd rather see temporary forks that are updated as needed until we can switch to upstream.

Hmmm, but I think applying patches on top of baselines is quite common for example in OpenEmbedded Yocto recipes - it looks fine to manage both baselines and patches, and both the upstream baselines revision and the missing patches are clearly record in the manifest.

Of course, if you prefer the temporary forks, I can create them on RISE's official Gitlab repos and point the manifest to them.

Please let me know your thoughts.

jenswikl: Haha, I was just about to mention that we're not yocto. :-)

There's a difference between what Buildroot/yocto do and what we do. They prepare a build directory with source code and everything, while we here build from the checkout since we might want to play around with git, edit files, and stuff.

raymo200915: OK. That is fine. I will give an update shortly to point all dependent components to the Gitlab forks that managed by RISE SSWG.

raymo200915:
Moved HOST_TOOLS_ENV setting to `common.mk`.
Move common parts from `qemu-riscv64-check.exp` and `qemu-check.exp` to `qemu-check-common.exp`.
Droped `apply_optional_patch_series`.

qemu_riscv64.mk
raymo200915: I have moved `HOST_TOOLS_ENV` as common setting.

qemu-riscv64-check.exp
raymo200915: `qemu-check-common.exp` is created and included by `qemu-riscv64-check.exp` and `qemu-check.exp`.

qemu_riscv64.mk
raymo200915: I have dropped the external patch applying.

jenswikl:
Please squash in the fixups.

raymo200915:
All review comments are addressed. All fixups are squashed.

maroueneboubakri:
`Tested-by: Marouene Boubakri <marouene.boubakri@nxp.com>`

jenswikl:

jforissier:
`Acked-by: Jerome Forissier <jerome.forissier@arm.com>`

raymo200915:
Collected tags from @maroueneboubakri, @jenswikl and @jforissier , thanks!

